PR #1, opened 2017-03-21
Update for optional silverstripe/versioned module

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/6725


src/File.php
```diff
@@ -121,17 +121,16 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer, Thumb
...
     );
 
     private static $defaults = array(
         "ShowInSearch" => 1,
     );
 
     private static $extensions = array(
-        "SilverStripe\\ORM\\Hierarchy\\Hierarchy",
-        "SilverStripe\\ORM\\Versioning\\Versioned"
+        Hierarchy::class,
```
sminnee: How is Versioned added to File? It's important that this is on by default in our default installs.

sminnee: Aah versionedfiles.yml  in the versioned module

PR #8, opened 2017-04-13
Fix asset thumbnail size

phalkunz:


src/ImageManipulation.php
```diff
@@ -516,7 +516,7 @@ public function PreviewThumbnail()
      */
     public function Thumbnail($width, $height)
     {
-        return $this->Pad($width, $height);
+        return $this->Fill($width, $height);
```
flamerohr: I'm not sure about using `Fill()`, you lose the aspect ratio of the image uploaded...
Maybe `Fit()` would be better for this? Resizes to within the given width and height, but keeps aspect ratio.

phalkunz: `Fill()` keep the aspect ratio but crops the image to specified dimensions but `Fit()` doesn't crop at all. With this test image:

url

You can it with `Fit()` and `Fill()`

fit__

silverstripe_-_campaigns

The circle is still around in `Fill()` screen.

PR #9, opened 2017-04-20
FIX: Add PHP extension requirements to composer.

sminnee:
This will raise incompatibilities sooner rather than relying on the
installer.

All of the other required extensions are included in framework.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,11 +19,15 @@
     }
   ],
   "require": {
-    "silverstripe/framework": "^4@dev"
+    "silverstripe/framework": "^4@dev",
+    "php": ">=5.6.0"
   },
   "require-dev": {
     "silverstripe/versioned": "^1@dev",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -19,7 +19,10 @@
     }
   ],
   "require": {
-    "silverstripe/framework": "^4@dev"
+    "silverstripe/framework": "^4@dev",
+
+    "php": ">=5.6.0",
+    "ext-gd": "*"
```
kinglozzer: Is GD required if you just want to use Imagick?

dhensby: Good point, it's one or the other. I suggest we move them to suggestions...

tractorcow: +1

sminnee:
Suggestion applied

dhensby:
Oh, I think we should suggest both gd and imagick

PR #13, opened 2017-05-05
API Implement inherited file permissions

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6063

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -473,7 +473,18 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer, Thumb
             return $result;
         }
 
-        return $this->canEdit($member);
+        if (Permission::checkMember($member, self::EDIT_ALL)) {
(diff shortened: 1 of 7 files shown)


src/File.php
```diff
@@ -118,6 +125,8 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer, Thumb
         "File" => "DBFile",
         // Only applies to files, doesn't inherit for folder
         'ShowInSearch' => 'Boolean(1)',
+        "CanViewType" => "Enum('Anyone, LoggedInUsers, OnlyTheseUsers, Inherit', 'Inherit')",
+        "CanEditType" => "Enum('LoggedInUsers, OnlyTheseUsers, Inherit', 'Inherit')",
     );
```
tractorcow: Done.

PR #25, opened 2017-06-19
[WIP] App object

tractorcow:
Main issue: https://github.com/silverstripe/silverstripe-framework/issues/6681

See https://github.com/silverstripe/silverstripe-framework/pull/7037 for framework pull request

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -12,6 +12,7 @@ use League\Flysystem\Util;
 use SilverStripe\Assets\Storage\AssetNameGenerator;
 use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Assets\Storage\AssetStoreRouter;
+use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPStreamResponse;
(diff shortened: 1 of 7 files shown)


src/Flysystem/FlysystemAssetStore.php
```diff
@@ -417,21 +418,23 @@ protected function moveBetweenFilesystems($fileID, Filesystem $from, Filesystem
 
     public function grant($filename, $hash)
```
sminnee: For a like a change to this public API will be needed to avoid the `Controller::curr()` reference. I don't think it blocks this PR but I wonder if we should do that before beta1?

tractorcow:
Added a link back from the parent framework issue.

src/Flysystem/FlysystemAssetStore.php
chillu: I can't quite parse your sentence, but I assume you mean changing the request signature to `grant($filename, $hash, $session)`? I think that's a good idea, do you want to create a beta1 issue?

tractorcow: Since this discussion we were also thinking of having the current request registered with injector, rather than using Controller::curr().

I've raised https://github.com/silverstripe/silverstripe-framework/issues/7049 as the issue.

PR #26, opened 2017-06-21
intervention/image powered `Image_Backend`

dhensby:
I've spent this evening implementing an [intervention/image](https://github.com/intervention/image) powered backend for image manipulation.

The aim is to remove our need to maintain our own image manipulation classes and to allow greater manipulation of images going forward.

`intervention/image` provides many features out of the box that we don't currently have (watermarking, blurring, orientating of images based on exif data and more).

Feedback, please :)


src/InterventionBackend.php
```diff
@@ -0,0 +1,185 @@
...
+        // make sure the image is correctly orientated
+        $image->orientate();
+        $this->image = $image;
+        return $this;
+    }
+
+    public function writeToStore(AssetStore $assetStore, $filename, $hash = null, $variant = null, $config = array())
+    {
+        // Write to temporary file, taking care to maintain the extension
+        $path = tempnam(sys_get_temp_dir(), 'intervention');
```
tractorcow: How easy would it be to not need a temp file? Could we use setFromStream() instead? :P

tractorcow:
Please don't merge until at least 1 other @silverstripe/core-team has approved this.

src/InterventionBackend.php
dhensby: if we hand roll `$image->encode()` and then we could `setFromString($image->getEncoded()` but this approach allows the `ImageManager` to deal with the encoding automatically...

sminnee: My preference would be to use a subfolder of `TEMP_PATH` for this — what do you think?

stojg: I wanted to announce a small warning since I've worked with image manipulations in PHP, be careful of what image data that goes through PHP memory, especially larger files.

composer.json
```diff
@@ -20,6 +20,7 @@
   ],
   "require": {
     "silverstripe/framework": "^4@dev",
+    "intervention/image": "^2.3",
```
sminnee: We should get rid of `suggest: ext-gd` below, as that's intervention's responsibility.

dhensby: agreed

dhensby:
right, greens except for an sqlite failure which is due to that module not being compatable with SS4 currently

sminnee:
@dhensby could you do 2nd commit here that removes SQLite from this module's test suite.

dhensby:
Oh, and don't merge yet as I need to add docs:

- [ ] Note removal of `GDBackend` and `ImagickBackend`
- [ ] Explain how to switch between GD/Imagick drivers using config/injector
- [ ] Update image docs to ensure no mention of old backends

dhensby:
`SQLite` build removed.

sminnee:
This generally seems fine to me, I posted a few minor comments. But a net loss of 1000 LoC is nice reduction in maintenance burden.

As well as the package itself, guzzle/psr7 and psr/http-messages will also be pulled in as dependencies, but these are relatively small.

src/InterventionBackend.php
dhensby: Yes - good point - though the manipulated image is stored in memory already at this point.

That reminds me: I've actually removed the memory exhaustion protection, so I need to add that back in, if I can.

dhensby:
Right, I'm going to bed now but will complete the docs changes and memory exhaustion protection tomorrow :)

src/InterventionBackend.php
kinglozzer: Can we make use of `Image::stream()`? http://image.intervention.io/api/stream

src/InterventionBackend.php
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public function resize($width, $height)
+    {
+        return $this->createCloneWithResource($this->getImageResource()->resize($width, $height));
+    }
+
+    public function resizeRatio($width, $height, $useAsMinimum = false)
+    {
+        return $this->createCloneWithResource($this->getImageResource()->resize($width, $height, function ($constraint) use ($useAsMinimum) {
```
kinglozzer: Can you sort the line length?

src/InterventionBackend.php
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public function resizeByHeight($height)
+    {
+        return $this->createCloneWithResource($this->getImageResource()->heighten($height));
+    }
+
+    public function paddedResize($width, $height, $backgroundColor = "FFFFFF")
+    {
+        return $this->createCloneWithResource($this->getImageResource()->resizeCanvas($width, $height, 'center', false, $backgroundColor));
```
kinglozzer: Line length

src/InterventionBackend.php
```diff
@@ -0,0 +1,192 @@
...
+    public function setAssetContainer($assetContainer)
+    {
+        $this->container = $assetContainer;
+        return $this;
+    }
+
+    public function getImageManager()
+    {
+        if (!$this->manager) {
+            $this->setImageManager(new InterventionImage([ 'driver' => 'gd' ]));
```
kinglozzer: Can you include docs for how to use imagick with your docs PR?

kinglozzer:
Loving this ❤️. I know I added it, but I’ve never been truly happy with how the memory limit protection works... does `intervention/image` offer anything we might be able to use to tidy it up a bit?

dhensby:
> I’ve never been truly happy with how the memory limit protection works... does intervention/image offer anything we might be able to use to tidy it up a bit?

Nope - I googled and there are tickets about memory exhaustion in GD but their recommendation is to use  imagick instead...

src/InterventionBackend.php
dhensby: stream/encode are basically the same thing and we'll need to hand roll the settings that `save` determines for us... I'll take a look, though

dhensby: ok, `save` just pulls out the file extension (which we already do) and passes that along to encode, so there's no reason we can't do that

dhensby: Right, because stream returns a `Psr\Http\Message\StreamInterface` (which we don't currently support) we can't use the stream method.

We can use a string, though - I think it would be worth being compatible with PSR7 streams for this module, though so that could be another PR

src/InterventionBackend.php
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * @param \Intervention\Image\Image $image
+     * @return $this
+     */
+    public function setImageResource($image)
+    {
+        // make sure the image is correctly orientated
+        $image->orientate();
```
dhensby: I don't think so. I can't think of a time when you'd permanently want to turn off image orientation correction. perhaps for a one off you would, but I'm struggling to really know when you'd want that *from the automated image manipulations*

composer.json
chillu: This "fixes" the master build because it raises the PHPUnit dep from 4.x to 5.x, which in turn seems to fix a weird bug around invoking `@dataProvider` functions before calling test setup (which results in `No config manifests available`)

composer.json
```diff
@@ -20,14 +20,12 @@
...
     "php": ">=5.6.0"
   },
   "require-dev": {
     "silverstripe/versioned": "^1@dev",
-    "phpunit/PHPUnit": "~4.8"
-  },
-  "suggest": {
-    "ext-gd": "*"
+    "phpunit/phpunit": "^5.7"
```
chillu: This "fixes" the master build because it raises the PHPUnit dep from 4.x to 5.x, which in turn seems to fix a weird bug around invoking @dataProvider functions before calling test setup (which results in No config manifests available)

dhensby: Is that good or bad? I was just bumping the phpunit version to be in line with framework.

PR #29, opened 2017-06-23
NEW Decouple core from GD

dhensby:
`getDimensions` is currently coupled to the GD method `getimagesizefromstring`.

`getDimensions` is also a horrible API which is underused and has no real value given that all of the information it provides is accessible through the `Image_Backend` API or from `File` directly.

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -597,52 +597,13 @@ trait ImageManipulation
     }
 
     /**
-     * Get the dimensions of this Image.
-     *
-     * @param string $dim One of the following:
-     *  - "string": return the dimensions in string form
-     *  - "array": it'll return the raw result
-     *  - 0: return the height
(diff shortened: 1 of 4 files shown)


dhensby:
some other PRs for other modules incoming

src/ImageManipulation.php
```diff
@@ -597,52 +597,13 @@ public function getImageBackend()
...
-
-    /**
      * Get the width of this image.
      *
      * @return int
      */
     public function getWidth()
     {
-        return $this->getDimensions(0);
+        return $this->getImageBackend()->getWidth();
```
tractorcow: This breaks if calling getWidth() on something that's not an image. It should be some empty value.

tractorcow: Fix at https://github.com/silverstripe/silverstripe-assets/pull/38

PR #31, opened 2017-06-23
NEW Make sure we handle unsupported image encoding

dhensby:
Fixes an issue where file types that can't be manipulated breaks the CMS (see test failures in asset admin) https://travis-ci.org/silverstripe/silverstripe-asset-admin/jobs/246251060

This actually masks an issue where replacing an image with a non-image still attempts resamples...

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Assets;
 
+use Intervention\Image\Exception\NotSupportedException;
 use Intervention\Image\Exception\NotWritableException;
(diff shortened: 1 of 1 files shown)


dhensby:
Just to update this: we need this fix regardless of the issue in asset-admin. If someone tries to manipulate an unsupported image, we'll have issues.

tractorcow:
We need this fix on load of images, as well as on write. I'll mock up a PR.

tractorcow:
See https://github.com/silverstripe/silverstripe-assets/pull/38

PR #32, opened 2017-06-23
NEW Add complete list of supported image types

dhensby:
The `InterventionBackend` supports many more image types - these are now added to the config

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -189,10 +189,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
             'potm', 'potx', 'pps', 'ppt', 'pptx', 'rtf', 'txt', 'xhtml', 'xls', 'xlsx', 'xltm', 'xltx', 'xml',
         ),
         'image' => array(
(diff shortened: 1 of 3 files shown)


dhensby:
The test I've had to update seems fairly arbitrary as it just tests what the config is set to... I'd say it either needs removing or it needs to be made less opinionated

kinglozzer:
Do we need to worry about browser support here? i.e. if I upload a PSD, SilverStripe treats it as an image and tries to resample for the `UploadField` preview, does it output a jpeg? Or do we get an ``?

dhensby:
Hmm - good question - I guess it'd try to embed it as a PSD which wouldn't work :/

kinglozzer:
Possibly helpful: https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support

tractorcow:
Yeah, I think it needs to be the intersection of all:
 - images that can be resized
 - images that can be displayed in an  tag safely

tractorcow:
It looks like bmp and ico are fine, tiff has chrome issues.

dhensby:
so remove tiff and psd?

PR #34, opened 2017-06-26
FIX Orientation of images based on EXIF data now works

dhensby:
Exif data can only be read from either file paths or base64 encoded data-urls. It's not currently feasible to fetch the `data-urls` from the intervention image object because the exif data is stripped immediately on loading the image.


src/InterventionBackend.php
```diff
@@ -117,7 +141,22 @@ public function loadFromContainer(AssetContainer $assetContainer)
     {
         $hash = $assetContainer->getHash();
         $this->markStart($hash);
-        $this->setImageResource($this->getImageManager()->make($assetContainer->getStream()));
+        // write the file to a local path so we can extract exif data if it exists.
+        // Currently exif data can only be read from file paths and not streams
+        $path = tempnam(sys_get_temp_dir(), 'interventionimage_');
```
kinglozzer: Can we use the `TEMP_FOLDER` constant here?

kinglozzer:
One minor change but otherwise happy with the approach. We ultimately have two choices:

- Add this workaround to core
- Leave this out, meaning we can’t automatically fix image rotations in the CMS

This _may_ be fixed upstream at some point (https://github.com/Intervention/image/issues/745), but waiting for that could push this out quite far into the 4.x release cycle (or beyond, or never). I’m in favour of taking this in based on that, and the fact that CMS users expect their photos to be correctly oriented

dhensby:
review addressed

dhensby:
@tractorcow I'd like your view on this - FYI this is why I wanted to know if you can pull out a local path from an AssetContainer (if it existed locally) because that saves this double write... I'd like to save on disk I/O if we can...

tractorcow:
The point of having asset abstraction is that you can't get a local path, because you don't need it.

Looking at intervention repo there is already a workaround you can use.

https://github.com/Intervention/image/issues/336#issuecomment-77877177

Just read from the stream up to some safe max (e.g. 10k) and plug into exif_read_data(). Intervention requires that extension, so it's not any additional dependencies.

https://github.com/Intervention/image/blob/master/src/Intervention/Image/Commands/ExifCommand.php#L27

```php
$headerData = stream_get_contents($assetContainer->getStream(), 10240);
$exif = exif_read_data ( "data://image/jpeg;base64," . base64_encode($headerData) , 0 , true );
switch ($exif['Orientation']) {}
```

Then do whatever intervention does based on that (see https://github.com/Intervention/image/blob/master/src/Intervention/Image/Commands/OrientateCommand.php#L5)

dhensby:
> The point of having asset abstraction is that you can't get a local path, because you don't need it. Even if you did find a way to get it, I'd probably reject that as a PR.

I don't really follow this logic, nor think it's particularly relevant given that you've said you'd reject it anyway. Generating the `` tags for assets produces a path to a local file, so it's clearly possible and has valid use-cases where obtaining the path to an asset is both necessary and desirable.

Re-implementing the logic in the image library isn't a satisfactory solution given it defeats the point of using a library in the first place *and* the orientation needs to be done at write time, so we have to keep a reference to the original image data somewhere in memory so we can access it when we need it.

Writing tmp files to get images loaded into GD was already present before moving to a lib, so that shouldn't be a blocker now.

As I see it, our options are:
1. Merge this and roll back when/if the upstream library manages a fix
2. Wait for when/if the upstream library manages a fix

tractorcow:
> Generating the  tags for assets produces a path to a local file

Not always, it could be an s3:// url or even a base64 encoded inline image. :) @sminnee is working on splitting url generation into a service, so it's being abstracted even further.

I agree re-implementing logic in an upstream library sucks. I blame intervention for it's limitations.

sminnee:
> Not always, it could be an s3:// url or even a base64 encoded inline image. :) @sminnee is working on splitting url generation into a service, so it's being abstracted even further.

The use-case of what I did was primarily about exposing resources contained within your codebase to the browser. I think it's fine for assets to go through a separate pipeline.

sminnee:
In the case of an S3 back-end the flow would be:

 * Download the file from S3 to a local tmp1
 * Process the file writing to another local tmp2
 * Upload tmp2 to S3

If we mirrored that for local files it would be:

 * Copy the file from the assets folder to a local tmp1
 * Process the file writing to another local tmp2
 * Copy tmp2 to the assets folder

Is Dan's issue that this is less efficient than it needs to be because it could process straight from the assets folder to a different place on the assets folder?

I would assume that solving this would require new APIs so that the execution flow could be different for a local vs a remove asset store. It's not impossible, but it doesn't rate high on my priority list, as resampling only occurs when content changes which is less frequent than pageviews.

sminnee:
I'm not sure that forcing file/image processing tools to only ever use streams is particularly helpful. Processing on a temporary copy seems appropriate.

tractorcow:
Is there a unit test with a before - after oriented image that we can write here?

you can use https://github.com/recurser/exif-orientation-examples as a data source.

dhensby:
> Is Dan's issue that this is less efficient than it needs to be because it could process straight from the assets folder to a different place on the assets folder?

This is essentially what I'm getting at. I'm only asking *if* it's currently possible with existing APIs, I don't want a new API just for this purpose.

I'd like it if we could do:

```php
if ($assetContainer->isLocalFile()) {
    $path = $this->getLocalPath();
} else {
    $path = /* write stream to temp path and get set path var */
}
```

dhensby:
I'll look at writing a test

Issue #27, opened 2017-06-22
Tests referenced silverstripe/versioned but do not require it


robbieaverill:
[This test class](https://github.com/silverstripe/silverstripe-assets/blob/master/tests/php/AssetControlExtensionTest/VersionedObject.php#L20-L24) uses versioned, but composer doesn't require it.

Travis will be passing because it's required manually from the Travis configuration.

This is causing problems like this: https://travis-ci.org/silverstripe/silverstripe-tagfield/jobs/245966412

```
1) SilverStripe\TagField\Tests\StringTagFieldTest
InvalidArgumentException: silverstripe\assets\tests\assetcontrolextensiontest\versionedobject references nonexistent SilverStripe\Versioned\Versioned in 'extensions'
```

dhensby:
@tractorcow I think the problem here is that the `Extensible` trait and the `ExtensionMiddleware` are too opinionated about whether extension classes exist for tests to be able to run with or without optional modules.

Any ideas of ways around this?

tractorcow:
I'd add versioned to require-dev to get around this issue, to declare that although the module itself doesn't require versioned, it is expected to be installed for tests.

If you add an extension that doesn't exist the error should surface, as it has.

robbieaverill:
It's already a dev dependency here as of 0a779e8923df84959f4f70770113b1bd4aded4f6, which was there at the time that example Travis job failed - it didn't install silverstripe/versioned at the time.

dhensby:
I think this can be closed - there's no other way to fix this problem than what we've done.

The ideal solution is to have a way of adding extensions in tests only and not killing an entire build if the extension is there - that's an issue for `framework` and one that may never be addressed as killing the system is intended, it seems :(

PR #35, opened 2017-06-26
FIX Dont automatically apply Versioned module in tests as this will b…

dhensby:
…reak builds where versioned is not included

potential fix for #27


dhensby:
gah - doesn't work...

tractorcow:
If you remove versioned from our current test suite you will skip this entire test, even if you do it it "working". I don't think that's wise, as it drops our coverage.

tests/php/AssetControlExtensionTest.php
```diff
@@ -19,9 +20,28 @@ class AssetControlExtensionTest extends SapphireTest
...
+            Versioned::class,
+        ],
+    ];
+
+    public static function setUpBeforeClass()
+    {
+        try {
+            parent::setUpBeforeClass();
+        } catch (\InvalidArgumentException $e) {
+            static::markTestSkipped('Unable to build databas');
```
tractorcow: You've removed versioned from `.travis.yml` setup so when will this class ever be tested?

PR #38, opened 2017-06-28
BUG Fix non-images crashing when querying dimensions

tractorcow:
This is needed for graphql since object serialisation doesn't know if an object is an image or not until it's serialised. We need to report 0 for non-image files.

### DIFF
--- a/src/Image.php
+++ b/src/Image.php
@@ -39,9 +39,9 @@ class Image extends File implements ShortcodeHandler
      */
     private static $plural_name = "Images";
 
-    public function __construct($record = null, $isSingleton = false, $model = null, $queryParams = array())
+    public function __construct($record = null, $isSingleton = false, $queryParams = array())
     {
-        parent::__construct($record, $isSingleton, $model, $queryParams);
+        parent::__construct($record, $isSingleton, $queryParams);
         $this->File->setAllowedCategories('image/supported');
     }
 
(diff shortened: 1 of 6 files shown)


src/Image.php
```diff
@@ -39,9 +39,9 @@ class Image extends File implements ShortcodeHandler
      */
     private static $plural_name = "Images";
 
-    public function __construct($record = null, $isSingleton = false, $model = null, $queryParams = array())
```
tractorcow: Forgot to remove this when we removed datamodel class.

tests/php/FileTest.php
```diff
@@ -222,6 +206,38 @@ public function testValidateExtension()
...
+        $this->assertEquals(0, $pdf->getWidth());
+        $this->assertEquals(0, $pdf->getHeight());
+        $this->assertFalse($pdf->getIsImage());
+        $this->assertTrue($pdf->exists());
+        $this->assertNull($pdf->Pad(100, 100));
+
+        // Non-existant image
+        $image = new Image();
+        $image->Filename = 'folder/some-non-file.jpg';
+        $image->Hash = sha1('oogleeboogle');
```
tractorcow: Must be this exact string of course.

flamerohr: great string for a hash

PR #41, opened 2017-07-03
Update code for upgrader

tractorcow:
Part of https://github.com/silverstripe/silverstripe-framework/issues/7104


.upgrade.yml
```diff
@@ -38,13 +38,12 @@ mappings:
   Folder: SilverStripe\Assets\Folder
   Image: SilverStripe\Assets\Image
   Image_Backend: SilverStripe\Assets\Image_Backend
-  ImagickBackend: SilverStripe\Assets\ImagickBackend
```
flamerohr: could `ImagickImageTest` also be removed then?

PR #44, opened 2017-07-11
BUG Prevent brutal discard of constructed Image_Backend instances

tractorcow:
@mikenz thanks for calling us out on our bad coding. :)

With this fix the asset-admin performs a LOT faster.

### DIFF
--- a/src/ImageBackendFactory.php
+++ b/src/ImageBackendFactory.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace SilverStripe\Assets;
+
+use BadMethodCallException;
+use SilverStripe\Assets\Storage\AssetContainer;
+use SilverStripe\Core\Injector\Factory;
+
+/**
+ * Creates backends for images as necessary, avoiding redundant asset writes and loads
+ */
+class ImageBackendFactory implements Factory
+{
+    /**
+     * In memory cache keyed by hash/variant
+     *
+     * @var array
+     */
+    protected $cache = [];
+
+    /**
+     * @var Factory
+     */
+    protected $creator = null;
+
+    public function __construct(Factory $creator)
+    {
+        $this->creator = $creator;
+    }
+
+    /**
+     * Creates a new service instance.
+     *
+     * @param string $service The class name of the service.
+     * @param array $params The constructor parameters.
+     * @return object The created service instances.
+     */
+    public function create($service, array $params = array())
+    {
+        /** @var AssetContainer */
+        $store = reset($params);
+        if (!$store instanceof AssetContainer) {
+            throw new BadMethodCallException("Can only create Image_Backend for " . AssetContainer::class);
+        }
+
+        // Check cache
+        $key = sha1($store->getHash().'-'.$store->getVariant());
+        if (array_key_exists($key, $this->cache)) {
+            return $this->cache[$key];
+        }
+
+        // Verify file exists before creating backend
+        $backend = null;
+        if ($store->exists() && $store->getIsImage()) {
+            $backend = $this->creator->create($service, $params);
+        }
+
+        // Cache
+        $this->cache[$key] = $backend;
+        return $backend;
+    }
+}
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -8,9 +8,11 @@ use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Injector\InjectorNotFoundException;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use InvalidArgumentException;
+use SilverStripe\View\HTML;
 
 /**
  * Provides image manipulation functionality.
@@ -542,10 +544,12 @@ trait ImageManipulation
      */
     public function IconTag()
     {
-        return DBField::create_field(
+        /** @var DBHTMLText $image */
+        $image = DBField::create_field(
             'HTMLFragment',
-            '<img src="' . Convert::raw2att($this->getIcon()) . '" />'
+            HTML::createTag('img', ['src' => $this->getIcon()])
         );
+        return $image;
     }
 
     /**
@@ -587,14 +591,18 @@ trait ImageManipulation
      */
     public function getImageBackend()
     {
-        // Skip if non-image or file is broken
-        if (!$this->getIsImage() || !$this->exists()) {
+        // Skip files we know won't be an image
+        if (!$this->getIsImage() || !$this->getHash()) {
             return null;
         }
 
-        // Create backend for this object
-        /** @skipUpgrade */
-        return Injector::inst()->createWithArgs(Image_Backend::class, array($this));
+        // Pass to backend service factory
+        try {
+            return Injector::inst()->createWithArgs(Image_Backend::class, array($this));
+        } catch (InjectorNotFoundException $ex) {
+            // Handle file-not-found errors
+            return null;
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


tractorcow:
Note: This PR has raised an issue where the InterventionBackend isn't acting in an immutable way. Need to fix this before it can be merged.

mikenz:
Thanks :-) I'll have a play

tractorcow:
It looks as though

```php
$resized = $image->SizeAction(100, 200);
```

Is modifying the underlying $image. Perhaps a bug in InterventionImage::createCloneWithResource?

tractorcow:
Oh, DOH!

It's because InterventionImage is creating a temp file and that file is NOT segmented based on SHA1 or variant! This means multiple resizes will overwrite the same temp location.

tractorcow:
I think it's this `$this->setTempPath($path);`

The temp path isn't set correctly for the cloned image resource.

tractorcow:
Solution: Don't use a temp file.

tractorcow:
Ah, seems to be more than just that... will keep digging.

tractorcow:
ok, it seems to be a basic mutability issue with the backend itself.

```
return $this->createCloneWithResource($resource->crop($width, $height, $left, $top));
```

This code DOES clone the resource, but it clones it AFTER the operation is performed, meaning it's modifying the root object first.

createCloneWithResource should take a callback to be called AFTER the clone, not before it.

I'll fix tomorrow.

tractorcow:
@dhensby do you want to look at this mutability issue before I get my grubby hands on it?

src/ImageManipulation.php
```diff
@@ -581,20 +581,25 @@ public function getIcon()
...
      */
     public function getImageBackend()
     {
-        // Skip if non-image or file is broken
-        if (!$this->getIsImage() || !$this->exists()) {
-            return null;
+        // Pass to backend service factory
+        $backend = Injector::inst()->createWithArgs(Image_Backend::class, array($this));
+        if ($backend instanceof Image_Backend) {
+            return $backend;
```
flamerohr: Is this to ensure that `InvalidBackend` doesn't make it through?

Also noticed `$this->image_backend_cache` hasn't been used yet, and may not be needed anymore :)

src/ImageManipulation.php
```diff
@@ -581,20 +581,25 @@ public function getIcon()
     }
 
     /**
+     * Cache of image backends
+     *
+     * @var array
+     */
+    protected static $image_backend_cache = [];
```
kinglozzer: This can be removed now, right?

src/ImageBackendFactory.php
```diff
@@ -0,0 +1,63 @@
...
+        $key = sha1($store->getHash().'-'.$store->getVariant());
+        if (isset($this->cache[$key])) {
+            return $this->cache[$key];
+        }
+
+        // Create appropriate result
+        if ($store->exists() && $store->getIsImage()) {
+            $backend = $this->creator->create($service, $params);
+        } else {
+            $backend = new InvalidBackend();
```
flamerohr: so I guess  `InvalidBackend` is just that dummy so cache doesn't invalidate, a little bit sly.

kinglozzer:
One thing I’ve noticed in my testing (unrelated to this I think): when uploading a transparent png in assets, the copies that use `Fit` all end up with black backgrounds.

src/ImageManipulation.php
tractorcow: Yeah that was a mistake. Leftover from a previous attempt.

sminnee: As an aside, should we deprecate snake case for non-config statics?

tractorcow:
Ok, I've updated this PR with mutability protection. Tests are green!

If we can ever address https://github.com/Intervention/image/issues/745 we can also get rid of the temp file write; Another potential performance boost there. :)

src/ImageManipulation.php
tractorcow: We should be using non-config statics extremely sparingly in any case. We should use psr-2.

src/ImageBackendFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param string $service The class name of the service.
+     * @param array $params The constructor parameters.
+     * @return object The created service instances.
+     */
+    public function create($service, array $params = array())
+    {
+        /** @var AssetContainer */
+        $store = reset($params);
+        if (!$store instanceof AssetContainer) {
+            throw new BadMethodCallException("Cannot only create Image_Backend for " . AssetContainer::class);
```
tractorcow: Ok fixed now.

src/ImageBackendFactory.php
tractorcow: I got rid of this class. Now I just catch the injector exception, but this should only happen if you have a broken image anyway (getIsImage() should prevent this being called at the top level).

src/ImageManipulation.php
tractorcow: Fixed and removed the static.

tractorcow:
@mikenz have you any feedback from your testing?

sminnee:
I tested this PR with a folder of 185 x 1mb files and it's gone from timing out to running out of memory.

A snip of the call stack is here. It's reading the every file with Image_Backend to get the width/height metadata for the GraphQL responses. This needs to be persistently cached somehow (or saved to the DB)

Ideally, the cache for this data would be populated on upload.

```
96    8.0835    520243016    GraphQL\Executor\Executor::resolveField( )    .../Executor.php:227
97    8.0835    520243272    GraphQL\Executor\Executor::resolveOrError( )    .../Executor.php:442
98    8.0835    520243328    call_user_func:{/Users/sminnee/Sites/ss4/vendor/webonyx/graphql-php/src/Executor/Executor.php:488} ( )    .../Executor.php:488
99    8.0835    520243328    SilverStripe\GraphQL\TypeCreator->SilverStripe\GraphQL\{closure}( )    .../Executor.php:488
100    8.0835    520243704    call_user_func_array:{/Users/sminnee/Sites/ss4/graphql/src/TypeCreator.php:189} ( )    .../TypeCreator.php:189
101    8.0835    520244144    SilverStripe\AssetAdmin\GraphQL\FileTypeCreator->resolveField( )    .../TypeCreator.php:189
102    8.0835    520244144    SilverStripe\GraphQL\Util\CaseInsensitiveFieldAccessor->getValue( )    .../FileTypeCreator.php:248
103    8.0837    520244560    SilverStripe\View\ViewableData->__get( )    .../CaseInsensitiveFieldAccessor.php:66
104    8.0838    520244600    SilverStripe\Assets\File->getHeight( )    .../ViewableData.php:128
105    8.0845    520244840    SilverStripe\Assets\InterventionBackend->getHeight( )    .../ImageManipulation.php:631
106    8.0845    520244840    SilverStripe\Assets\InterventionBackend->getImageResource( )    .../InterventionBackend.php:313
107    8.0846    520244840    SilverStripe\Assets\InterventionBackend->loadFromContainer( )    .../InterventionBackend.php:214
108    8.0896    520245256    Intervention\Image\ImageManager->make( )    .../InterventionBackend.php:170
109    8.0896    520245504    Intervention\Image\AbstractDriver->init( )    .../ImageManager.php:50
110    8.0896    520245504    Intervention\Image\AbstractDecoder->init( )    .../AbstractDriver.php:64
111    8.0903    520245600    Intervention\Image\Gd\Decoder->initFromPath( )    .../AbstractDecoder.php:338
112    8.1281    532089528    Intervention\Image\Gd\Decoder->gdResourceToTruecolor( )    .../Decoder.php:68
113    8.1281    532089528    imagecreatetruecolor ( )    .../Decoder.php:136
```

tractorcow:
@sminnee I'll look at adding caching to image dimensions. That's probably the best I can do without implementing a complete filesystem cache;

sminnee:
:+1:

PR #48, opened 2017-07-13
BUG Ensure dimensions are cached

### LINKED ISSUE #47: getWidth / getHeight has poor performance
These values should be cached since they rely on an Image_Backend being constructed.

## Acceptance Criteria
 - A getWidth / getHeight only loads an Image_Backend once
 - A cache exists for saving this info

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-assets/pull/48

tractorcow:
Fixes #47 

I'll do some benchmarking now, but it seems to work if I run a bunch of pages, and then throw an Exception before loading assets from stream (to prove the image isn't loaded by the Image_Backend).

@sminnee do you want to review?

@dhensby I've refactored loadFromContainer so it doesn't immediately load the file into memory. Instead we wait until absolutely necessary.

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -9,6 +9,7 @@ use Intervention\Image\Exception\NotSupportedException;
 use Intervention\Image\Exception\NotWritableException;
 use Intervention\Image\Image as InterventionImage;
 use Intervention\Image\ImageManager;
+use Intervention\Image\Size;
 use InvalidArgumentException;
 use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\Storage\AssetContainer;
@@ -18,6 +19,15 @@ use SilverStripe\Core\Injector\Injector;
 
 class InterventionBackend implements Image_Backend, Flushable
 {
+    /**
+     * Cache prefix for marking
+     */
+    const CACHE_MARK = 'MARK_';
+
+    /**
+     * Cache prefix for dimensions
+     */
+    const CACHE_DIMENSIONS = 'DIMENSIONS_';
 
     /**
      * @var AssetContainer
@@ -110,6 +120,7 @@ class InterventionBackend implements Image_Backend, Flushable
      */
     public function setAssetContainer($assetContainer)
     {
+        $this->image = null;
         $this->container = $assetContainer;
         return $this;
     }
@@ -144,10 +155,32 @@ class InterventionBackend implements Image_Backend, Flushable
      */
     public function loadFromContainer(AssetContainer $assetContainer)
     {
+        return $this->setAssetContainer($assetContainer);
+    }
+
+    /**
+     * Get the currently assigned image resource, or generates one if not yet assigned.
+     * Note: This method may return null if error
+     *
+     * @return InterventionImage
+     */
+    public function getImageResource()
+    {
+        // Get existing resource
+        if ($this->image) {
+            return $this->image;
+        }
+
+        // Load container
+        $assetContainer = $this->getAssetContainer();
+        if (!$assetContainer) {
+            return null;
+        }
+
         // Avoid repeat load of broken images
         $hash = $assetContainer->getHash();
         if ($this->hasFailed($hash)) {
-            return $this;
+            return null;
         }
 
         // Handle resource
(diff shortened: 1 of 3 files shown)


src/InterventionBackend.php
```diff
@@ -144,10 +154,32 @@ public function setImageManager($manager)
      */
     public function loadFromContainer(AssetContainer $assetContainer)
```
sminnee: Should this method be deprecated now? It seems like clone of setAssetContainer.

src/InterventionBackend.php
```diff
@@ -294,27 +324,102 @@ public function getQuality()
...
+        $result = [0, 0];
+
+        // We can only cache dimensions from the source file (before any manipulations are performed),
+        // so ensure that the original image is un-loaded into memory.
+        // If a manipulation has been performed, we will return the dimensions from the already loaded
+        // image resource. In any case the 'variant' for these manipulations is unknown, so we couldn't
+        // cache those anyway.
+        $resource = $this->image;
+        if ($resource) {
+            return [
```
sminnee: I'd use `getSize()` here, which returns a `Intervention\Image\Size` object with `->width` and `->height` public params. It looks like the underlying code will be called twice otherwise.

Note that `getSize()` is a magic method implemented by one of the GetSizeCommand classes in intervention.

sminnee:
Note that this fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/536

src/InterventionBackend.php
tractorcow: It's a part of Image_Backend interface I can't remove it.

src/InterventionBackend.php
tractorcow: Size is a bit hard to serialize, since it has a pivot as well. I can look at it though.

src/InterventionBackend.php
sminnee: Didn't we invent Image_Backend, though?

src/InterventionBackend.php
sminnee: I just meant call once and convert to an array:

```php
$size = $resource->getSize();
return [ $size->width, $size->height ];
```

tractorcow: Oh I see, ok. :D

tractorcow: ok, done and pushed up.

tractorcow:
Updated with suggested fix.

src/InterventionBackend.php
tractorcow: I think removing it from the interface counts as API removal.

src/InterventionBackend.php
```diff
@@ -294,27 +325,111 @@ public function getQuality()
...
+     * @param string $variant
+     */
+    protected function warmCache($hash, $variant)
+    {
+        // Warm dimension cache
+        $key = $this->getDimensionCacheKey($hash, $variant);
         $resource = $this->getImageResource();
         if ($resource) {
-            return $resource->getHeight();
+            $result = [
```
sminnee: use getResourceDimensions() here too

src/InterventionBackend.php
```diff
@@ -294,27 +325,111 @@ public function getQuality()
...
+            return $this->getResourceDimensions($resource);
+        }
+
+        // Check if we have a container
+        $container = $this->getAssetContainer();
+        if (!$container) {
+            return $result;
+        }
+
+        // Check cache for unloded image
```
sminnee: The cache should probably be checked first

sminnee: Specifically, because the image can't change without the SHA also changing, the cases in which you'd want to discard the cache are rare enough that they can be handled with an explicit cache purge.

This means that the other logic isn't used unnecessarily: the normal flow will be that these fields are warmed on upload and then fetched from the cache for the rest of the object's lifetime.

tractorcow: You can't generate the key if the image has been resized, since the original tuple has been modified (and you can't get the updated "Variant" from anywhere).

tractorcow: > Specifically, because the image can't change without the SHA also changing

The SHA of the tuple is always the SHA of the original, not the SHA of the resized variant.

tractorcow: To put your fears at ease, even though we don't load from cache here, we CAN assume that these values are written to the cache in writeToStore().

The next time this thumbnail is loaded into InterventionBackend (another page load after the initial generation) the cache will be warmed,  $this->image is null, and cache should kick in.

sminnee: Ah yep I see, all good :+1:

tractorcow: This is what I attempted to convey in the code comment.

```php
// We can only cache dimensions from the source file (before any manipulations are performed),
// so ensure that the original image is un-loaded into memory.
// If a manipulation has been performed, we will return the dimensions from the already loaded
// image resource. In any case the 'variant' for these manipulations is unknown, so we couldn't
// cache those anyway.
```

tractorcow:
I've added the extra getResourceDimensions, fixed a typo, and improved the documentation to help explain why we seemingly incorrectly forget to query the cache. :)

tractorcow:
Docs are now:
```
// If we have a resource already loaded, this means we have modified the resource since the
// original image was loaded. This means the "Variant" tuple key is out of date, and we don't
// have a reliable cache key to load from, or save to. If we use the original tuple as a key,
// we would run the risk of overwriting the original dimensions in the cache, with the values
// of the resized instead.
// Instead, we use the immediately available dimensions attached to this resource, and we will
// rely on cache warming in writeToStore to save these values, where the "Variant" becomes available,
// before the next time this variant is loaded into memory.
```

tractorcow:
@dhensby requesting a review please :)

PR #50, opened 2017-07-13
BUG Fix ini2bytes

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7145

Note: Return type has been changed to float to work around max memory > PHP_MAX_INT


src/File.php
```diff
@@ -981,28 +981,12 @@ public static function format_size($size)
...
-                // deliberate no break
-            case 'k':
-                $quantity *= 1024;
-                // deliberate no break
-            default:
-                // no-op: pre-existing behaviour
-                break;
-        }
-        return $quantity;
+        Deprecation::notice('5.0', 'Use Convert::memstring2bytes instead');
```
dhensby: Should this be 2.0?

tractorcow: Ah, I think you're right sorry. We also don't set a per-module deprecation in the same way we do for framework.

tractorcow: https://github.com/silverstripe/silverstripe-assets/issues/51

PR #53, opened 2017-07-17
FIX Use merge and set instead of update for config calls

robbieaverill:
In preparation for https://github.com/silverstripe/silverstripe-framework/issues/6803


dhensby:
I think you can rebase out the ini2bytes commit

robbieaverill:
Cheers @dhensby - got a bit trigger happy on the rebase by the look of it 😄

PR #55, opened 2017-07-25
ENHANCEMENT Support for eventually consistent assets

tractorcow:
Partial implementation of https://github.com/silverstripe/silverstripe-framework/issues/5544

This enhancement allows us to take different approaches to different error conditions: Missing images (null streams) will approach a backing-off approach to re-attempts, where non-image content will permanently mark a specific file SHA against future attempts.

Both re-attempt durations and backing-off approaches can be customised on a per-error basis.

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -11,14 +11,19 @@ use Intervention\Image\Image as InterventionImage;
 use Intervention\Image\ImageManager;
 use Intervention\Image\Size;
 use InvalidArgumentException;
+use LogicException;
+use Psr\Http\Message\StreamInterface;
 use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\Storage\AssetContainer;
 use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 
 class InterventionBackend implements Image_Backend, Flushable
 {
+    use Configurable;
+
     /**
      * Cache prefix for marking
      */
@@ -28,6 +33,36 @@ class InterventionBackend implements Image_Backend, Flushable
      * Cache prefix for dimensions
      */
     const CACHE_DIMENSIONS = 'DIMENSIONS_';
+
+    /**
+     * How long to cache each error type
+     *
+     * @config
+     * @var array Map of error type to config.
+     * each config could be a single int (fixed cache time)
+     * or list of integers (increasing scale)
+     */
+    private static $error_cache_ttl = [
+        self::FAILED_INVALID => 0, // Invalid file type should probably never be retried
+        self::FAILED_MISSING => '5,10,20,40,80', // Missing files may be eventually available
+        self::FAILED_UNKNOWN => 300, // Unknown (edge case). Maybe system error? Needs a flush?
+    ];
+
+    /**
+     * This file is invalid because it is not image data, or it cannot
+     * be processed by the given backend
+     */
+    const FAILED_INVALID = 'invalid';
+
+    /**
+     * This file is invalid as it is missing from the filesystem
+     */
+    const FAILED_MISSING = 'missing';
+
+    /**
+     * Some unknown error
+     */
+    const FAILED_UNKNOWN = 'unknown';
 
     /**
      * @var AssetContainer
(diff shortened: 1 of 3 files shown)


src/InterventionBackend.php
```diff
@@ -30,6 +35,36 @@ class InterventionBackend implements Image_Backend, Flushable
...
+    /**
+     * This file is invalid as it is missing
+     */
+    const FAILED_INVALID = 'invalid';
+
+    /**
+     * This file is invalid because it is not image data, or it cannot
+     * be processed by the given backend
+     */
+    const FAILED_MISSING = 'missing';
```
flamerohr: The description of `FAILED_MISSING` and `FAILED_INVALID` look like they've been switch, confused or both.

src/InterventionBackend.php
```diff
@@ -30,6 +35,36 @@ class InterventionBackend implements Image_Backend, Flushable
...
+     * How long to cache each error type
+     *
+     * @config
+     * @var array Map of error type to config.
+     * each config could be a single int (fixed cache time)
+     * or list of integers (increasing scale)
+     */
+    private static $error_cache_ttl = [
+        self::FAILED_INVALID => 0, // Invalid file type should probably never be retried
+        self::FAILED_MISSING => '5,10,20,40,80', // Missing files may be eventually available
```
flamerohr: Couldn't `FAILED_MISSING` be an array of integers rather than a comma separated string?
Is it this way for convenience when using `yml` files?

src/InterventionBackend.php
```diff
@@ -636,36 +714,67 @@ protected function createCloneWithResource($resourceOrTransformation)
...
+                count($ttl) - 1
+            );
+            $this->getCache()->set($key.'_ttl', $index);
+            $ttl = $ttl[$index];
+        }
+        if (!is_numeric($ttl)) {
+            throw new LogicException("Invalid TTL {$ttl}");
+        }
+        // Treat 0 as unlimited
+        $ttl = $ttl ? (int)$ttl : null;
```
flamerohr: this could leak from the `if (is_array($ttl))` condition

src/InterventionBackend.php
```diff
@@ -636,36 +714,67 @@ protected function createCloneWithResource($resourceOrTransformation)
...
+        }
+        if (is_array($ttl)) {
+            $index = min(
+                $this->getCache()->get($key.'_ttl', -1) + 1,
+                count($ttl) - 1
+            );
+            $this->getCache()->set($key.'_ttl', $index);
+            $ttl = $ttl[$index];
+        }
+        if (!is_numeric($ttl)) {
```
flamerohr: wouldn't this throw if `$ttl` is an array?

src/InterventionBackend.php
tractorcow: Yeah, it's too hard to override a non-associative array.

You can still override the string with an array, or vice versa, however. :)

src/InterventionBackend.php
tractorcow: Ah true.

src/InterventionBackend.php
tractorcow: By this point it should no longer be an array.

src/InterventionBackend.php
tractorcow: It will no longer be an array by this point.

src/InterventionBackend.php
flamerohr: True, I got thrown off by the mix of `$ttl` and `$cacheTtl` :D

src/InterventionBackend.php
tractorcow: @flamerohr I've fixed this now and pushed up.

tractorcow:
Updated.

src/InterventionBackend.php
```diff
@@ -636,36 +714,67 @@ protected function createCloneWithResource($resourceOrTransformation)
...
+     * Determine reason this file could not be loaded.
+     * Will return one of the FAILED_* constant values, or null if not failed
      *
      * @param string $hash Hash of the original file being manipulated
-     * @return bool
+     * @param null $variant
+     * @return string|null
      */
-    protected function hasFailed($hash)
+    protected function hasFailed($hash, $variant)
```
flamerohr: maybe make `$variant` optional with null as default?
I've noticed the docblock doesn't match the string it expects and this being called with `$variant = null` occasionally.

src/InterventionBackend.php
```diff
@@ -386,6 +452,18 @@ protected function getResourceDimensions(InterventionImage $resource)
...
     /**
+     * Cache key for recording errors
+     *
+     * @param string $hash
+     * @param string $variant
+     * @return string
+     */
+    protected function getErrorCacheKey($hash, $variant)
+    {
+        return self::CACHE_MARK . sha1($hash . '-' . $variant);
```
flamerohr: `$variant` could potentially be null, but I guess leaving the trailing `-` is fine right? it's just generating a sha1 anyway :D

tractorcow: Fixed.

src/InterventionBackend.php
tractorcow: Fixed.

PR #56, opened 2017-07-26
API Implement copy / move at the filesystem level

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/562

I've also added some better tests for string manipulation of file identifier.

Note that this PR introduces two new API methods to both the AssetContainer and AssetStore interfaces; This would introduce a potentially (but low risk) breaking change to beta2.

CC @silverstripe/core-team for approval prior to merge.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -591,26 +591,18 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
             $this->OwnerID = Security::getCurrentUser()->ID;
         }
 
-        $name = $this->getField('Name');
+        $currentname = $name = $this->getField('Name');
         $title = $this->getField('Title');
 
         $changed = $this->isChanged('Name');
 
-        // Name can't be blank, default to Title
+        // Name can't be blank, default to Title or singular name
         if (!$name) {
-            $changed = true;
-            $name = $title;
+            $name = $title ?: $this->i18n_singular_name();
         }
-
-        $filter = FileNameFilter::create();
-        if ($name) {
-            // Fix illegal characters
-            $name = $filter->filter($name);
-        } else {
-            // Default to file name
+        $name = $this->filterFilename($name);
+        if ($name !== $currentname) {
             $changed = true;
-            $name = $this->i18n_singular_name();
-            $name = $filter->filter($name);
         }
 
         // Check for duplicates when the name has changed (or is set for the first time)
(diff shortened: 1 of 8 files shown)


tractorcow:
Well, this is an enhancement that fixes a bug. I've tagged as both. :)

src/File.php
```diff
@@ -803,6 +803,20 @@ public function onBeforePublish()
         }
     }
 
+    public function renameFile($newName)
```
sminnee: Under what circumstance will this method be called? I feel that in not actually renaming the file until a write we've broken the contract of the interface.

An alternative approach would be to call `$this->write()` as part of this call, or to trigger a single-field write if we didn't want  to do that.

However, the implications of this depend on where this method will be called from.

tractorcow: If you don't ever write the file you'll break any future reference to this file. Ensuring it immediately writes would fix this.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -318,6 +317,53 @@ public function delete($filename, $hash)
...
+        return $this->copyOrRename($filename, $hash, $newName, false);
+    }
+
+    /**
+     * Copy or move a file and all variants
+     *
+     * @param string $filename urrent filename
+     * @param string $hash Hash
+     * @param string $newName filename
+     * @param bool $asRename Set to true to rename (instead of copy)
```
sminnee: Boolean flag to switch between 2 different methods is a bit yuck. :-P

This might be better private than protected as even subclasses should use the copy & rename methods. That limits the impact of the less than ideal method signature.

src/Storage/AssetContainer.php
```diff
@@ -151,6 +151,23 @@ public function getVariant();
...
+     * Rename to new filename, and point to new file
+     *
+     * @param string $newName
+     * @return string Updated Filename
+     */
+    public function renameFile($newName);
+
+    /**
+     * Copy to new filename.
+     * This will not automatically point to the new file (as moveFile does)
```
sminnee: moveFile => renameFile?

tractorcow: Synonyms, sure, I should use consistent references.

src/Flysystem/FlysystemAssetStore.php
tractorcow: Yeah but private would make a rather complicated behaviour un-overridable. I'd rather duplicate the method inline and switch a few method calls instead.

dhensby:
Failure looks related

tractorcow:
Oh right, that's because we changed renameFile() to immediately save the record. Will fix.

tractorcow:
Yay green and ticked. :D

tests/php/Storage/AssetStoreTest.php
```diff
@@ -520,4 +621,68 @@ public function testProtect()
...
+        // New files exist
+        $this->assertTrue($backend->exists($newFilename, $fish1Tuple['Hash']));
+        $this->assertTrue($backend->exists($newFilename, $fish1Tuple['Hash'], 'somevariant'));
+        $this->assertTrue($backend->exists($newFilename, $fish1Tuple['Hash'], 'anothervariant'));
+
+        // Ensure we aren't getting false positives for exists()
+        $this->assertFalse($backend->exists($fish1Tuple['Filename'], $fish1Tuple['Hash'], 'nonvariant'));
+        $this->assertFalse($backend->exists($fish1Tuple['Filename'], sha1('nothash')));
+        $this->assertFalse($backend->exists($newFilename, $fish1Tuple['Hash'], 'nonvariant'));
+        $this->assertFalse($backend->exists('notfilename.jpg', $fish1Tuple['Hash']));
```
tractorcow: The point is to test that the moved file doesn't cause the false positives. It's meant to be a part of this test.

PR #59, opened 2017-08-16
FIX Ensure source file comments are disabled when generating htaccess file

### LINKED ISSUE #58: BUG SSViewer.source_file_comments enabled adds comments to .htaccess in assets
Steps to reproduce:

* Enable source file comments
```yaml
# mysite/_config/config.yml
SilverStripe\View\SSViewer:
  source_file_comments: true
```
* Go to `/admin?flush`
* Click on a page
* Observe internal server error
* Check `assets/.htaccess`, see source file comments added to it in HTML comments

robbieaverill:
Fixes #58


dhensby:
We really need a way to say what content type the template is so appropriate commenting styles can be used :/

robbieaverill:
True. Maybe that's a bigger thing, in that you could use SilverStripe for many types of templates content, in which case it might be useful. It's almost exclusively HTML at the moment

PR #61, opened 2017-08-29
ENHANCEMENT Prevent migration of disallowed file types

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/163

### DIFF
--- a/src/FileMigrationHelper.php
+++ b/src/FileMigrationHelper.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Assets;
 
 use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\ORM\DataList;
@@ -17,6 +18,17 @@ use SilverStripe\Versioned\Versioned;
 class FileMigrationHelper
 {
     use Injectable;
+    use Configurable;
+
+    /**
+     * If a file fails to validate during migration, delete it.
+     * If set to false, the record will exist but will not be attached to any filesystem
+     * item anymore.
+     *
+     * @config
+     * @var bool
+     */
+    private static $delete_invalid_files = true;
 
     /**
      * Perform migration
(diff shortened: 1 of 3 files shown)


src/FileMigrationHelper.php
```diff
@@ -78,6 +78,13 @@ protected function migrateFile($base, File $file, $legacyFilename)
             return false;
         }
 
+        // Remove this file if it violates allowed_extensions
+        $allowed = array_filter(File::config()->allowed_extensions);
+        $extension = strtolower($file->getExtension());
+        if (!in_array($extension, $allowed)) {
+            $file->delete();
+            return false;
+        }
```
flamerohr: This seems quite aggressive, given that files could've been added but disallowed some time after.

although I guess editing the file would always have a failed validation, but you wouldn't have lost the file.

Maybe @chillu could provide some insight for the direction of this.

dhensby: Agreed, this does seem aggressive - deleting files shouldn't be something we are doing. If they got there they should stay. Perhaps some fields have different allowed_types per instance that mean files that are ordinarily blocked are allowed under some points of entry...

chillu: I think that's too agressive as well, thanks for flagging that Chris. There might be viable use cases where files were added via other means to the CMS, but aren't user uploadable (not in allowed_extensions). Also, will this cause error pages (`assets/error-500.html`) to be served dynamically until they're republished manually by CMS authors?

tractorcow: This deletes ONLY the record. static files stay in place.

What should happen with the records for these types if they can't be imported, then?

flamerohr: @tractorcow perhaps we could have an opt-out for this disallowed file types feature?

The case @dhensby raised is quite valid, although I wonder how common that would be?

tractorcow: I could have a config flag to disable deletion; Instead it'd just not migrate those files.

What should the default for this be?

dhensby:
>And given that the file is still on the file system, you an access it like previously in ss3.

Using `File::get()` or `$this->FileRelation()` ? Surely if it's deleted from the DB it's going to break things?

src/FileMigrationHelper.php
flamerohr: we can have it not migrate disallowed_files and there's the ability opt-out of that.

flamerohr:
> Surely if it's deleted from the DB it's going to break things?

From what I understand, if it's kept in the `assets` folder, you'd still be able to access it as per a normal public file, although you'd need to know the file path or be able to guess it.

tractorcow:
Opt-out has been added, as has upgrading notes.

PR #72, opened 2017-09-16
Tests: FileShortcodeProviderTest assumes SiteTree create_default_page…

mikenz:
…s is enabled


tests/php/Shortcodes/FileShortcodeProviderTest.php
```diff
@@ -38,8 +40,9 @@ public function setUp()
             $file->setFromString(str_repeat('x', 1000000), $file->getFilename());
         }
 
-        // Conditional fixture creation in case the 'cms' module is installed
-        if (class_exists(ErrorPage::class)) {
+        // Conditional fixture creation in case the 'cms' and 'errorpage' modules are installed
+        if (class_exists(ErrorPage::class) && class_exists(SiteTree::class)) {
```
dhensby: `ErrorPage` can't exist without `SiteTree` so this addition is not needed

PR #79, opened 2017-10-04
Use index.php for serving content

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/7419 and https://github.com/silverstripe/silverstripe-framework/issues/7430


chillu:
@tractorcow I've moved the change to initParentURLPrefix to a different branch, since it's only related to the public webroot. Sorry been doing this all in one go, and was only partially untangled heh.

PR #80, opened 2017-10-04
Fix asset redirect to index.php

chillu:


tractorcow:
At the moment index.php isn't used / needed, and is in fact deleted in many sites which rely exclusively on `.htaccess`.

I might wait until we finish the public webroot folder work before merging this.

tractorcow:
Don't forget to pair this with the installer `assets/.htaccess` which comes with a pre-generated version of this file.

chillu:
Brainfart, sorry - this is already part of https://github.com/silverstripe/silverstripe-assets/pull/79

Issue #82, opened 2017-10-06
RFC Cache ImageShortcodeProvider::handle_shortcode()


tractorcow:
In sites with many images this can be a significant source of slowdown.

# Acceptance criteria:

 - On a cache-hit, no DB queries are required
 - File modification flushes the cache
 - Works with protected files (will need to store appropriate granted Hash / Filename in the cache)
 
# Suggested implementation
 
 ImageShortcodeProvider takes a CacheInterface dependency, which it uses to cache and restore HTML + grant information for each shortcode. Each key would be cached by shortcode args, as well as `SSViewer.global_key` config arg.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-assets/pull/89

tractorcow:
cc @chillu for review

chillu:
This isn't a regression since 3.x, right? Unless you're using a remote backend like S3, the shortcode handler should still just do filesystem stat calls like before. Through more layers of abstraction in 4.x, but I'd expect that the performance would roughly be the same. With that assumption, I'm removing the milestone, sounds like something for 4.1.

tractorcow:
The regression is the extra cost from permission checks and versioning that files have over 3.x.

axllent:
I would really urge you to reconsider this for 4.0 @chillu. The reason it appears that this, together with silverstripe/silverstripe-framework#7442 seem to be two bottlenecks that are **seriously** affecting front-end performance.

chillu:
OK, I've done a quick performance check, and embedding five images in the default `SiteTree.Content` field on a default `simple` theme page causes 13 queries in 3.x, and 70 queries in 4.x. That's in a single subfolder relative to the root. If I move them into another subfolder, it's up to 79 queries. So yes, we have a problem here. Prioritising this for 4.0.

chillu:
@tractorcow I'm assuming this will fix the main performance regressions that might hit websites?
https://github.com/silverstripe/silverstripe-framework/issues/7442 will apply more broadly (e.g. for CMS operations), but feels like something that's too late to introduce in 4.0.

> File modification flushes the cache

Do you think we can limit this to flushing one cache item when a specific file is flushed? If we flush every cache every time you modify an image, that cache will be a lot less effective.

tractorcow:
I'm going to look into whether or not we can turn off canView() in shortcodes; Potentially we can inherit from the parent record instead, and save a thousand queries. :)

tractorcow:
[...]

PR #83, opened 2017-10-08
ENHANCEMENT Inherit canView for shortcodes to speed up rendering

tractorcow:
Related to https://github.com/silverstripe/silverstripe-assets/issues/82, but it does not fix this issue.

Instead of checking canView() on all objects in a shortcode, simply rely on the parent object's canView(), and assume that it's inherited.

You can set this config to false to deny access to shortcode items, even when viewing the parent record.

Testing shows this can cut down on 500ms on some image-heavy pages.


chillu:
That looks reasonable. I can't think of a common use case where you'd want to publish a page, but only allow access to *some* content within shortcodes. And this is different from published vs. draft records in shortcodes (which presumably would still be enforced)

PR #89, opened 2017-10-17
Feature Cache image and file shortcodes

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/82

### DIFF
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -2,20 +2,27 @@
 
 namespace SilverStripe\Assets\Shortcodes;
 
+use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\File;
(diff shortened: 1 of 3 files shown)


src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -131,4 +160,42 @@ protected static function find_error_record($errorCode)
...
+
+    /**
+     * Gets the cache used by this provider
+     *
+     * @return CacheInterface
+     */
+    public static function getCache()
+    {
+        /** @var CacheInterface $cache */
+        return Injector::inst()->get(CacheInterface::class . '.ImageShortcodeProvider');
```
tractorcow: Incorrect key here.

flamerohr: Nice catch, thanks

PR #98, opened 2017-11-12
Added missing return possibility for Parent

zanderwar:
Parent() can either be one of `File` or `Folder`


dhensby:
A `Folder` is a `File` so this distinction isn't required...

PR #100, opened 2017-11-20
ENHANCEMENT Disable force_resample by default

### LINKED ISSUE #69: Resampled images are kept even if filesize is larger than original
# Acceptance Criteria

 * `ImageManipulation::$force_resample` changes its default from `true` to `false`
 * PHPDoc and dev docs are updated to reflect this change
 * If an image transformation is requested without requiring one (same size as the original image), the transformation is still applied to enforce the consistent application of image compression

# Notes

`ImageManipulation::$force_resample` defaults to `true`, which triggers `Resampled()` on upload (e.g. through `updateURL()`). The docs describe the feature as follows:

```
Since the 'master' images in your asset store may have a large file size, by
default SilverStripe will always apply compression to your images to save
bandwidth - even if no other manipulation (such as a crop or resize) is taking
place. If you expect the images in your asset store to already have
compression applied and want to serve up the original when no resampling is
necessary, you can add this to your mysite/config/config.yml file:
```

But: On some standard digital camera images I've uploaded, the images get *larger*. This clutters up the store with unnecessary data, right?

image

I'm not sure I understand this - why would anyone want their uploads to be resampled for the original downloads? I can understand setting compression ratios when image manipulation is required (resizing, cropping), but applying it to the original seems a like a bad default. It also does strange stuff like remove the embedded image profile from my digital camera uploads. And it will increase processing time overall on every upload. @dhensby @tractorcow What's the point of this feature? :D

# PR

 - [ ] https://github.com/silverstripe/silverstripe-assets/pull/100
 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7618

tractorcow:
Fixes #69

Docs PR https://github.com/silverstripe/silverstripe-framework/pull/7618

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -169,10 +169,13 @@ trait ImageManipulation
     abstract public function getIsImage();
 
     /**
+     * Force all images to resample in all cases
+     * Off by default, as this can be resource intensive to apply to multiple images simultaneously.
+     *
      * @config
-     * @var bool Force all images to resample in all cases
+     * @var bool
      */
-    private static $force_resample = true;
+    private static $force_resample = false;
 
     /**
      * @config


dhensby:
OK - so this has caused a regression because now (non-resampled) images that are orientated with EXIF data are not properly orientated when they are output via templates.

Doing `$Image` in the template will show an image that is incorrectly orientated because the resampler hasn't kicked in and so the image is incorrectly orientated

tractorcow:
Maybe we should implement upload middleware to fix that.

dhensby:
> Maybe we should implement upload middleware to fix that.

That's not going to allow us to get a patch fix in. At the moment there's no way for an `Image_Backend` to expose information about whether an image will require orientation.

My preferred fix would have been having some kind of `requiresResampling` method, but there's no nice way to expose that information without changing the `Image_Backend` interface which is going to be a breaking change...

PR #104, opened 2017-12-11
Feature Set .gitattributes to ignore docs and tests folder, and …

flamerohr:
some source asset files which are not needed

Parent issue: https://github.com/silverstripe/silverstripe-framework/issues/6890

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/artifacts/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -0,0 +1 @@
+artifacts/
```
dhensby: should this be `/artifacts/` (the directory in the root) or `artifacts/` (any directory in the project)?

flamerohr: Ack, I thought I had put the `/` prefix, will fix!

PR #105, opened 2017-12-12
BUG Less restrictive arguments for image resize

### LINKED ISSUE #103: ImageManipulation::castDimension triggers exceptions with values other than string or int
Seems like this was introduced with `1.0.1`. Suddenly `ImageManipulation` starts throwing several exceptions for me, because of this change.

There are modules (as well as my own code) that use  `round(…)` to pass image dimensions to image manipulation methods, but the result of that is always a `float` and will throw an exception.
It seems like the `filter_var($value, FILTER_VALIDATE_INT)` doesn't like floats…

Doesn't this violate SemVer as well?

tractorcow:
Fixes #103

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -943,18 +943,22 @@ trait ImageManipulation
      */
     protected function castDimension($value, $dimension)
     {
+        // Check type
+        if (!is_numeric($value)) {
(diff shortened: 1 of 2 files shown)


tractorcow:
Treating this as a bug, since the initial implementation of the input validation validated out completely valid inputs we simply hadn't considered.

PR #106, opened 2017-12-17
API Upgrade to PHPUnit 7, raise minimum PHP to 7.1

robbieaverill:
See https://github.com/silverstripe/silverstripe-framework/issues/7653

Requires:

- [x] https://github.com/silverstripe/silverstripe-framework/pull/7716

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,7 @@ sudo: false
 
 env:
   global:
-    - COMPOSER_ROOT_VERSION=1.x-dev
-    - CORE_RELEASE=master
+    - COMPOSER_ROOT_VERSION=2.x-dev
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -24,11 +23,11 @@ before_script:
   - phpenv config-rm xdebug.ini
   - export PATH=~/.composer/vendor/bin:$PATH
   - composer validate
-  - composer install --prefer-dist
-  - composer require symfony/config:^3.2 silverstripe/framework:4.0.x-dev embed/embed:^3.0@stable silverstripe/config:1.0.x-dev silverstripe/versioned:1.0.x-dev --prefer-dist --no-update
+  - composer install --prefer-source
+  - composer require symfony/config:^3.2 embed/embed:^3.0@stable --prefer-dist --no-update
```
tractorcow: I think this may need a rebase after 1 branch is merged up;

Why did you remove framework / config ? You need to force those to dev version. (y.x-dev).
At the moment we prefer using recipe-core 2.x-dev rather than framework directly though.

robbieaverill: That's cool, I think I must've assumed they were handled in the composer requirements. I'll update this PR once 1 is merged up =)

tractorcow: For the sake of testing in travis only we enforce dev instead of the composer-enforced prefer-dist, so the change in behaviour is intentional.

dhensby: It's merged up

tractorcow:
Can be merged on green.

PR #108, opened 2018-01-09
API Support public/ webroot folder

tractorcow:
Also Update filesystem utilities for path manipulations

Parent story https://github.com/silverstripe/silverstripe-framework/issues/7419

### DIFF
--- a/src/Flysystem/ProtectedAssetAdapter.php
+++ b/src/Flysystem/ProtectedAssetAdapter.php
@@ -20,14 +20,14 @@ class ProtectedAssetAdapter extends AssetAdapter implements ProtectedAdapter
      */
     private static $secure_folder = '.protected';
 
-    private static $server_configuration = array(
-        'apache' => array(
-            '.htaccess' => "SilverStripe\\Assets\\Flysystem\\ProtectedAssetAdapter_HTAccess"
-        ),
-        'microsoft-iis' => array(
-            'web.config' => "SilverStripe\\Assets\\Flysystem\\ProtectedAssetAdapter_WebConfig"
-        )
-    );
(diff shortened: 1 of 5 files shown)


tests/php/FilesystemTest.php
```diff
@@ -0,0 +1,120 @@
...
+     * @param string $error Expected path
+     */
+    public function testJoinPathsErrors($args, $error)
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage($error);
+        Filesystem::joinPaths($args);
+    }
+
+    public function providerTestJoinPathsErrors()
```
dhensby: only `test` methods should be public, providers should be `protected`

robbieaverill: That’s not true in phpunit 5, though I wish it was :-D

tractorcow: phpunit docs declare public providers still.

https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

Can you show me a reference to protected providers being preferred?

tractorcow:
```md
A data provider method must be public and either return an array of arrays or an object that implements the Iterator interface and yields an array for each iteration step. For each array that is part of the collection the test method will be called with the contents of the array as its arguments.
```

tractorcow: Same docs in 7.0 of phpunit.

dhensby: Yikes. I've always made them protected. Maybe it's a Codeception thing, then... My bad

PR #110, opened 2018-01-16
ENHANCEMENT Add DBFile::Link() alias for DBFile::getURL() so that it matches File::Link()

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7764

I know this is 4.0 branch, but if it's expected to be there and it isn't it's possibly a bugfix.


dhensby:
Is there a concept of absolute and relative link too? 😇

tractorcow:
I think this is probably enough. Most usages are `.Link` in templates and that is the major api surface I want to cover.

PR #111, opened 2018-01-18
fix stream readability detection

emitor-emitor:


filiplikavcan:
Why is it even necessary to check here? https://github.com/silverstripe/silverstripe-assets/blob/1.0/src/InterventionBackend.php#L227 it seem that NotReadableException is being handled later in the code https://github.com/silverstripe/silverstripe-assets/blob/1.0/src/InterventionBackend.php#L270 so isn’t it duplicity?

flamerohr:
For reference, table 1 here:
http://php.net/manual/en/function.fopen.php

tractorcow:
Seems ok.

PR #114, opened 2018-01-26
[WIP] add SiteConfig extension for root asset permissions

andrewandante:
Starting this out to see where I can get to.

Having trouble getting the permission dropdowns to toggle based on the radio buttons a la the page versions, any advice there would be appreciated.

Past that I just need to make sure that all the permissions that currently exist work together in an expected way.

### DIFF
--- a/_config/permissions.yml
+++ b/_config/permissions.yml
@@ -15,3 +15,6 @@ SilverStripe\Core\Injector\Injector:
     properties:
       Services:
         - '%$SilverStripe\Security\PermissionChecker.file'
+SilverStripe\SiteConfig\SiteConfig:
+  extensions:
+    - SilverStripe\Assets\Extension\SiteConfigPermissionsExtension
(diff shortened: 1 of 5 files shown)


dhensby:
Rather than adding a dependency on SiteConfig (and thus admin) could we do this PR against asset-admin and add some interface to that section that allows the defining of permissions there?

andrewandante:
I can see the logic in that - would you propose then a similar change to `SiteTree`? Have each section manage it's own root permissions? Not sure exactly where it would fit UI wise - maybe a second tab on the root?

dhensby:
> I can see the logic in that - would you propose then a similar change to SiteTree? Have each section manage it's own root permissions? Not sure exactly where it would fit UI wise - maybe a second tab on the root?

maybe, but not part of this PR ;)

andrewandante:
Use asset-admin as a POC for the eventual deprecation of SiteConfig it is :)

PR #115, opened 2018-02-01
BUG Ensure CMS authors can all see draft files by default

tractorcow:
Partial fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/730

Because files are used in many places, we allow anyone with CMS permissions to view draft files by default.


tractorcow:
Related https://github.com/silverstripe/silverstripe-versioned/issues/112

PR #118, opened 2018-02-13
API Shift TestAssetStore to non-test dev namespace

tractorcow:
Because `.gitattributes` is deleting this and breaking other modules that use this class.


flamerohr:
Did you need a follow pull request to dependent modules as well?

PR #122, opened 2018-03-19
API Model-agnostic file link tracking

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/75

Not done yet: Needs unit tests, but is functional.

### DIFF
--- a/src/Storage/DBFile.php
+++ b/src/Storage/DBFile.php
@@ -11,6 +11,7 @@ use SilverStripe\ORM\FieldType\DBComposite;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\ORM\ValidationResult;
 use SilverStripe\Security\Permission;
+use SilverStripe\View\ViewableData;
 
 /**
  * Represents a file reference stored in a database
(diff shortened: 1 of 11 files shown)


tractorcow:
Error should be fixed by https://github.com/silverstripe/silverstripe-framework/pull/7988

src/Storage/DBFile.php
```diff
@@ -403,7 +404,9 @@ public function getAbsoluteSize()
      */
     public function setOriginal($original)
     {
-        $this->failover = $original;
+        if ($original instanceof ViewableData) {
```
dhensby: This change seems unnecessary and also contradicts the PHPDoc which says the `$original` param should be an `AssetContainer`.

dhensby: I don't see how the type check has anything to do with using the setter or not?

tractorcow: Because you can't set a non-viewabledata object as a failover?

dhensby: 1. Is that true? Where are those limtations enforced?
2. The PHPDoc is thus wrong and needs correcting
3. This is a risky / major change as if someone had passed the documented type through (or any duck-typed object) then it would now break...

I'm just struggling to see how this change provides a defined benefit

tractorcow:
1. Look in setfailover and the failover logic.
2. No it's not wrong. Phpdoc is the external contract. At aren't violating it. We just say if the container is ALSO a viewabledata then use it as failover. At worst its ineligant not wrong.
3. It's making the code less risky by preventing possible invalid types being assumed as viewabledata.

PR #137, opened 2018-06-10
PATCH: removing duplicate key

sunnysideup:


tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-assets/pull/138

PR #143, opened 2018-06-12
BUG Remove duplicate .upgrade.yml keys

tractorcow:


robbieaverill:
It's a shame to lose your contributions on these PRs @sunnysideup, but we appreciate your work nonetheless!

sunnysideup:
@robbieaverill that is totally fine.  I made my PR requests super fast without thinking much hoping that it may be helpful.  The key is that they get done ASAP.  It takes me a ton of brain power to work out exactly what branch should be patched, what wording to use, etc... so I will leave that up to you as experts.

tractorcow:
My fault I actually created this without checking if it existed already. Rushing to get another release out that's why!

Issue #147, opened 2018-06-18
File/Image should cast use of "getTag" to HTMLFragment.


silbinarywolf:
**The problem**
Calling `$MyImage.Tag`  will just render out an `<img src>` tag as plain text.

**The solution**
Extend File's casting config to `Tag` and `getTag`. This allows you to use $Tag and $getTag in templates.
```
private static $casting = array (
        'TreeTitle' => 'HTMLFragment',
        'Tag' => 'HTMLFragment',
        'getTag' => 'HTMLFragment',
);
```

Source:
https://github.com/silverstripe/silverstripe-assets/blob/master/src/File.php#L160

**Current workaround / alternate solution**
I was using $Tag in a `<% with $Thumbnail %>` block. Instead I'm now using $Me

maxime-rainville:
All done.

PR #148, opened 2018-06-18
fix(File): Add HTMLFragment casting to $Tag / $getTag  - Github Issue #147

silbinarywolf:
fix(File): Add HTMLFragment casting to $Tag / $getTag  - Github Issue #147

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -157,7 +157,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     );
 
     private static $casting = array (
-        'TreeTitle' => 'HTMLFragment'
+        'TreeTitle' => 'HTMLFragment',
+        'getTreeTitle' => 'HTMLFragment',
+        'Tag' => 'HTMLFragment',
+        'getTag' => 'HTMLFragment',
     );
 
     private static $table_name = 'File';


tractorcow:
If this is a bugfix should we put it into 4 or 4.1 / 4.2 branches?

silbinarywolf:
Are you asking me? There are no 4.1/4.2 branches for this repo. It's using `1`, `1.1`, etc.

src/File.php
```diff
@@ -157,7 +157,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     );
 
     private static $casting = array (
-        'TreeTitle' => 'HTMLFragment'
+        'TreeTitle' => 'HTMLFragment',
+        'getTreeTitle' => 'HTMLFragment',
+        'Tag' => 'HTMLFragment',
+        'getTag' => 'HTMLFragment',
```
kinglozzer: Is `getTag` as well as `Tag` required? Perhaps we should fix that (separate to this PR) if so 😕

tractorcow: It should really just be the short name ('Tag') but it doesn't hurt I guess?

maxime-rainville:
That went to the wrong branch. I should have merge it against `1`

maxime-rainville:
I cherry picked the commit and did a PR against `1` #152.

PR #153, opened 2018-07-17
NEW: Support setting quality on a per-image basis

kinglozzer:
For example:

```
{$Image.Quality(5).Fit(10, 10)}
{$Image.Quality(100).Fit(1128, 500)}
```

This functionality currently exists in [modules](https://github.com/chrometoasters/silverstripe-image-quality), with the limitation that `.Quality` must be the last manipulation called. That limitation is addressed in this PR by re-using the `Image_Backend` instance.

Currently `{$Image.Quality(5).Fit(10, 10)}` will:

1. Set the quality to 10, save a resampled copy at the image’s original size with that quality setting
2. Re-load the new resampled copy into memory, then perform the `.Fit(10, 10)`
3. Save the down-sized copy out, but at the default quality setting (`90` is the default)

That often results in the `Image_Quality5_Fit1010.jpg` having a _larger_ filesize than `Image_Quality5.jpg`

`setQuality()` is already in the `Image_Backend` interface so there should be no BC-risk there.


kinglozzer:
Failure is real, I’ve an idea for a solution I’ll probably get around to never

kinglozzer:
Travis is segfaulting on PHP 5.6, and I can’t reproduce that on 5.6 locally (yay...). Looking into it

kinglozzer:
Okay, I’ve worked around the segfault - caused by passing a variable-by-ref into a closure... no idea why. Ready for review

src/ImageManipulation.php
```diff
@@ -831,7 +865,8 @@ function (AssetStore $store, $filename, $hash, $variant) use ($callback) {
                 // Write from another container
                 if ($result instanceof AssetContainer) {
                     try {
-                        return $store->setFromStream($result->getStream(), $filename, $hash, $variant);
+                        $tuple = $store->setFromStream($result->getStream(), $filename, $hash, $variant);
+                        return [$tuple, $result];
```
maxime-rainville: If you return a different type, won't this potentially cause problems for extension that expect the original type?

src/ImageManipulation.php
```diff
@@ -45,6 +45,11 @@
  */
 trait ImageManipulation
 {
+    /**
+     * @var Image_Backend
+     */
+    protected $imageBackend;
```
maxime-rainville: I would call this an implementation detail and make it private.

src/ImageManipulation.php
kinglozzer: My rationale here is:

1. This is an “internal” closure that we pass to `->manipulate()`, so I don’t think there’s any BC risk with this as it’s impossible for anyone to call the closure directly.
2. The [other approach I came up with](https://github.com/silverstripe/silverstripe-assets/commit/463bb89b95de329368fa5c561ded39d21ed71338#diff-64fe0e1766e08130ce105afde5d1ad67R853) that didn’t change the signature of the closure caused segfaults on PHP 5.6 😞

src/ImageManipulation.php
```diff
@@ -893,29 +930,36 @@ public function manipulate($variant, $callback)
...
         $store = Injector::inst()->get(AssetStore::class);
-        $result = null;
+        $tuple = $manipulationResult = null;
         if (!$store->exists($filename, $hash, $variant)) {
             // Circumvent generation of thumbnails if we only want to get existing ones
             if (!$this->getAllowGeneration()) {
                 return null;
             }
-            $result = call_user_func($callback, $store, $filename, $hash, $variant);
+            list($tuple, $manipulationResult) = call_user_func($callback, $store, $filename, $hash, $variant);
```
kinglozzer: @maxime-rainville I think this section might pose a potential BC-risk - if anyone is calling `$image->manipulate($variant, $myCustomClosure);` they’d have to update their return type. I’ll add in a check here to preserve BC

dhensby:
Still getting a seg fault

dhensby:
Green :)

kinglozzer:
Looks like it might have been a variant of https://bugs.php.net/bug.php?id=72286 causing the segfaults. I’ve adjusted the tests (fewer mock objects) and it should pass now

src/ImageManipulation.php
maxime-rainville: All good. I didn't click when I saw the conditional logic with the `Deprecation::notice` below.

src/ImageManipulation.php
```diff
@@ -588,6 +594,20 @@ public function Fill($width, $height)
         });
     }
 
+    /**
+     * Set the quality of the resampled image
+     *
+     * @param int $quality Quality level from 0 - 100
+     * @return AssetContainer
+     */
+    public function Quality($quality)
```
maxime-rainville: Do we want to have some range validation here? If you put a non-sensical value you'll get a `Intervention\Image\Exception\InvalidArgumentException` which is probably good enough. However `castDimension` does explicitly validate the provided dimension.

maxime-rainville: It probably would be worthwhile to add a `@throws` statement.

kinglozzer: I’ve implemented `castDimension` and added a `@throws` statement

PR #158, opened 2018-08-15
BUG Fix routing for files with dots in filename

### LINKED ISSUE #141: Files with dots in filename will no longer be served.
Introduced with version `1.1.1`, specifically: https://github.com/silverstripe/silverstripe-assets/commit/1e27835e0e7619de734c8b79074e4c7d8cfc071b

The addition of `^[^.]*` to the rewrite condition will deny access to files that have a dot in the filename. Example:
`assets/myfolder/d0ee4b8243/test.2.png` -> 403
whereas `assets/myfolder/d0ee4b8243/test.png` works as expected

Since dots are still allowed characters in a filename (eg. they don't get replaced when a file is being uploaded), this is probably a bug.

tractorcow:
Fixes #141

### DIFF
--- a/templates/SilverStripe/Assets/Flysystem/PublicAssetAdapter_HTAccess.ss
+++ b/templates/SilverStripe/Assets/Flysystem/PublicAssetAdapter_HTAccess.ss
@@ -25,7 +25,7 @@ AddHandler default-handler php phtml php3 php4 php5 inc
     RewriteRule error[^\\\\/]*\\.html$ - [L]
 
     # Block invalid file extensions
-    RewriteCond %{REQUEST_URI} !^[^.]*\\.(?i:css|js<% loop $AllowedExtensions %>|$Extension<% end_loop %>)$
+    RewriteCond %{REQUEST_URI} !^[^.]*[^\\/]*\\.(?i:css|js<% loop $AllowedExtensions %>|$Extension<% end_loop %>)$
     RewriteRule .* - [F]
 
     # Non existant files passed to requesthandler
--- a/tests/php/Flysystem/AssetAdapterTest.php
+++ b/tests/php/Flysystem/AssetAdapterTest.php
@@ -61,7 +61,7 @@ class AssetAdapterTest extends SapphireTest
         $htaccess = $adapter->read('.htaccess');
         $content = $htaccess['contents'];
         // Allowed extensions set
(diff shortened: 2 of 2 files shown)


robbieaverill:
Does this need an installer PR as well @tractorcow? https://github.com/silverstripe/silverstripe-installer/blob/4/public/assets/.htaccess#L28

tractorcow:
yeah it'll need updating there too.

madmatt:
Maybe I'm missing something, but if I try this on a live project (updating public/.htaccess to match what is listed in this PR) this stops normal files from working.

In other words, assets/abc123/filename.jpg now doesn't work, but assets/abc123/filename.v2.jpg does.

tractorcow:
@madmatt don't forget that this isn't a htaccess file, it's a template, so escaping may be different slightly (one extra level of `\` escaping).

The best way to test is to overwrite your template, do a dev/build flush and then test the auto-generated `.htaccess`. Does that help ?

unclecheese:
@maxime-rainville Makes a good point here about updating the UI, but it shouldn't be considered a prerequisite to merging this. We'll write up a separate story for fixing the UI.

pjayme:
Hi there, 

Would just like to ask why this PR was merged into `1.1` but not merged into the latest version (`1.2`)?

Cheers!

ScopeyNZ:
Hey Tobi. Sometimes bug fixes target earlier branches, but it's okay because the commits all exist in later branches, meaning they can be merged up easily.

One of the team (or another core committer) will probably merge it up (merge 1.1 into 1.2) soonish.

pjayme:
Makes sense, cheers for the quick response! :)

robbieaverill:
@pjayme I've merged the branches up now 💪

filiplikavcan:
Why not just \.(css|js)$ ?

It finishes for (test.js - test.jss) in 15 steps - 16 steps. The one from the pull request finishes for the same uri in 17 steps - 81 steps

It is even worse for assets/224287678/some/real/life/structure/test-filename.jss: 67 vs. 546 steps

tractorcow:
Do you mean only css / js files with dots in filename? We probably want to support `my.image.jpg` as well surely?

filiplikavcan:
That is just an example/testing regular expression to measure performance. The point is the the current expression takes longer to evaluate than my expression (which just checks the end of the string for allowed extensions)

PR #160, opened 2018-08-24
BUGFIX: Graceful validation of image shortcode

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-errorpage/issues/4

### DIFF
--- a/tests/php/Shortcodes/ImageShortcodeProviderTest.php
+++ b/tests/php/Shortcodes/ImageShortcodeProviderTest.php
@@ -100,7 +100,7 @@ class ImageShortcodeProviderTest extends SapphireTest
         $parser->register('image', [ImageShortcodeProvider::class, 'handle_shortcode']);
 
         $nonExistentImageID = 9999;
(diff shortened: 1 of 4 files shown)


tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -93,4 +93,23 @@ public function testShortcodeHandlerAddsDefaultAttributes()
...
         );
     }
+
+    public function testShortcodeHandlerFailsGracefully()
+    {
+        $parser = new ShortcodeParser();
+        $parser->register('image', [ImageShortcodeProvider::class, 'handle_shortcode']);
+
+        $nonExistentImageID = 9999;
+        while(Image::get()->byID($nonExistentImageID)) {
```
robbieaverill: A little space after `while` is probably what's breaking the builds

unclecheese: Ha. I ran `lint-clean`, but it's not defined for this package, so I figured I'd just let CI yell at me if it had a problem.

PR #162, opened 2018-09-04
FIX: Better error message for invalid upload

sminnee:
The invalid-extension upload error message now lists the extension of
the file uploaded.


sminnee:
OK, something I noticed while making this.

If you have no allowed-extensions set, then the default behaviour prior to this change was to allow any extension _except_ files without an extension. I assume that this is a bug, in that it should allow _all_ files?

sminnee:
OK, on that point, I've added a test in the final commit that confirms this previously untested behaviour.

(Recommend merging this as 3 separate commits as submitted)

src/Upload_Validator.php
```diff
@@ -236,15 +234,26 @@ public function isFileEmpty()
...
+            return strtolower($pathInfo['extension']);
 
-        // Special case for filenames without an extension
-        if (!isset($pathInfo['extension'])) {
-            return in_array('', $this->allowedExtensions, true);
+        // Special case for files without extensions
         } else {
-            return (!count($this->allowedExtensions)
-                || in_array(strtolower($pathInfo['extension']), $this->allowedExtensions));
+            return '';
```
robbieaverill: We don't really need the else here

src/Storage/DBFile.php
```diff
@@ -506,7 +506,11 @@ public function validate(ValidationResult $result, $filename = null)
             return true;
         }
 
-        $message = _t('SilverStripe\\Assets\\File.INVALIDEXTENSION_SHORT', 'Extension is not allowed');
+        $message = _t(
+            'SilverStripe\\Assets\\File.INVALIDEXTENSION_SHORT',
+            'Extension \'{extension}\' is not allowed',
```
robbieaverill: I only have one concern with this in that if anyone does a patch release on 4.0, 4.1 or 4.2 and updated translations, we'll get this as a string literal pulled into new projects. I think the safer option would be to create a new translation key with this new value in it

robbieaverill:
@sminnee fancy resurrecting this? If we can use a new translation key then it's probably good to go

src/Storage/DBFile.php
robbieaverill: Cool, needs a text collector update too

PR #164, opened 2018-09-04
Fix files with dots being marked as invalid (Fixes #141)

### LINKED ISSUE #141: Files with dots in filename will no longer be served.

wilr:


michalkleiner:
Can we consider this option? It only cares about the last dot and an extension and doesn't have to "eat" all the previous characters.
`!\\.(?i:css|js<% loop $AllowedExtensions %>|$Extension<% end_loop %>)$`

PR #165, opened 2018-09-20
FIX FlysystemAssetStore::getAsURL() only grant for protected filesystems

christopherdarling:
Fix for https://github.com/silverstripe/silverstripe-framework/issues/8384

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -226,15 +226,15 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
             /** @var PublicAdapter $publicAdapter */
             $publicAdapter = $public->getAdapter();
             return $publicAdapter->getPublicUrl($fileID);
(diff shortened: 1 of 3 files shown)


robbieaverill:
I'm seeing the first hit to a page containing a published image in its content NOT starting a session and requesting a protected file grant, but subsequent requests do request a grant still. I'll have a bit more of a look into this, my assumption is that public assets shouldn't require a grant at all, and therefore not start a session.

robbieaverill:
Ah, it's still being called from `ImageShortcodeProvider::handle_shortcode()`

robbieaverill:
I've pushed a new commit which changes `ImageShortcodeProvider::handle_shortcode()` to use `FlysystemAssetStore::getAsURL()` rather than `::grant()`, so that it will grant if it needs to but not always. I've also added some unit tests to solidify the logic that we're adding in this pull request, which is "The asset store should only create a grant for protected assets, and only when it's told to (default: true)"

PR #166, opened 2018-10-01
BUG Keep folder Name and Title in sync on update

lukereative:
This was discovered while working to improve folder sorting, folders are sorted by Title but when you update their Name through the admin interface it would not update the Title and they would end up out of sync and quietly mess up alphabetical sorting.

**Related but not dependant PR** https://github.com/silverstripe/silverstripe-asset-admin/pull/844

### DIFF
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -92,16 +92,9 @@ class Folder extends File
 
     public function onBeforeWrite()
     {
-        // Make sure Title is updated if Name is chaged (as we keep these in sync for folders)
-        if ($this->isChanged('Name')) {
-            $name = $this->Name;
-            $this->setField(
-                'Title',
-                str_replace(array('-','_'), ' ', preg_replace('/\.[^.]+$/', '', $name))
-            );
-        }
-
         parent::onBeforeWrite();
+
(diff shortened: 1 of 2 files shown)


robbieaverill:
I'm kind of tempted to suggest that this is a feature, not a bug

lukereative:
It isn't a feature, it's just missed because we already do it on 'setTitle' https://github.com/silverstripe/silverstripe-assets/blob/63cabc7fc84f295a95e88f2ce37f940b61b97223/src/Folder.php#L127-L133 
I'm going to add a unit test

robbieaverill:
Nice one!

src/Folder.php
```diff
@@ -90,6 +90,13 @@ public function onBeforeDelete()
         parent::onBeforeDelete();
     }
 
+    public function onBeforeWrite()
+    {
+        parent::onBeforeWrite();
+
+        $this->Title = $this->getField('Name');
```
robbieaverill: So you can still manually change the Title field because this is in a beforeWrite right?

robbieaverill:
Nice branch name

PR #170, opened 2018-10-08
BUG Fix migrating files with an incorrect class

lukereative:
Fixes silverstripe/silverstripe-framework#8312

### DIFF
--- a/src/FileMigrationHelper.php
+++ b/src/FileMigrationHelper.php
@@ -102,6 +102,17 @@ class FileMigrationHelper
             return false;
         }
 
+        // If this file has the wrong class for it's extension, update the classname
+        $destinationClass = $file->get_class_for_file_extension($extension);
+        if ($file->ClassName !== $destinationClass) {
+            // We disable validation so that we are able to write a corrected
+            // classname, once that is changed we re-enable it for subsequent writes
+            Config::modify()->set(File::class, 'validation_enabled', false);
+            $file = $file->newClassInstance($destinationClass);
+            $file->write();
+            Config::modify()->set(File::class, 'validation_enabled', true);
+        }
+
         // Copy local file into this filesystem
         $filename = $file->generateFilename();
         $result = $file->setFromLocalFile(
--- a/src/FileMigrationHelper.php
+++ b/src/FileMigrationHelper.php
@@ -9,6 +9,7 @@ use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\Versioned\Versioned;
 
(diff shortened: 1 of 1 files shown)


src/FileMigrationHelper.php
```diff
@@ -102,6 +102,17 @@ protected function migrateFile($base, File $file, $legacyFilename)
...
 
+        // If this file has the wrong class for it's extension, update the classname
+        $destinationClass = $file->get_class_for_file_extension($extension);
+        if ($file->ClassName !== $destinationClass) {
+            // We disable validation so that we are able to write a corrected
+            // classname, once that is changed we re-enable it for subsequent writes
+            Config::modify()->set(File::class, 'validation_enabled', false);
+            $file = $file->newClassInstance($destinationClass);
+            $file->write();
+            Config::modify()->set(File::class, 'validation_enabled', true);
```
robbieaverill: I _think_ that doing this will keep a nested config manifest around, whereas if you explicitly nest it then change the config prop, then unnest it it will remove the nested manifest for the rest of the execution. Thoughts?

lukereative: I tried nesting it, but it doesn't work when I modify it that way.

src/FileMigrationHelper.php
```diff
@@ -102,6 +102,17 @@ protected function migrateFile($base, File $file, $legacyFilename)
...
         }
 
+        // If this file has the wrong class for it's extension, update the classname
+        $destinationClass = $file->get_class_for_file_extension($extension);
+        if ($file->ClassName !== $destinationClass) {
+            // We disable validation so that we are able to write a corrected
+            // classname, once that is changed we re-enable it for subsequent writes
+            Config::modify()->set(File::class, 'validation_enabled', false);
+            $file = $file->newClassInstance($destinationClass);
+            $file->write();
```
maxime-rainville: This line is still throwing a `[Emergency] Uncaught SilverStripe\ORM\ValidationException: Extension 'pdf' is not allowed` error for me despite disabling validation.

src/FileMigrationHelper.php
```diff
@@ -102,6 +102,17 @@ protected function migrateFile($base, File $file, $legacyFilename)
             return false;
         }
 
+        // If this file has the wrong class for it's extension, update the classname
+        $destinationClass = $file->get_class_for_file_extension($extension);
+        if ($file->ClassName !== $destinationClass) {
+            // We disable validation so that we are able to write a corrected
+            // classname, once that is changed we re-enable it for subsequent writes
+            Config::modify()->set(File::class, 'validation_enabled', false);
```
maxime-rainville: We probably should read the value of `validation_enabled` before setting it `false`. When we're done doing things that require validation to be false, then we reset it to its original value.

The way it's done right now, if someone decides to disable this validation, we'll end up enabling it for them by accident.

src/FileMigrationHelper.php
```diff
@@ -102,6 +102,17 @@ protected function migrateFile($base, File $file, $legacyFilename)
             return false;
         }
 
+        // If this file has the wrong class for it's extension, update the classname
+        $destinationClass = $file->get_class_for_file_extension($extension);
```
maxime-rainville: I would have move this block of code to a private function just for readability.

maxime-rainville: I'm not sure there's a elegant way of unit testing this. And it's affecting a part of code that's probably not going to be used very much. So not having unit tests for this is probably kosher.

lukereative: Moved and yea, I don't think a unit test is necessary.

src/FileMigrationHelper.php
lukereative: Done

src/FileMigrationHelper.php
lukereative: Fixed, it worked if you ran it multiple times which is why I didn't pick it up. It now get's the updated file before continuing.

maxime-rainville:
Looks good ... for future reference I met with a bunch of guys on Friday that explained that the mis-classified files usually occur when you upgrade a site from SS2 to SS3.

PR #171, opened 2018-10-15
FIX: allow base path of /

sminnee:
A base path of / was sanitised to "", which is illegal.

Fixes https://github.com/silverstripe/silverstripe-framework/issues/1586

Although I don't have a chroot environment in which to test this, a review of the code suggests that this is the only place it's broken.

Note: if there's reluctance to merge this I'd suggest closing the referenced bug as wontfix. I've submitted this principally for the purpose of reducing the number of open bug ticktes.

### DIFF
--- a/src/FileFinder.php
+++ b/src/FileFinder.php
@@ -136,7 +136,7 @@ class FileFinder
     {
         $base = rtrim($base, '/');
         // Special case for a base path of /, eg. a chroot environment
-        if ($base == '') {
+        if ($base === '') {
             $base = '/';
         }
 
--- a/src/FileFinder.php
+++ b/src/FileFinder.php
@@ -134,7 +134,13 @@ class FileFinder
      */
     public function find($base)
     {
-        $paths = array(array(rtrim($base, '/'), 0));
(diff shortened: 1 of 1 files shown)


src/FileFinder.php
```diff
@@ -134,7 +134,13 @@ public function setOptions(array $options)
      */
     public function find($base)
     {
-        $paths = array(array(rtrim($base, '/'), 0));
+        $base = rtrim($base, '/');
+        // Special case for a base path of /, eg. a chroot environment
+        if ($base == '') {
```
robbieaverill: Super minor, but we should use strict comparison here

unclecheese:
Have retargeted to `1.3`

src/FileFinder.php
robbieaverill:
```suggestion
        if ($base === '') {
```

sminnee:
Would recommend putting this into 1, rather than 1.3, as I don't think it's important enough to be a late-stage change. This has mainly been done to clear out a historic archive of bugs.

PR #173, opened 2018-10-18
Ensure Versioned exists

### LINKED ISSUE #172: FileLinkTracking should test for Versioned class
I have hit a stalemate whereby the `SilverStripe\Assets\Shortcodes\FileLinkTracking` DataExtension physically requires `SilverStripe\Versioned\Versioned` to exist.

In this implementation (which basically consists of just the framework & assets modules installed) I do not use Versioned, or actually even want `silverstripe/versioned` installed, which causes `FileLinkTracking:augmentSyncLinkTracking()` to fail on:
```
if (Versioned::get_stage() == Versioned::LIVE && $this->owner->hasExtension(Versioned::class))
```
This will fail on `FileLinkTracking:onAfterDelete()` too.

I don't need FileLinkTracking nor want Versioned (for this implementation only), so if you have any suggestions (other than "don't upgrade") I'm all ears. I don't believe the answer here is requiring `silverstripe/versioned` for `silverstripe/assets`, so I believe `FileLinkTracking` should at least test if Versioned is even exists in the `augmentSyncLinkTracking` / `onAfterDelete` logic.

axllent:
Ensure `class_exists(Versioned::class)` exists before calling `Versioned::get_stage()` on `augmentSyncLinkTracking()` and `onAfterDelete()`  - resolves #172

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -650,8 +650,11 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     {
         // Force query of draft object and update (as source record is bound to live stage)
         /** @var File $draftRecord */
-        $draftRecord = Versioned::get_by_stage(self::class, Versioned::DRAFT)->byID($this->ID);
-        $draftRecord->updateDependantObjects();
+        if (class_exists(Versioned::class) &&
+            $draftRecord = Versioned::get_by_stage(self::class, Versioned::DRAFT)->byID($this->ID)
+        ) {
+            $draftRecord->updateDependantObjects();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


axllent:
No worries. I saw those too but figured they would only be called (automatically) if Versioned was in use, but best to be safe. I've made those changes you requested (I'll let you squash on merge), though I'm stumped as to which branch you actually _want_ it against. I'll leave it as-is for now unless you want me to change it (if so then name one) - whichever makes it to release soonest :) `1` is the default.

robbieaverill:
I've rebased it onto 1.2, then we can merge up from there

PR #174, opened 2018-10-19
Change block invalid file extensions regex

shot131:
Need to change regex  for allowing files with multiple dots like image.1.2.jpg
Now it is access denied.


robbieaverill:
Hi @shot131, thanks for contributing! We've fixed this already with https://github.com/silverstripe/silverstripe-assets/pull/158, I'll make sure the branches are merged up to 1. You can expect the patch to be released in the next version (in the next couple of weeks).

robbieaverill:
It looks like this was already merged up. Does the patch at #158 not work for you?

shot131:
Yes, its work. Thanks.

PR #189, opened 2018-11-08
update ini2Bytes

wernerkrauss:
fixes broken test due to changes in `Convert::memstring2Bytes`

See https://github.com/silverstripe/silverstripe-framework/pull/8583

reflecting deprecation in docblock


robbieaverill:
Merged up to `1`

robbieaverill:
Hey @wernerkrauss, I just noticed that the deprecation notice for this is saying "this will be deprecated from 5.0 onwards" - is that what you had intended? I thought maybe you wanted it to be deprecated from 4.4 and removed in 5.0, or something like that

robbieaverill:
Ah yes, of course - thanks =D

PR #190, opened 2018-11-08
MINOR: Change expected type to int, per framework PR change

unclecheese:
This gets the tests passing due to the API change (!!!) in https://github.com/silverstripe/silverstripe-framework/commit/3f321f935a0b7faa3d4ef788b4f69425db979095

A good reminder that even small API changes like that can break stuff silently.


robbieaverill:
@kinglozzer good point. That PR has more to it than this one, so let's do that

PR #206, opened 2019-01-18
FIX: Add missing file upload error types (fixes #205)

### LINKED ISSUE #205: Upload_Validator doesn’t handle UPLOAD_ERR_PARTIAL errors
I don’t know how to accurately reproduce this, but I keep getting bugged by Sentry with errors:

`Call to a member function isInDB() on null`
https://github.com/silverstripe/silverstripe-asset-admin/blob/1.3.0/code/Forms/UploadField.php#L162

The HTTP POST body looks like:

```
{
  "SecurityID": "x",
  "Upload": {
    "size": 0,
    "type": "",
    "name": "pic.jpg",
    "tmp_name": "",
    "error": 3
  },
  "ParentID": "x"
}
```

That error code `3` is for when the file was only partially uploaded: http://php.net/manual/en/features.file-upload.errors.php. I have no idea how that happens, but there ya go.

[This line](https://github.com/silverstripe/silverstripe-asset-admin/blob/1.3.0/code/Forms/UploadField.php#L146) should populate the `$error` variable (pass-by-ref) if something has gone wrong and prevent the fatal error, but it’s failing to do that because `Upload_Validator` doesn’t handle “partially uploaded” errors. In fact, I think it’s missing:

- `UPLOAD_ERR_PARTIAL`
- `UPLOAD_ERR_NO_TMP_DIR`
- `UPLOAD_ERR_CANT_WRITE`
- `UPLOAD_ERR_EXTENSION`

kinglozzer:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/205

Error types reference: http://php.net/manual/en/features.file-upload.errors.php

I added `UPLOAD_ERR_NO_TMP_DIR` and `UPLOAD_ERR_CANT_WRITE` to the generic “not a valid upload” error as I think they’re sufficiently rare to not need their own message: if you don’t have a tmp directory, or can’t write to it, you’re probably going to have bigger fish to fry...

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -21,6 +21,7 @@ en:
     MpgType: 'MPEG video file'
     NOFILESIZE: 'Filesize is zero bytes.'
     NOVALIDUPLOAD: 'File is not a valid upload'
+    PARTIALUPLOAD: 'File did not finish uploading, please try again'
     PATH: Path
     PLURALNAME: Files
     PLURALS:
--- a/src/Upload_Validator.php
+++ b/src/Upload_Validator.php
@@ -276,6 +276,14 @@ class Upload_Validator
             return false;
         }
 
+        if (!$this->isCompleteUpload()) {
+            $this->errors[] = _t(
+                'SilverStripe\\Assets\\File.PARTIALUPLOAD',
+                'File did not finish uploading, please try again'
+            );
+            return false;
+        }
+
         // Check file isn't empty
         if ($this->isFileEmpty()) {
             $this->errors[] = _t('SilverStripe\\Assets\\File.NOFILESIZE', 'Filesize is zero bytes.');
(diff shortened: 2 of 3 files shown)


kinglozzer:
> I would rather have a somewhat more generic message however. "File was only partially uploaded" is probably going to confuse most users.

Yeah, I wanted to keep it concise like the other ones but it’s probably not too helpful. How about something like:

- File was only partially uploaded, perhaps due to a connection interruption
- File was only partially uploaded, please try again
- File did not finish uploading, please try again

maxime-rainville:
"File did not finish uploading, please try again" sounds like the best one from a regular user perspective. 

If there was some way to get the detail technical message printed out to the sever logs or JS console, that would be a nice-to-have, but not essential.

kinglozzer:
I’ve updated the message, anything else? :)

PR #218, opened 2019-03-08
BUG Redirect legacy file name and older hash url to latest live URL

maxime-rainville:
This PR rewrites the `FlysystemAssetStore::getResponseFor()` method so that SS3 style url and the hash-url of old publish URL can get redirected to the latest live version. This works for SS4 variants as well.

# Limitations
* Hash URL don't get redirected when using `legacy_filenames`.
* Redirects do not work when the Filename is change in the CMS or when a file is moved to a different folder.

# Parent issue

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -132,6 +132,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         "Owner" => Member::class,
     );
 
+    private static $indexes = array(
+        'FileHash' => true
+    );
+
     private static $defaults = array(
         "ShowInSearch" => 1,
     );
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -815,7 +815,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
     private function parseLegacyFileID($fileID)
     {
         // assets/folder/_resampled/ResizedImageWzEwMCwxMzNd/basename.extension
-        $ss3Pattern = '#^(?<folder>([^/_]+/)*?)(_resampled/(?<variant>([^/.]+))/)?((?<basename>[^/._]+))(?<extension>(\..+)*)$#';
+        $ss3Pattern = '#^(?<folder>([^/_]+/)*?)(_resampled/(?<variant>([^/.]+))/)?((?<basename>((?<!__)[^/.])+))(?<extension>(\..+)*)$#';
         // assets/folder/basename__ResizedImageWzEwMCwxMzNd.extension
         $ss4LegacyPattern = '#^(?<folder>([^/]+/)*)(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
 
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -9,6 +9,7 @@ use League\Flysystem\Exception;
 use League\Flysystem\Filesystem;
 use League\Flysystem\Util;
 use LogicException;
+use SilverStripe\Assets\File;
 use SilverStripe\Assets\Storage\AssetNameGenerator;
 use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Assets\Storage\AssetStoreRouter;
@@ -19,6 +20,8 @@ use SilverStripe\Control\HTTPStreamResponse;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\ORM\DB;
+use SilverStripe\Versioned\Versioned;
 
 /**
  * Asset store based on flysystem Filesystem as a backend
@@ -80,6 +83,16 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
      * @var int
      */
     private static $missing_response_code = 404;
+
+
+    /**
+     * Define the HTTP Response code for request that should be redirected to a different URL. Defaults to a temporary
+     * redirection (307). Set to 308 if you would rather your redirections be permanent and indicate to search engine
+     * that they should index the other file.
+     * @config
+     * @var int
+     */
+    private static $redirect_response_code = 307;
 
     /**
      * Custom headers to add to all custom file responses
@@ -770,9 +783,9 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
      * @param string $fileID
      * @return array
      */
-    protected function parseFileID($fileID)
+    protected function parseFileID($fileID, $forceLegacyName = false)
     {
-        if ($this->useLegacyFilenames()) {
+        if ($forceLegacyName || $this->useLegacyFilenames()) {
             $pattern = '#^(?<folder>([^/]+/)*)(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
         } else {
             $pattern = '#^(?<folder>([^/]+/)*)(?<hash>[a-zA-Z0-9]{10})/(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
@@ -785,9 +798,11 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
 
         $filename = $matches['folder'] . $matches['basename'] . $matches['extension'];
         $variant = isset($matches['variant']) ? $matches['variant'] : null;
+        $hash = isset($matches['hash']) ? $matches['hash'] : null;
         return [
             'Filename' => $filename,
             'Variant' => $variant,
+            'Hash' => $hash
         ];
     }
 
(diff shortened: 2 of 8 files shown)


src/Flysystem/FlysystemAssetStore.php
```diff
@@ -785,9 +798,37 @@ protected function parseFileID($fileID)
...
+        ];
+    }
+
+    /**
+     * Try to parse a file ID using the old SilverStripe 3 format legacy or the SS4 legacy filename format.
+     *
+     * @param string $fileID
+     * @return array
+     */
+    private function parseLegacyFileID($fileID)
```
maxime-rainville: Originally, I had this as part of `parseFileID` with an extra `$forceLegacy` argument. Decided to make it its own method to avoid potentially breaking a class that would extend `FlysystemAssetStore` and override `parseFileID`. Also the addition of the SS3 file pattern made this use case sufficiently different to warrant a separate method.

I made the method private because, I presume I'll be potentially rewriting this in the SS44 implementation and I don't want to introduce new API to support in a patch release.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -913,24 +954,94 @@ public static function flush()
...
-        // Serve up file response
-        return $this->createResponseFor($filesystem, $asset);
+        // We've looked everywhere and couldn't find a file
+        return $this->createMissingResponse();
+    }
+
+    /**
+     * @return string
+     */
+    private function searchForEquivalentFileID($asset)
```
maxime-rainville: Made this private because I don't want to introduce new APIs in a patch release and limit my option for the SS44 fix.

maxime-rainville:
A few notes on the unit tests. I didn't write any test for regular scenarios like accessing: 
* a file directly, 
* accessing a regular variant,
* accessing a file that doesn't exist,
* accessing a protected file.

Those scenarios should already be tested by the existing tests.

You'll find that there's 2 new unit test class:
* `RedirectFileControllerTest`
* `RedirectKeepArchiveFileControllerTest` which extends `RedirectFileControllerTest`

`RedirectKeepArchiveFileControllerTest` is meant to run the same tests as `RedirectFileControllerTest` only with the `keep archive files` flag enable. The expected results are all the same except for one method which was overridden.

maxime-rainville:
@bergice Just set up nginx on my local box (such a lovely time) and configured my environment using the instruction in our nginx doc. Redirect work like a charm.

tests/php/RedirectFileControllerTest.php
```diff
@@ -0,0 +1,453 @@
...
+            '',
+            $file->getURL(false),
+            $response,
+            'Legacy URL for published file should return 307'
+        );
+
+        $response = $this->get($response->getHeader('location'));
+        $this->assertResponse(
+            200,
+            str_repeat('x', 1000000),
```
chillu: This is a bit noisy - could it just be `version 1` as content? Or would that mess with the test setup?

tests/php/RedirectFileControllerTest.php
```diff
@@ -0,0 +1,453 @@
...
+        $this->assertResponse(
+            307,
+            '',
+            $icoV2Url,
+            $response,
+            'Old URL to variant should redirect with 307'
+        );
+    }
+
+    public function testVariantInFolderRedirect()
```
chillu: Same here, please test with multiple folder depth and underscores in file/folder name

tests/php/Storage/AssetStoreTest.php
```diff
@@ -278,55 +278,55 @@ public function dataProviderFileIDs()
                 'directory/2a17a9cb4b/file.jpg',
                 [
                     'Filename' => 'directory/file.jpg',
-                    'Hash' => sha1('puppies'),
+                    'Hash' => substr(sha1('puppies'), 0, 10),
```
chillu: Why did this change from full sha to shortened?

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -785,9 +798,37 @@ protected function parseFileID($fileID)
...
+    /**
+     * Try to parse a file ID using the old SilverStripe 3 format legacy or the SS4 legacy filename format.
+     *
+     * @param string $fileID
+     * @return array
+     */
+    private function parseLegacyFileID($fileID)
+    {
+        // assets/folder/_resampled/ResizedImageWzEwMCwxMzNd/basename.extension
+        $ss3Pattern = '#^(?<folder>([^/_]+/)*?)(_resampled/(?<variant>([^/.]+))/)?((?<basename>((?<!__)[^/.])+))(?<extension>(\..+)*)$#';
```
chillu: How much did you prod and poke at those regexes for edge cases?

src/File.php
```diff
@@ -132,6 +132,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         "Owner" => Member::class,
     );
 
+    private static $indexes = array(
+        'FileHash' => true
```
chillu: This can be a potentially expensive `dev/build` depending on how many files you have. I'm not too concerned, but wanted to point it out.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -913,24 +954,94 @@ public static function flush()
...
+        $public = $this->getPublicFilesystem();
+        $protected = $this->getProtectedFilesystem();
+
+        // If the file exists on the public store, we just straight return it.
+        if ($public->has($asset)) {
+            return $this->createResponseFor($public, $asset);
         }
 
-        // Block directory access
-        if ($filesystem->get($asset) instanceof Directory) {
```
chillu: Why is this check no longer required? If you look at the implementation of `Local->has()`, it just does a `file_exists()`, which returns true for both files and folders.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -913,24 +954,94 @@ public static function flush()
...
         }
 
-        // Serve up file response
-        return $this->createResponseFor($filesystem, $asset);
+        // We've looked everywhere and couldn't find a file
+        return $this->createMissingResponse();
+    }
+
+    /**
+     * @return string
```
chillu: Needs some PHPDoc, even if the method is private - it contains some non-trivial logic which needs a summary

chillu: Also, missing the method parameter or type in PHPDoc

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -913,24 +954,94 @@ public static function flush()
...
+        if ($parsedFileID && $parsedFileID['Hash']) {
+            // Try to find a live version of this file
+            $stage = Versioned::get_stage();
+            Versioned::set_stage(Versioned::LIVE);
+            $file = File::get()->filter(['FileFilename' => $parsedFileID['Filename']])->first();
+            Versioned::set_stage($stage);
+
+            // If we found a matching live file, let's see if our hash was publish at any point
+            if ($file) {
+                $oldVersionCount = $file->allVersions(
```
chillu: Could we shortcut this and only run the check if the *current* live version (`$file`) doesn't have this `FileHash`? That'll be the most common scenario, right?

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -785,9 +798,37 @@ protected function parseFileID($fileID)
...
+
+    /**
+     * Try to parse a file ID using the old SilverStripe 3 format legacy or the SS4 legacy filename format.
+     *
+     * @param string $fileID
+     * @return array
+     */
+    private function parseLegacyFileID($fileID)
+    {
+        // assets/folder/_resampled/ResizedImageWzEwMCwxMzNd/basename.extension
```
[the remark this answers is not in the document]
maxime-rainville: All the URLs my local ss3 install spit out look like this:
`http://ss3x.max.silverstripe.com/assets/_resampled/FillWzUwLDUwXQ/sam.jpg`

The variant is always it's own folder. Is there other variant format in SS3?

tests/php/RedirectFileControllerTest.php
maxime-rainville: That's copied over from some other test. I can change it to `version 1`.

tests/php/Storage/AssetStoreTest.php
maxime-rainville: The original hash wasn't being use in the test, so it didn't matter what the hash was. My test needs a sensible hash, so I had to normalise it to what will actually be used in production

maxime-rainville: FYI there's probably some super edge casy scenario where we end up with a collision when someone uploads a file whose sha1 hash has the same 10 character as some other file.

src/File.php
maxime-rainville: Yes, that's a downside, but it speeds the other end when we search file by hash.

maxime-rainville: FYI I confirm that MySQL will use the index for SQL expression like `HASH like 'ABCDEFGHIJ%'`.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: The check has been moved to `createResponseFor`? With the way I've reorganised `getResponseFor()` it made a bit more sense to have it there.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: That wouldn't make a difference. If you try to access the current live file using the regular hash URL, you will have tripped this condition in `getResponseFor()` and you will never reach this point:
```
        if ($public->has($asset)) {
            return $this->createResponseFor($public, $asset);
        }
```

src/Flysystem/FlysystemAssetStore.php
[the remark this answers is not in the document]
chillu: WTF, you're right. I must've imagined that :/

src/Flysystem/FlysystemAssetStore.php
chillu: True, so it doesn't have to get the asset twice from the asset store (once in `getResponseFor()`, then again in `createResponseFor()`. It does mean that it'll check `isGranted()` for protected folders though. I guess that's pretty low impact.

tests/php/RedirectFileControllerTest.php
chillu: Yeah you're building more complexity on that baseline, so I think it'd be appropriate to change

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -81,6 +84,16 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
...
 
+
+    /**
+     * Define the HTTP Response code for request that should be redirected to a different URL. Defaults to a temporary
+     * redirection (307). Set to 308 if you would rather your redirections be permanent and indicate to search engine
+     * that they should index the other file.
+     * @config
+     * @var int
+     */
+    private static $redirect_response_code = 307;
```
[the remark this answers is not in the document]
maxime-rainville: Done.

tests/php/RedirectFileControllerTest.php
```diff
@@ -0,0 +1,453 @@
...
+    {
+        TestAssetStore::reset();
+        parent::tearDown();
+    }
+
+    public function testLegacyFilenameRedirect()
+    {
+        $file = File::find('FileTest-subfolder/FileTestSubfolder.txt');
+
+        $response = $this->get('/assets/FileTest-subfolder/FileTestSubfolder.txt');
```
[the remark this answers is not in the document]
maxime-rainville: Add a bunch of new test

tests/php/RedirectFileControllerTest.php
maxime-rainville: Change to `version 1`

tests/php/RedirectFileControllerTest.php
maxime-rainville: The tests in this method are now covered by the data provider in `testVariantRedirect`. So I removed the method.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: Added a bunch of extra unit tests.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: Fixed.

maxime-rainville:
All of @chillu points have been addressed.

PR #223, opened 2019-03-14
Default to using naturl path for accessing files

maxime-rainville:
# Change to how paths works
This PR allows the protected and public file store to use different file format. The two principal file format are:
* Natural path (`actual/path/shown/in/the/cms.txt`) which is equivalent to the path used when _legacy file names_ is enabled.
* Hash path (`some/folder/abcdf7890/myfile.txt`) which is equivalent to the default path used in SilverStripe 4.0 to 4.3.

Hash paths provide a way for multiple versions of the same file to co-exist which is needed for protected files. Natural paths are easier to understand and persistent across versions which pleases search engines and smelly humans.

The public store now defaults to using natural path, while the protected store still uses hash paths.

# Redirection
Just like the recent change  to SS4.3, each store will try to resolve paths that are in a different format to a matching file in a different format. This will allow sites that are using the old format to upgrade to the newer one progressively, without the need to immediately run a file migration task.

# How is this magic handled
Three new constructs have been added `FileIDHelper`, `FileResolutionStrategy` and `ParsedFileID`

In this context, a file ID is a string pointing to a file relative to the root of a file store. It's a file path by another name.

## FileIDHelper

Each class implementing the `FileIDHelper` interface represent a file format and provide logic to decode and encode file IDs. FileIDHelpers are glorified regular expression boxes. They don't look directly at files ... only at FileID strings.

3 FileIDHelpers are povided:
* `NaturalFileIDHelper` which defines natural paths ;
* `HashFileIDHelper` which defines hash paths ;
* `LegacyFileIDHelper` which is very similar `NaturalFileIDHelper`, but can understand SS3 variant fileID (e.g. `my-folder/_resampled/ResizeXYZ/image.jpg`)

In theory, a developer who wanted to use their own custom file ID format could create their own `FileIDHelper`. Or someone who wanted to keep using the SS3 format could define `LegacyFileIDHelper` as their default. Neither of those scenarios have been tested however and are not officially supported.

## FileResolutionStrategy

A `FileResolutionStrategy` is responsible for trying to understand a FileID provided in an unknown format and find an existing file in a potentially different format. This bridges the gap between FileIDHelpers and the file system.

A `FileResolutionStrategy` isn't associated to a specific Filesystem. Instead most method on the `FileResolutionStrategy` expect to be provided a `Filesystem` reference.

There's only one implementation of `FileResolutionStrategy` at this moment: `FileIDHelperResolutionStrategy`. This helper expects to be provided a default FileIDHelper which will be used to generate new FileIDs for new files and a list of resolution FileIDHelpers which will be used to try to resolve request for existing files.

`FileIDHelperResolutionStrategy` will also attempt to read information from the DB to find information about files and redirect users to newer version of past files. `FileIDHelperResolutionStrategy` accepts a Versioned stage to control on what stages we'll look for files.

The Protected store and public store will both have their own unique `FileResolutionStrategy`.

## ParsedFileID

A `ParseFileID` is a more sophisticated equivalent of a file tuple array that gets passed around between AssetStore, the FileResolutionStrategy and the FileIDHelper. It has the following properties:
* Filename
* Hash
* Variant
* FileID

It's just a convenient way to encapsulate some File identification data. `ParseFileID` is immutable but it has some setter methods. When calling those setter methods, you will get a slightly a new copy with some slightly altered data. 

# What else has change

The bulk of the other change have occurred on `FlysystemAssetStore`. Most of the changes involves figuring out on what store an operation should occur and calling the right combination of Filesystem and ResolutionStrategy. 

That magic occurs in `applyToFileOnFilesystem()`. This method receives some file information and will be looking for a matching physical file with these steps:
* an exact match on the public store, 
* an exact match on the protected store,
* a resolution match on the public store,
* a resolution match on the protected store.

Once `applyToFileOnFilesystem` has manage to resolve a file, it will call a closure with a matching ParsedFileID, a Filesystem, a FileResolutionStrategy and a visibility flag.

# Other changes
* There's 3 methods that can be use to write files on the AssetStore. They use to be completely separate. `setFromLocalFile` and `setFromString` now wrap their data around a stream resource and call `setFromStream`.
* A bunch of methods have been marked as `protected` methods on FlysystemAssetStore have been marked as deprecated. They usually have an equivalent private method. I don't want to remove them because some people may be relying on them in their own custom asset store, but it's annoying as hell that we have to keep them around.
* There's a new `swapPublish` method on `FlysystemAssetStore`. I didn't want to make it part of the interface to avoid breaking people code. That methods stashes existing publish files to avoid overriding them.
* Following discussion with @chillu, `legacy_filenames` is now compeltly ignored and doesn't do anything.
* `FlysystemAssetStore` was throwing `League\Flysystem\Exception` in many places. This was a bit confusing because people could have read this to mean that we were throwing native PHP `\Exception`. I've aliased `League\Flysystem\Exception` to `FlysystemException` for clarity.

# Testing this crazy thing

When testing this locally, you probably want to remove this line from the `.htaccess` file in your assets folder:  `RewriteCond %{REQUEST_FILENAME} !-f`. Otherwise a big chunk of your request will never it PHP.

Another thing to keep in mind is that his new approach does a lot more file existence check and file reads (to build hashes) than the previous ones. There's not really a way around it because of the nature of the change, but we do want to make sure we don't affect serving time to much.

## Things to pay special attention to
* The deprecated protected methods on FlysystemAssetStore  don't have full unit test coverage and they are not indirectly call by other methods any more.
* The order in which FileIDHelpers are added to a FileIDHelperResolution strategy is very important. `NaturalFileIDHelper` adn `LegacyFileIDHelper` will interpret the hash in a Hash path as folder name. `NaturalFileIDHelper` adn `LegacyFileIDHelper` are only different when looking at variants.

# Important things left to do
* Update the file migration script
* Run the asset-admin unit test and behat test on this crazy thing
* Run the kitchen sink test against this crazy thing

# Parent issue

### DIFF
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -130,12 +130,6 @@ class AssetControlExtension extends DataExtension
         // When deleting from stage then check if we should archive assets
         $archive = $this->owner->config()->get('keep_archived_assets');
 
-        // Publish assets
-        $this->publishAll($manipulations->getPublicAssets());
-
-        // Protect assets
-        $this->protectAll($manipulations->getProtectedAssets());
-
         // Check deletion policy
         $deletedAssets = $manipulations->getDeletedAssets();
         if ($archive && $this->isVersioned()) {
@@ -145,6 +139,11 @@ class AssetControlExtension extends DataExtension
             // Otherwise remove all assets
             $this->deleteAll($deletedAssets);
         }
+
+        // Publish assets
+        $this->publishAll($manipulations->getPublicAssets());
+        // Protect assets
+        $this->protectAll($manipulations->getProtectedAssets());
     }
 
     /**
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -133,15 +133,17 @@ class AssetControlExtension extends DataExtension
         // Check deletion policy
         $deletedAssets = $manipulations->getDeletedAssets();
         if ($archive && $this->isVersioned()) {
+            // Publish assets
+            $this->swapAll($manipulations->getPublicAssets());
             // Archived assets are kept protected
             $this->protectAll($deletedAssets);
         } else {
+            // Publish assets
+            $this->publishAll($manipulations->getPublicAssets());
             // Otherwise remove all assets
             $this->deleteAll($deletedAssets);
         }
 
-        // Publish assets
-        $this->publishAll($manipulations->getPublicAssets());
         // Protect assets
         $this->protectAll($manipulations->getProtectedAssets());
     }
@@ -264,6 +266,32 @@ class AssetControlExtension extends DataExtension
         $store = $this->getAssetStore();
         foreach ($assets as $asset) {
             $store->delete($asset['Filename'], $asset['Hash']);
+        }
+    }
+
+    /**
+     * Move all assets in the list to the public store
+     *
+     * @param array $assets
+     */
+    protected function swapAll($assets)
+    {
+        if (empty($assets)) {
+            return;
+        }
+
+        $store = $this->getAssetStore();
+
+        // The `swap` method was introduced in the 1.4 release. It wasn't added to the interface to avoid breaking
+        // custom implementation. If it's not available on our store, we fallback to a publish/protect
+        if (method_exists($store, 'swapPublish')) {
+            foreach ($assets as $asset) {
+                $store->swapPublish($asset['Filename'], $asset['Hash']);
+            }
+        } else {
+            foreach ($assets as $asset) {
+                $store->publish($asset['Filename'], $asset['Hash']);
+            }
         }
     }
 
--- a/src/Dev/TestAssetStore.php
+++ b/src/Dev/TestAssetStore.php
@@ -190,6 +190,8 @@ class TestAssetStore extends FlysystemAssetStore implements TestOnly
     protected function isSeekableStream($stream)
     {
         if (isset(self::$seekable_override)) {
+            // Unset the override so we don't get stuck in an efficit loop
+            self::$seekable_override = null;
             return self::$seekable_override;
         }
         return parent::isSeekableStream($stream);
--- a/src/Dev/TestAssetStore.php
+++ b/src/Dev/TestAssetStore.php
@@ -177,6 +177,11 @@ class TestAssetStore extends FlysystemAssetStore implements TestOnly
         return parent::getOriginalFilename($fileID);
     }
 
+    public function getFilesystemFor($fileID)
+    {
+        return parent::getFilesystemFor($fileID);
+    }
+
     public function removeVariant($fileID)
     {
         return parent::removeVariant($fileID);
--- a/src/Dev/TestAssetStore.php
+++ b/src/Dev/TestAssetStore.php
@@ -5,6 +5,7 @@ namespace SilverStripe\Assets\Dev;
 use League\Flysystem\Adapter\Local;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Filesystem;
+use SilverStripe\Assets\FilenameParsing\FileResolutionStrategy;
 use SilverStripe\Assets\Filesystem as SSFilesystem;
 use SilverStripe\Assets\Flysystem\FlysystemAssetStore;
 use SilverStripe\Assets\Flysystem\ProtectedAssetAdapter;
@@ -76,6 +77,8 @@ class TestAssetStore extends FlysystemAssetStore implements TestOnly
         $backend = new TestAssetStore();
         $backend->setPublicFilesystem($publicFilesystem);
         $backend->setProtectedFilesystem($protectedFilesystem);
+        $backend->setPublicResolutionStrategy(Injector::inst()->get(FileResolutionStrategy::class . '.public'));
+        $backend->setProtectedResolutionStrategy(Injector::inst()->get(FileResolutionStrategy::class . '.protected'));
         Injector::inst()->registerService($backend, AssetStore::class);
         Injector::inst()->registerService($backend, AssetStoreRouter::class);
 
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -34,6 +34,4 @@ interface FileIDHelper
      * @return ParsedFileID|null
      */
     public function parseFileID($fileID);
-
-
 }
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -11,12 +11,12 @@ interface FileIDHelper
     /**
      * Map file tuple (hash, name, variant) to a filename to be used by flysystem
      *
-     * @param string $filename Name of file
+     * @param string|ParsedFileID $filename Name of file or ParsedFileID object
      * @param string $hash Hash of original file
      * @param string $variant (if given)
      * @return string Adapter specific identifier for this file/version
      */
-    public function buildFileID($filename, $hash, $variant = null);
+    public function buildFileID($filename, $hash = null, $variant = null);
 
 
     /**
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -2,6 +2,9 @@
 
 namespace SilverStripe\Assets\FilenameParsing;
 
+/**
+ * Helps build and parse FileIDs according to a predefined format.
+ */
 interface FileIDHelper
 {
 
@@ -25,10 +28,10 @@ interface FileIDHelper
     public function cleanFilename($filename);
 
     /**
-     * Get Filename, Variant and Hash from a file id
+     * Get Filename, Variant and Hash from a fileID. If a FileID can not be parsed, returns `null`.
      *
      * @param string $fileID
-     * @return ParsedFileID
+     * @return ParsedFileID|null
      */
     public function parseFileID($fileID);
 
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+interface FileIDHelper
+{
+
+    /**
+     * Map file tuple (hash, name, variant) to a filename to be used by flysystem
+     *
+     * @param string $filename Name of file
+     * @param string $hash Hash of original file
+     * @param string $variant (if given)
+     * @return string Adapter specific identifier for this file/version
+     */
+    public function buildFileID($filename, $hash, $variant = null);
+
+
+    /**
+     * Performs filename cleanup before sending it back.
+     *
+     * @param string $filename
+     * @return string
+     */
+    public function cleanFilename($filename);
+
+    /**
+     * Get Filename, Variant and Hash from a file id
+     *
+     * @param string $fileID
+     * @return ParsedFileID
+     */
+    public function parseFileID($fileID);
+
+
+}
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -34,4 +34,20 @@ interface FileIDHelper
      * @return ParsedFileID|null
      */
     public function parseFileID($fileID);
+
+    /**
+     * Determine if the provided fileID is a variant of `$parsedFileID`.
+     * @param string $fileID
+     * @param ParsedFileID $parsedFileID
+     * @return boolean
+     */
+    public function isVariantOf($fileID, ParsedFileID $parsedFileID);
+
+    /**
+     * Compute the relative path where variants of the provided parsed file ID are expected to be stored.
+     *
+     * @param ParsedFileID $parsedFileID
+     * @return string
+     */
+    public function lookForVariantIn(ParsedFileID $parsedFileID);
 }
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -3,7 +3,9 @@
 namespace SilverStripe\Assets\FilenameParsing;
 
 /**
- * Helps build and parse FileIDs according to a predefined format.
+ * Helps build and parse Filename Identifiers (ake: FileIDs) according to a predefined format.
+ *
+ * @internal This is still an evolving API. It may change in the next minor release.
  */
 interface FileIDHelper
 {
@@ -20,7 +22,7 @@ interface FileIDHelper
 
 
     /**
-     * Performs filename cleanup before sending it back.
+     * Clean up filename to remove constructs that might clash with the underlying path format of this FileIDHelper.
      *
      * @param string $filename
      * @return string
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -1,0 +1,213 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+use League\Flysystem\Filesystem;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Assets\File;
+use SilverStripe\ORM\DB;
+
+/**
+ * File resolution strategy that relies on a list of FileIDHelpers to find files.
+ * * `DefaultFileIDHelper` is the default helper use to generate new file ID.
+ * * `ResolutionFileIDHelpers` can contain a list of helpers that will be used to try to find existing file.
+ *
+ * This file resolution strategy can be helpfull when the approach to resolving files has changed over time and you need
+ * older file format to resolve.
+ *
+ * You may also provide a `VersionedStage` to only look at files that were published.
+ */
+class FileIDHelperResolutionStrategy implements FileResolutionStrategy
+{
+    /**
+     * The FileID helper that will be use to build FileID for this adapter.
+     * @var FileIDHelper
+     */
+    private $defaultFileIDHelper;
+
+    /**
+     * List of FileIDHelper that should be use to try to parse FileIDs on this adapter.
+     * @var FileIDHelper[]
+     */
+    private $resolutionFileIDHelpers;
+
+    /**
+     * Constrain this strategy to the a specific versioned stage.
+     * @var string
+     */
+    private $versionedStage = Versioned::DRAFT;
+
+    public function resolveFileID($fileID, Filesystem $adapter)
+    {
+        // If File is not versionable, let's bail
+        if (!class_exists(Versioned::class) || !File::has_extension(Versioned::class)) {
+            return null;
+        }
+
+        foreach ($this->resolutionFileIDHelpers as $fileIDHelper) {
+            $parsedFileID = $fileIDHelper->parseFileID($fileID);
+            if (!$parsedFileID) {
+                continue;
+            }
+
+            $hash = $parsedFileID->getHash();
+
+            $tuple = $hash ? $this->resolveWithHash($parsedFileID) : $this->resolveHashless($parsedFileID);
+            if ($tuple && $redirect = $this->searchForTuple($tuple, $adapter)) {
+                return $redirect;
+            }
+
+            // If our helper managed to parse the file id, but could not resolve to an actual physical file,
+            // there's nothing else we can do.
+            return null;
+        }
+    }
+
+    /**
+     * Try to find a DB reference for this parsed file ID. Return a file tuple if a equivalent file is found.
+     * @param ParsedFileID $parsedFileID
+     * @return array|null
+     */
+    private function resolveWithHash(ParsedFileID $parsedFileID)
+    {
+        // Try to find a version for a given stage
+        $stage = Versioned::get_stage();
+        Versioned::set_stage($this->getVersionedStage());
+        $file = File::get()->filter(['FileFilename' => $parsedFileID->getFilename()])->first();
+        Versioned::set_stage($stage);
+
+        // Could not find a valid file, let's bail.
+        if (!$file) {
+            return null;
+        }
+
+        // If we found a matching live file, let's see if our hash was publish at any point
+
+        // Build a version filter
+        $versionFilters = [
+            ['"FileHash" like ?' => DB::get_conn()->escapeString($parsedFileID->getHash()) . '%'],
+            ['not "FileHash" like ?' => DB::get_conn()->escapeString($file->getHash())],
+        ];
+        if ($this->getVersionedStage() == Versioned::LIVE) {
+            // If we a limited to the Live stage, let's only look at files that have bee published
+            $versionFilters['WasPublished'] = true;
+        }
+
+        $oldVersionCount = $file->allVersions($versionFilters, "", 1)->count();
+        // Our hash was published at some other stage
+        if ($oldVersionCount > 0) {
+            return [
+                'Filename' => $file->getFilename(),
+                'Hash' => $file->getHash(),
+                'Variant' => $parsedFileID->getVariant()
+            ];
+        }
+    }
+
+    /**
+     * Try to find a DB reference for this parsed file ID that doesn't have an hash. Return a file tuple if a
+     * equivalent file is found.
+     * @param ParsedFileID $parsedFileID
+     * @return array|null
+     */
+    private function resolveHashless(ParsedFileID $parsedFileID)
+    {
+        $filename = $parsedFileID->getFilename();
+        $variant = $parsedFileID->getVariant();
+        // Let's try to match the plain file name
+        $stage = Versioned::get_stage();
+        Versioned::set_stage($this->getVersionedStage());
+        $file = File::get()->filter(['FileFilename' => $filename])->first();
+        Versioned::set_stage($stage);
+
+        if ($file) {
+            return [
+                'Filename' => $filename,
+                'Hash' => $file->getHash(),
+                'Variant' => $variant
+            ];
+        }
+    }
+
+    /**
+     * Given a file tuple, try to find it on this adapter using one of the supported format.
+     * @param array $tuple
+     * @return null
+     */
+    public function searchForTuple($tuple, Filesystem $adapter)
+    {
+        // Pre-format our tuple
+        if (is_array($tuple)) {
+            $tupleData = $tuple;
+        } elseif ($tuple instanceof ParsedFileID) {
+            $tupleData = $tuple->getTuple();
+        } else {
+            throw new \InvalidArgumentException(
+                'AssetAdapter::hashForTuple expect $tuples to be an array or a ParsedFileID'
+            );
+        }
+        $filename = $tupleData['Filename'];
+        $hash = $tupleData['Hash'];
+        $variant = $tupleData['Variant'];
+
+
+        $helpers = $this->getResolutionFileIDHelpers();
+        array_unshift($helpers, $this->getDefaultFileIDHelper());
+
+        foreach ($helpers as $helper) {
+            $fileID = $helper->buildFileID($filename, $hash, $variant);
+            if ($adapter->has($fileID)) {
+                return $fileID;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * @return FileIDHelper
+     */
+    public function getDefaultFileIDHelper()
+    {
+        return $this->defaultFileIDHelper;
+    }
+
+    /**
+     * @param FileIDHelper $defaultFileIDHelper
+     */
+    public function setDefaultFileIDHelper($defaultFileIDHelper)
+    {
+        $this->defaultFileIDHelper = $defaultFileIDHelper;
+    }
+
+    /**
+     * @return FileIDHelper[]
+     */
+    public function getResolutionFileIDHelpers()
+    {
+        return $this->resolutionFileIDHelpers;
+    }
+
+    /**
+     * @param FileIDHelper[] $resolutionFileIDHelpers
+     */
+    public function setResolutionFileIDHelpers($resolutionFileIDHelpers)
+    {
+        $this->resolutionFileIDHelpers = $resolutionFileIDHelpers;
+    }
+
+    /**
+     * @return string
+     */
+    public function getVersionedStage()
+    {
+        return $this->versionedStage;
+    }
+
+    /**
+     * @param string $versionedStage
+     */
+    public function setVersionedStage($versionedStage)
+    {
+        $this->versionedStage = $versionedStage;
+    }
+}
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -53,7 +53,7 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
             $hash = $parsedFileID->getHash();
 
             $tuple = $hash ? $this->resolveWithHash($parsedFileID) : $this->resolveHashless($parsedFileID);
-            if ($tuple && $redirect = $this->searchForTuple($tuple, $adapter)) {
+            if ($tuple && $redirect = $this->searchForTuple($tuple, $adapter, false)) {
                 return $redirect;
             }
 
@@ -71,10 +71,11 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
     private function resolveWithHash(ParsedFileID $parsedFileID)
     {
         // Try to find a version for a given stage
-        $stage = Versioned::get_stage();
-        Versioned::set_stage($this->getVersionedStage());
-        $file = File::get()->filter(['FileFilename' => $parsedFileID->getFilename()])->first();
-        Versioned::set_stage($stage);
+        /** @var File $file */
+        $file = Versioned::withVersionedMode(function () use ($parsedFileID) {
+            Versioned::set_stage($this->getVersionedStage());
+            return File::get()->filter(['FileFilename' => $parsedFileID->getFilename()])->first();
+        });
 
         // Could not find a valid file, let's bail.
         if (!$file) {
@@ -114,11 +115,13 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
     {
         $filename = $parsedFileID->getFilename();
         $variant = $parsedFileID->getVariant();
+
         // Let's try to match the plain file name
-        $stage = Versioned::get_stage();
-        Versioned::set_stage($this->getVersionedStage());
-        $file = File::get()->filter(['FileFilename' => $filename])->first();
-        Versioned::set_stage($stage);
+        /** @var File $file */
+        $file = Versioned::withVersionedMode(function () use ($filename) {
+            Versioned::set_stage($this->getVersionedStage());
+            return File::get()->filter(['FileFilename' => $filename])->first();
+        });
 
         if ($file) {
             return [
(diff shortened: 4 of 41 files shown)


src/FilenameParsing/FileResolutionStrategy.php
```diff
@@ -0,0 +1,61 @@
...
+    public function softResolveFileID($fileID, Filesystem $filesystem);
+
+    /**
+     * Try to find a file ID for an existing file the provided file tuple.
+     * @param array|ParsedFileID $tuple
+     * @param Filesystem $filesystem
+     * @param boolean $strict Whatever we should enforce a hash check on the file we find
+     * @return ParsedFileID|null FileID
+     */
+    public function searchForTuple($tuple, Filesystem $filesystem, $strict = true);
```
chillu: So this accepts a `ParsedFileID` of a file that might not exist in this path, and then returns a matching one that exists in a different path?

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -0,0 +1,334 @@
...
+
+    public function resolveFileID($fileID, Filesystem $filesystem)
+    {
+        foreach ($this->resolutionFileIDHelpers as $fileIDHelper) {
+            $parsedFileID = $fileIDHelper->parseFileID($fileID);
+            if (!$parsedFileID) {
+                continue;
+            }
+
+            if ($redirect = $this->searchForTuple($parsedFileID, $filesystem, false)) {
```
chillu: That naming is a bit confusing - it's not actually a redirect in the HTTP sense, but rather a different ParsedFileID, right?

src/FilenameParsing/FileResolutionStrategy.php
chillu: Can you define how a "file tuple" is different from a "file id"? Wondering if this method is named a bit too confusing

maxime-rainville:
> The introduction of ParsedFileID makes the higher level APIs a bit inconsistent now, e.g. DBFile with public function setFromLocalFile($path, $filename = null, $hash = null, $variant = null, $config = array()).

Yeah, I've been resisting the urge to update some existing method to use ParsedFileID. My thinking right now is to treat `FileIDHelper` and `FileResolutionStrategy` as internal constructs that the average developer doesn't need to worry about.

> We're doing more file_exists() checks now (more strategies), which isn't too impactful on a local filesystem. But we also don't want to make it harder to support object storage like S3, where it's a more expensive HTTP request. Could we configure new projects not to use the "legacy" helpers via YAML config, and recommend that in the upgrading guide?

Yeah, it might be worthwhile seeing if we can add some sort of caching layer for file checks. There's also a bunch of places where I need to read the content of the file to figure out what its hash is because I can't know it from reading its path.

I was thinking at updating the installer to have something like this:

```yml
  # Define public resolution strategy
  SilverStripe\Assets\FilenameParsing\FileResolutionStrategy.public:
    class: SilverStripe\Assets\FilenameParsing\FileIDHelperResolutionStrategy
    properties:
      ResolutionFileIDHelpers:
        - '%$SilverStripe\Assets\FilenameParsing\NaturalFileIDHelper'
      DefaultFileIDHelper: '%$SilverStripe\Assets\FilenameParsing\NaturalFileIDHelper'
      VersionedStage: Live
  # Define protected resolution strategy
  SilverStripe\Assets\FilenameParsing\FileResolutionStrategy.protected:
    class: SilverStripe\Assets\FilenameParsing\FileIDHelperResolutionStrategy
    properties:
      DefaultFileIDHelper: '%$SilverStripe\Assets\FilenameParsing\HashFileIDHelper'
      ResolutionFileIDHelpers:
        - '%$SilverStripe\Assets\FilenameParsing\HashFileIDHelper'
      VersionedStage: Stage
```

Redirect won't be needed for a brand new project, because the files are all goign to be at the right place from the start.

> Also, should we turn around the order of ResolutionFileIDHelpers for the public strategy? We should assume that most users will opt-in to running those migration scripts, and check for NaturalFileIDHelper first, and only use HashFileIDHelper and LegacyFileIDHelper as a fallback?

Can't switch the order around. If you ask the NaturalFileIDHelper to resolve an hash path, it's going to think the hash folder is part of the filename.

The performance hit on that shouldn't be too bad because, the helpers only do regex comparisons. They don't actually read the file in any way.

> We should clarify the legacy_filenames setting: It's now the default mode, at least for public files. That advice should span back to the 4.x upgrade docs and the existing file migration docs. So the remaining reason for legacy_files=true is that you don't want to run any migration scripts, and don't care about consistent file archiving?

If you leave `legacy_filenames` disabled, your public resolution strategy will default to natural paths and your protected resolution strategy defaults to hash paths. This allows you to keep an archive of protected files.

If you enable `legacy_filenames`, your protected resolution strategy need to defaults to natural paths as well. I haven't done this, but I'll have to add some hack somewhere, so that if you enable `legacy_filenames`, the protected resolution strategy defaults to using `NaturalFileIDHelper`.

chillu:
> I was thinking at updating the installer to have something like this:

Yep, that what I was thinking.

> Can't switch the order around. If you ask the NaturalFileIDHelper to resolve an hash path, it's going to think the hash folder is part of the filename.

Derp derp, of course.

src/FilenameParsing/FileResolutionStrategy.php
maxime-rainville: Many methods expect either a file ID string, a ParsedFileID or file tuple array . File tuple arrays can easily be converted to ParsedFileID because they contain the same info.

My original idea was that logic that was using file tuple arrays could easily call a FileResolutionStrategy without needing to convert the data to ParsedFileID. In hindsight, that wasn't nearly as helpful as I thought it would be.

maxime-rainville:
@silverstripe/core-team This is a pretty big change in the way we handle assets. It probably would be helpful to have a wide array of people pitch in on the review.

maxime-rainville:
Just pushed a bugfix for PHP 5.6

The travis build failure is due to the file migration task which I haven't converted yet.

maxime-rainville:
I ran the unit tests locally for `asset-admin` with this PR. There were two minor failure that were cause by files being publish under a different URLs. Got a related PR at https://github.com/silverstripe/silverstripe-asset-admin/pull/932

I also ran the behat test for asset-admin, they were only two failures which appear to be pre-existing. https://travis-ci.org/silverstripe/silverstripe-asset-admin/jobs/515980041

I was planning on throwing the CWP kitchen sink at it, but it looks like that got some pre-existing failure right now with LDAP that crashes every single test.

tractorcow:
> When testing this locally, you probably want to remove this line from the .htaccess file in your assets folder: RewriteCond %{REQUEST_FILENAME} !-f. Otherwise a big chunk of your request will never it PHP.

Does this mean public assets aren't going through apache, and that they need to go through PHP to be served? Why do you need to remove this for testing locally only?

src/FilenameParsing/ParsedFileID.php
```diff
@@ -0,0 +1,122 @@
...
+
+    /** @var string */
+    private $hash;
+
+    /**
+     * ParsedFileID constructor.
+     * @param string $filename
+     * @param string $hash
+     * @param string $variant
+     * @param string $fileID Original FileID use to generate this ParsedFileID
```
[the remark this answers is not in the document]
tractorcow: Ah, no I came up with this. Internal to the flysystem backend, we combine name + hash + variant to come up with an internal combined string, that flysystem uses as a unique index within the store (i.e. base path + id = physical disk path).

tractorcow:
Another approach you could have taken to this is to completely deprecate $fileID entirely, and instead substitute by always passing in the entire tuple. That way you don't need to reverse engineer the tuple from the file ID, and you can conditionally ignore certain parts (e.g. hash, for public path) to generate the flysystem Id as necessary.

I would have just about created a new AssetStore backend, in order to avoid breaking the existing logic. You could make switching between the old and new stores as easy as a yml config. :P

This way works, of course, but it feels like a lot more code than probably necessary.

src/FilenameParsing/ParsedFileID.php
```diff
@@ -0,0 +1,122 @@
...
+
+    /** @var string */
+    private $variant;
+
+    /** @var string */
+    private $hash;
+
+    /**
+     * ParsedFileID constructor.
+     * @param string $filename
```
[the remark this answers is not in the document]
tractorcow: Yes :D

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -152,30 +179,177 @@ public function setProtectedFilesystem(Filesystem $filesystem)
...
+     * If the closure returns `false`, `applyToFileOnFilesystem` will carry on and try the follow up steps.
+     *
+     * Any other value the closure returns (including `null`) will be returned to the calling function.
+     *
+     * @param callable $closure Action to apply.
+     * @param string|array|ParsedFileID $fileID File identication. Can be a string, a file tuple or a ParsedFileID
+     * @param bool $strictHashCheck
+     * @return mixed
+     */
+    private function applyToFileOnFilesystem(callable $closure, $fileID, $strictHashCheck = true)
```
tractorcow: I feel like a lot of places we are using $fileID (and subsequently parsing these) we should really be using the tuple.

I'd not mind if this involves us deprecating protected methods that take $fileID as an argument (no public AssetStore methods take $fileID).

tractorcow: The more arguments you support, the behaviour of the code becomes more ambiguous and harder to test. Enforce the most normalised form for better stability. :)

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -152,30 +179,177 @@ public function setProtectedFilesystem(Filesystem $filesystem)
...
+        /** @var Filesystem $fs */
+        /** @var FileResolutionStrategy $strategy */
+        /** @var string $visibility */
+
+        // First we try to search for exact file id string match
+        foreach ([$publicSet, $protectedSet] as $set) {
+            list($fs, $strategy, $visibility) = $set;
+
+            // Get a FileID string based on the type of FileID
+            $fileIdStr = $fileID instanceof ParsedFileID ?
```
tractorcow: Argument normalisation should really be done as early as possible.

tractorcow: Instead of all of these argument type checking, either do argument normalisation early in the method, or (more preferably, as I earlier suggested) limit the allowed arguments.

src/FilenameParsing/ParsedFileID.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+/**
+ * Immutable representation of a parsed fileID broken down into its sub-components.
+ */
+class ParsedFileID
```
tractorcow: This class is really just another drop-in alternative to the tuple array right? Do we need both? Do we have a clear delineation for when one is used over another?

src/FilenameParsing/ParsedFileID.php
```diff
@@ -0,0 +1,122 @@
...
+    public function setFileID($fileID)
+    {
+        return new self($this->filename, $this->hash, $this->variant, $fileID);
+    }
+
+    /**
+     * @param string $filename
+     * @return self
+     */
+    public function setFilename($filename)
```
tractorcow: This method looks like a setter, but it isn't. Try to avoid deceptive naming.

maxime-rainville:
> > When testing this locally, you probably want to remove this line from the .htaccess file in your assets folder: RewriteCond %{REQUEST_FILENAME} !-f. Otherwise a big chunk of your request will never it PHP.
> 
> Does this mean public assets aren't going through apache, and that they need to go through PHP to be served? Why do you need to remove this for testing locally only?

Public assets should still be serve directly without going through PHP. However, if you do go through PHP to serve your public assets, you should get exactly the same results. I just wanted to highlight that to fully test the FlysystemAssetStore, you need to disable that HTACCESS rule.

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+use SilverStripe\Core\Injector\Injectable;
+
+/**
+ * Parsed Natural path URLs. Natural path are the same hashless path that appear in the CMS.
+ *
+ * Natural paths are used by the public adapter from SilverStripe 4.4 and on the protected adapter when
```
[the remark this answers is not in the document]
tractorcow: The difference between legacy_filenames, and this, is that protected files don't need to be moved during update. Even if the public files are the same, protected files will be "broken" until manually moved. That's what `legacy_filenames` is intended to do.

What you should do is change `legacy_filenames` so that it forces the same public resolution strategy, but for private files, and that will prevent regressions. :)

src/FilenameParsing/FileResolutionStrategy.php
tractorcow: I suggest we stop using `$fileID` as an argument, and just generate it from the tuple as needed.

We don't really need both tuple and ParsedFileID. I don't think the class adds much value over the array as a structure, but picking one is more important than picking the right one. ;P

chillu:
> Another thing to keep in mind is that his new approach does a lot more file existence check and file reads (to build hashes) than the previous ones. There's not really a way around it because of the nature of the change, but we do want to make sure we don't affect serving time to much.

If people run the recommended migration task to move published files into hash-less filesystem paths, the vast majority of asset requests should never hit PHP in the first place. Asset requests routed through PHP are then limited to protected/private files, as well as existing external links with hashes. And if those external links come from search engines, they'll update to the correct link after receiving a redirect on next reindex. So I think we're good there.

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -0,0 +1,334 @@
...
+                return $redirect;
+            }
+
+            // If our helper managed to parse the file id, but could not resolve to an actual physical file,
+            // there's nothing else we can do.
+            return null;
+        }
```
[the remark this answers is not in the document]
maxime-rainville: Fixed.

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -0,0 +1,334 @@
...
+                continue;
+            }
+
+            if ($redirect = $this->searchForTuple($parsedFileID, $filesystem, false)) {
+                return $redirect;
+            }
+
+            // If our helper managed to parse the file id, but could not resolve to an actual physical file,
+            // there's nothing else we can do.
+            return null;
```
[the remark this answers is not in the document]
maxime-rainville: The assumption is that all the method can return `null` when a file can not be found or that a file ID can not be parsed. It's not an especially uncommon or problematic outcome.

I think returning `null` is a better fit than throwing an exception.

src/FilenameParsing/ParsedFileID.php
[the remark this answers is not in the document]
maxime-rainville: I'll let Damian take the fall for this naming convention since he's volunteering to do so.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -162,6 +180,52 @@ public function getProtectedFilesystem()
...
+     * @return FileResolutionStrategy
+     */
+    public function getPublicResolutionStrategy()
+    {
+        if (!$this->publicResolutionStrategy) {
+            $this->publicResolutionStrategy = Injector::inst()->get(FileResolutionStrategy::class . '.public');
+        }
+
+        if (!$this->publicResolutionStrategy) {
+            throw new Exception("Filesystem misconfiguration error");
```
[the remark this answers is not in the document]
maxime-rainville: Switch it to a `LogicException`. Applied similar fixes to the `getPublicFilesystem()` and `getProtectedFilesystem()`.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -170,14 +234,96 @@ public function getProtectedFilesystem()
      */
     protected function getFilesystemFor($fileID)
     {
-        if ($this->getPublicFilesystem()->has($fileID)) {
+        $public = $this->getPublicFilesystem();
+        $protected = $this->getProtectedFilesystem();
+        $publicStrategy = $this->getPublicResolutionStrategy();
+        $protectedStrategy = $this->getPublicResolutionStrategy();
+
+        if ($public->has($fileID)) {
```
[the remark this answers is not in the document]
maxime-rainville: That's been rewritten.

src/FilenameParsing/NaturalFileIDHelper.php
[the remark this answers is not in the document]
maxime-rainville: We had a discussion about what to do with `legacy_filenames`. We decided to ignore `legacy_filenames` going forward. Everyone will use natural path for public and hash path for protected files.

Because both path type will be resolved on both store now, the files can stay where they are without issues.

If someone really wanted to keep natural path on the protected store, they could update the protected resolution strategy to default to NaturalFileIDHelper.

chillu:
We've introduced `FlysystemAssetStore.redirect_response_code` with a `302 Temporary Redirect` by default in 4.3 (via https://github.com/silverstripe/silverstripe-assets/commit/d76ce8535aba9cbf1c226963cb1003ef06aa24ac). We've discussed making that temporary because the redirections were going the other way around in 4.4 if you've opted into the migration.

4.3: 302 on `my-folder/file.jpg => my-folder/[hash]/file.jpg`
4.4 without migration: 301 on `my-folder/file.jpg => my-folder/[hash]/file.jpg`
4.4 with migration: 301 on `my-folder/[hash]/file.jpg => my-folder/file.jpg`

I think it's important to switch this to a `301 Permanent Redirect` in 4.4 in order to allow those redirects to be cached in intermediary HTTP layers (e.g. CDNs, see [CloudFlare docs](https://support.cloudflare.com/hc/en-us/articles/115003011091-3xx-Redirection) as an example). This avoids hitting PHP for assets which need a redirect. By default, no `Cache-Control` headers are added. Even if you've configured `.htaccess` or `nginx` to add cache-control to assets (based on mime-type or file extension), that won't apply to those redirects. So if you've got a lot of traffic, both the "4.4 without migration" and "4.4 with migration" decrease your CDN cache hit rates and might cause instability at the origin (SilverStripe servers). I'd prefer to just point that out to advanced users through the changelog (there's non-trivial ways to achieve this in `.htaccess`, rather than add extension points to `FlysystemAssetStore->createRedirectResponse()`.

chillu:
@tractorcow Talked a bit more with @maxime-rainville about the new APIs, and `$fileID` vs. `$tuple` vs. `ParsedFileID`. We think marking the new `FileResolutionStrategy` and `FileIDHelper` as `@internal` APIs for now will remove the need to get this perfect on the first shot (in 4.4.0). We're under the gun to get this into the release, and Max has spent more than a month on this already - so it's pretty late to do deeper refactoring without jeopardising the whole release timing. To SS devs, the main way you interact with assets is the `AssetStore` interface, and that remains unchanged (no addition of `ParsedFileID`). Apart from one additional method on `FlysystemAssetStore->swapPublish()`, which we only use in one place (`AssetControlExtension`). And that place has a `method_exists()` guard to avoid breaking existing third party `AssetStore` implementations.

As part of this, we'd make `FlysystemAssetStore->getPublicResolutionStrategy()` and `FlysystemAssetStore->getProtectedResolutionStrategy()` protected methods, and mark them as `@internal`. So the whole "resolution strategy" business becomes an implementation detail. One niggle: The Injector config for those is now added to the [recipe](https://github.com/silverstripe/recipe-core/pull/40), which we need for BC. If we change the approach completely in the future, we might need to leave some stub APIs, because we're forcing people to reference those APIs in their `mysite.yml` configs.

Does that make sense? Do you think it's a good middle ground to get this out of the door?

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * You may also provide a `VersionedStage` to only look at files that were published.
+ */
+class FileIDHelperResolutionStrategy implements FileResolutionStrategy
+{
+    /**
+     * The FileID helper that will be use to build FileID for this adapter.
+     * @var FileIDHelper
+     */
+    private $defaultFileIDHelper;
```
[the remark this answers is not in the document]
maxime-rainville: Strictly speaking  "file ID helper used for writing" is not 100% right. When new variants are written, they will get an FileID format matching the one of their main file.

Also, `searchForTuple` will always start by using the `defaultFileIDHelper` when trying to find a file. The idea being that it should be the the most common one and therefore the most likely to find a file and avoid unnecessary file existence check.


In this context `defaultFileIDHelper` sounds like an appropriate name.

src/FilenameParsing/FileIDHelper.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param string|ParsedFileID $filename Name of file or ParsedFileID object
+     * @param string $hash Hash of original file
+     * @param string $variant (if given)
+     * @return string Adapter specific identifier for this file/version
+     */
+    public function buildFileID($filename, $hash = null, $variant = null);
+
+
+    /**
+     * Performs filename cleanup before sending it back.
```
bergice: Maybe this should be more clear on what sort of cleanup is to be made to the filename?

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($extension) {
+            $fileID .= $extension;
+        }
+
+        return $fileID;
+    }
+
+    public function cleanFilename($filename)
+    {
+        // There's not really any relevant cleaning rule for legacy. It's not important any way because we won't
```
bergice:
```suggestion
        // There's not really any relevant cleaning rule for legacy. It's not important any way because we won't be
```

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -0,0 +1,102 @@
...
+        // generating legacy URLs, aside from maybe for testing.
+        return $filename;
+    }
+
+    /**
+     * @note LegacyFileIDHelper is meant to fail when parsing newer format fileIDs with a variant e.g.:
+     * `subfolder/abcdef7890/sam__resizeXYZ.jpg`. When parsing fileIDs without variant, it should return the same
```
bergice:
```suggestion
     * `subfolder/abcdef7890/sam__resizeXYZ.jpg`. When parsing fileIDs without a variant, it should return the same
```

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+use SilverStripe\Core\Injector\Injectable;
+
+/**
+ * Parsed Natural path URLs. Natural path are the same hashless path that appear in the CMS.
```
bergice:
```suggestion
 * Parsed Natural path URLs. Natural path is the same hashless path that appears in the CMS.
```

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+    public function cleanFilename($filename)
+    {
+        // Since we use double underscore to delimit variants, eradicate them from filename
+        return preg_replace('/_{2,}/', '_', $filename);
+    }
+
+    /** @todo Update unit to test _resampled variants file id */
```
bergice: Is this done yet?

src/FilenameParsing/FileResolutionStrategy.php
maxime-rainville: I've been prioritising using Filename/Hash combo over file ID. There's a few places like `getResponseFor` where we have to use the FileID because that's what we get from the outside world.

There's a few existing methods on the `AssetStore` interface that return file tuple array. We can't change the return format of those without breaking semver.

I do think ParsedFileID provides value beyond a file array tuple. There's a lot of places where I need to swap value in and out (e.g.: getting a variantless version a file). If I'm using arrays for that, I've got to copy the values to a different array and unset the variant part. There's also the mundane problems that if you keep writing array keys as string all over the place, you're bound to mess the spelling a few times and those mistake might not be immediately apparent.

Having an immutable class to track the file tuple data makes these problems considerably easier to manage.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: Yeah ... my thinking when I started writing this function was that I would just start by normalising $fileID to a ParsedID and that would be it. However, as I threw more problems at it, I realised I needed to do things differently based on whatever I started with a string or a ParsedFileID.

I'll split it into two different method.

src/FilenameParsing/ParsedFileID.php
maxime-rainville: In a perfect world, we would only use ParsedFileID. However, right now we'll consider it an `@internal` class. It's only being used when interacting with a FileResolutionStrategy at this time.

src/FilenameParsing/ParsedFileID.php
maxime-rainville: `ParsedFileID` is meant to be immutable. That's important because I don't want to update a value on a ParsedFileID that I got from a caller method that intends to use it again for a different purpose.

So yes, `setFilename` is not strictly speaking a proper setter, but it effectively serves that purpose in an immutable context.

tests/php/FilenameParsing/FileIDHelperResolutionStrategyTest.php
```diff
@@ -0,0 +1,625 @@
...
+        } else {
+            $this->assertNull($responsePfid);
+        }
+    }
+
+    public function listVariantParsedFiledID()
+    {
+        $pfid = new ParsedFileID('folder/file.txt', 'abcdef7890');
+        return [
+            [$pfid->setFileID('folder/file.txt')->setHash(''), 'folder/file.txt'],
```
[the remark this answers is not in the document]
maxime-rainville: Oh ... Didn't know that. 

`$mind->isBlown() === true`

src/Dev/TestAssetStore.php
```diff
@@ -187,6 +195,8 @@ public function getDefaultConflictResolution($variant)
     protected function isSeekableStream($stream)
     {
         if (isset(self::$seekable_override)) {
+            // Unset the override so we don't get stuck in an efficit loop
```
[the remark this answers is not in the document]
maxime-rainville:
```suggestion
            // Unset the override so we don't get stuck in an infinit loop
```

src/AssetControlExtension.php
```diff
@@ -130,21 +130,22 @@ protected function processManipulation(AssetManipulationList $manipulations)
...
-        $this->publishAll($manipulations->getPublicAssets());
-
-        // Protect assets
-        $this->protectAll($manipulations->getProtectedAssets());
-
         // Check deletion policy
         $deletedAssets = $manipulations->getDeletedAssets();
         if ($archive && $this->isVersioned()) {
+            // Publish assets
+            $this->swapAll($manipulations->getPublicAssets());
```
chillu: Is there a test for this? I can't see any reference of `keep_archived_assets` in tests which would toggle this flag

src/AssetControlExtension.php
```diff
@@ -268,6 +269,32 @@ protected function deleteAll($assets)
         }
     }
 
+    /**
+     * Move all assets in the list to the public store
+     *
+     * @param array $assets
+     */
+    protected function swapAll($assets)
```
chillu:
```suggestion
    protected function swapPublishAll($assets)
```

src/FilenameParsing/FileResolutionStrategy.php
```diff
@@ -0,0 +1,86 @@
+<?php
+namespace SilverStripe\Assets\FilenameParsing;
+
+use League\Flysystem\Filesystem;
+
+/**
+ * Represents a strategy for resolving files on a Flysystem Adapter.
+ */
+interface FileResolutionStrategy
```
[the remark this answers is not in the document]
maxime-rainville: I think `FileResolver` implies this is more deterministic than it actually is. You'll get different results based on what's on your file system. `Strategy` communicates the iterative nature of this process (aka throwing shit at the wall in a semi sensible order until something stick) more clearly.

But yes, the name of most of these classes are kind of a mouthful.

tests/php/FileTest.php
```diff
@@ -346,15 +348,24 @@ public function testGetURL()
     {
         /** @var File $rootfile */
         $rootfile = $this->objFromFixture(File::class, 'asdf');
-        $this->assertEquals('/assets/FileTest/55b443b601/FileTest.txt', $rootfile->getURL());
+        $this->assertEquals('/assets/55b443b601/FileTest.txt', $rootfile->getURL());
```
chillu: I don't understand why this is no longer in `assets/FileTest` as a protected file. The `TestAssetStore` should set the root folder to `assets/FileTest`, and this particular fixture doesn't have a parent folder set. The assertion below uses `assets/FileTest` again. Is this an inconsistency in the test or the implementation?

src/Dev/TestAssetStore.php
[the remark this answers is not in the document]
maxime-rainville:
```suggestion
            // Unset the override so we don't get stuck in an infinit loop
```

src/FilenameParsing/FileResolutionStrategy.php
chillu: > I do think ParsedFileID provides value beyond a file array tuple.

Yep, I agree - the data type benefits from a "schema" (via a PHP interface), and being able to make it immutable makes the code a lot cleaner (as opposed to arrays). But yeah, let's keep that discussion going on the main "thread" in this PR, resolving this convo and linking to it for reference from the main thread.

src/FilenameParsing/ParsedFileID.php
chillu: https://github.com/silverstripe/silverstripe-assets/pull/223#issuecomment-484260132

src/FilenameParsing/NaturalFileIDHelper.php
maxime-rainville: Fixed.

src/FilenameParsing/FileIDHelper.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+/**
+ * Helps build and parse FileIDs according to a predefined format.
```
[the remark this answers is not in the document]
maxime-rainville: Changed to:
> Helps build and parse Filename Identifiers (ake: FileIDs) according to a predefined format.

src/AssetControlExtension.php
```diff
@@ -268,6 +269,32 @@ protected function deleteAll($assets)
...
+    {
+        if (empty($assets)) {
+            return;
+        }
+
+        $store = $this->getAssetStore();
+
+        // The `swap` method was introduced in the 1.4 release. It wasn't added to the interface to avoid breaking
+        // custom implementation. If it's not available on our store, we fallback to a publish/protect
+        if (method_exists($store, 'swapPublish')) {
```
chillu:
```suggestion
        if (method_exists($store, 'swapPublishAll')) {
```

src/AssetControlExtension.php
```diff
@@ -268,6 +269,32 @@ protected function deleteAll($assets)
...
+            foreach ($assets as $asset) {
+                $store->swapPublish($asset['Filename'], $asset['Hash']);
```
chillu:
```suggestion
                $store->swapPublishAll($asset['Filename'], $asset['Hash']);
```

src/FilenameParsing/FileIDHelperResolutionStrategy.php
maxime-rainville: That was a left over from my SS4.3 redirect fix. Restructured things a bit so it's clearer.

tests/php/FileTest.php
maxime-rainville: The Protected File Adapter and Public File Adapter have slightly different logic for generating URLs.

When you're using the TestAssetStore, protected URLs won't have the extra folder and public URLs will. I spent an hour looking at it to see if there was an elegant way to fix it, but couldn't find one. Since this only affects unit test, I decided I had bigger fish to fry.

From memory the core problems is that we can't set a different ASSET_DIR constant for our tests.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: Done ... there's one method that accepts a `ParsedFiledID `and another that expects a File ID string.

I've removed the logic to handle a file file tuple array because, I'm never actually using it.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: Fix https://github.com/silverstripe/silverstripe-assets/pull/223#discussion_r276498373

tests/php/FileTest.php
chillu: I see. The problem is that `ProtectedAssetAdapter->getProtectedURL()` hardcodes the assets dir: https://github.com/silverstripe/silverstripe-assets/blob/1/src/Flysystem/ProtectedAssetAdapter.php#L58. The `AssetStore` which instanciates them should probably pass on its base path, rather than the adapter making guesses about `ASSETS_DIR` and `ASSETS_PATH`. But that's a pre-existing issue, and only affects tests. 

Further context from @maxime-rainville:

> The default behaviour of the various "setFileContent()" method use to be to create the file in the public store. That didn't make a lot of sense because most of the time that file would be attached to a draft entry, so the file would be move back to the protected store on save. It was also a bit problematic because natural paths don't work for versioning. So that would systematically overwrite the published file. To work around that I reverse the default. Now when in doubt we write to the protected store. A side effect of that is that some test that use to be run against the public store, are now being run against the protected one.

chillu: See https://github.com/open-sausages/silverstripe-assets/commit/632ddceab6068a98ba2d2a178a3546b18539fb27

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -0,0 +1,458 @@
...
+
+        return null;
+    }
+
+    public function stripVariant($fileID)
+    {
+        $hash = '';
+
+        // Normalise our parameters
+        if ($fileID instanceof ParsedFileID) {
```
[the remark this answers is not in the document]
maxime-rainville: I added a bit more detail. That's another of those method that received either `FileID`/`ParsedFileID`. We could split it up, but this one makes a little bit more sense, because it's easier to normalise a ParsedFileID to a FileID string.

tests/php/FilenameParsing/FileIDHelperResolutionStrategyTest.php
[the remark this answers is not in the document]
maxime-rainville: I've added it to this specific test, but not the other places where I use dataProvider.

src/Flysystem/FlysystemAssetStore.php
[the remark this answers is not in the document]
maxime-rainville: Fixed. I hadn't pondered the nuances between a `callable` and a `closure`.

src/FilenameParsing/FileIDHelper.php
maxime-rainville: I've added a somewhat more prescriptive description.

> Clean up filename to remove constructs that might clash with the underlying path format of this FileIDHelper.

src/AssetControlExtension.php
maxime-rainville: `RedirectKeepArchiveFileControllerTest` test this indirectly. It extends `RedirectFileControllerTest` and runs all the same tests but with `keep_archived_assets` enabled.

`keep_archived_assets` is perhaps a bit counter intuitive in that you can apply it to specific DataObjects without applying it to others. `AssetControlExtensionTest` tests this with `ArchivedObject`.

But to be honest, I worked off the assumption this was already properly tested before I began and didn't look to closely into the existing tests beyond confirming they weren't broken.

Still the `RedirectKeepArchiveFileControllerTest/RedirectFileControllerTest` combo is pretty trough and goes through various scenarios of publishing, unpublishing, archiving, creating a draft, etc.

src/FilenameParsing/ParsedFileID.php
tractorcow: Happy with the `@internal`

tractorcow:
@chillu @maxime-rainville given that the new class is internal, and I don't have major issues with the implementation (and really just the style) I'm happy to give this my approval. :)  However I'm not reviewing this at a deep level, so I'll leave the two of you to finalise getting this ready for merge.

Good job both of you. Glad to see you excelling in owning this space @maxime-rainville :)

src/AssetControlExtension.php
maxime-rainville: `swapPublish` on `AssetStore` only publishes one file at the time. So it doesn't make sense to call it `swapPublishAll`.

The swap publish method on AssetControllerExtension operates on a list of filename/hash combination.

maxime-rainville:
> I would have just about created a new AssetStore backend, in order to avoid breaking the existing logic. You could make switching between the old and new stores as easy as a yml config. :P

That actually makes a lot of sense. Didn't quite appreciate how intricate this change would be until I was half way through it.

chillu:
@maxime-rainville and myself discussed the need for permanent redirects, and came up with the following:

If `file.pdf` on filesystem: `GET xxxx/file.pdf => 301 file.pdf` and `GET file.pdf => 200`
If `xxxx/file.pdf` on filesystem: `GET xxxx/file.pdf => 200` and `GET file.pdf => 302 xxxx/file.pdf`

So this would avoid a situation where you don't opt-in to the 4.4 file migration task, cache redirects, then publish the file again which moves it (`xxxx/file.pdf => file.pdf`), and *then* the cached redirect breaks your requests.

maxime-rainville:
Permanent redirect fix is in.

Issue #219, opened 2019-03-12
Regorganise file migration task


maxime-rainville:
We have been doing lots of file management work lately. This will potentially require projects upgrading from a minor release to 4.4 to take extra step to bring their assets back to the recommended structure.

## Acceptance criteria

* Developers can update to the latest asset structure by running a single unified task, whether they are upgrading from 3.x or a pre-4.4 minor release.
* Unified task performs the following sub tasks:
  * Normalise DB File entries
  * Move files to the appropriate location
  * Find unknown files and adds matching DB entries
  * Generate thumbnails
  * Updates HTMLText references to files so they use short codes
* The unified task works with `legacy_filenames` or `keep_archive_assets` enabled.
* Developers can understand what the unified task does by reading the [File Migration doc](https://docs.silverstripe.org/en/4/developer_guides/files/file_migration/)
* Migration task works with 1000 files and images

## Related issues

* https://github.com/silverstripe/silverstripe-versioned/issues/177
* https://github.com/silverstripe/silverstripe-assets/issues/175
* https://github.com/silverstripe/silverstripe-assets/issues/188
* https://github.com/silverstripeltd/open-sourcerers/issues/45
* https://github.com/silverstripe/silverstripe-framework/issues/8664

## Notes

- Refactor `\SilverStripe\Assets\Flysystem\FlysystemAssetStore::parseFileID` to use file utility class.
- Consider if integration tests are appropriate given the complexity of combining the tasks

chillu:
Relies on https://app.zenhub.com/workspace/o/silverstripe/silverstripe-framework/issues/8924

chillu:
Has been done by implementing the referenced storycards in here.

> Normalise DB File entries
> Move files to the appropriate location

Those two aren't split out into individual tasks, but that's fine - they should always be run together

> Find unknown files and adds matching DB entries

We haven't completed that card, but it's not blocking closing this card here.

Issue #175, opened 2018-11-01
Inability to sync unknown files makes life difficult for devops


maxime-rainville:
SilverStripe 3 had a sync task that allowed SilverStripe to look through the assets folder and discover unknown files. Files in SilverStripe 4 no longer neatly maps to the assets folder in the file system. While this gives us more flexibility in how we serve the files in the browser, it also cause big headaches for devops.

# Example of problematic situations
* Let's say I restore a file-only snapshot. Some files won't be available because the database doesn't know about them.
* Let's say I restore a DB-only snapshot from a different environment. Some files won't be available anymore because that other environment didn't know about them.
* Let's say I have a huge number of files that I want to import into my CMS hosted on a remote server. 
  * I either have to build a custom script to do it for me,
  * or I have to do it locally through the UI and export a full snapshot to my remote server.

chillu:
Hmm, yeah that's why asset and database snapshots should be restored together - they're a connected point in time. But I get why that's not always feasible. The last case you mention (migration) is probably the biggest concern for me: We want to make migration into our systems as easy as possible. Marking as `impact/high`

brettt89:
This has caused issues with a migration of ours where the SS 3.x website didn't have database records for all assets that they had. So when they upgraded to SS4 and ran the "FileMigrationTask" it only knew about the ones in the database and didn't migrate most of the data.

Perhaps it would be easier to have the PHP script actually crawl the `assets/` directory (or defined directory) to get a list of files that need to migrate (and create DB records for them if they don't exist) rather than using the Database as the source of truth?

brettt89:
Otherwise perhaps we should be clearly saying to people upgrading from SS3 to SS4 to "Ensure you run the FilesystemSyncTask on your SilverStripe 3.x installation before upgrading to SS4."

As we have had to re-deploy the SS 3.x version of the website just to run this task so we can re-deploy the SS4 version.

maxime-rainville:
THis is kind of related to https://github.com/silverstripeltd/open-sourcerers/issues/45

kinglozzer:
I’ve also found myself needing this when client needed a ~70mb PDF uploaded to the asset area. I gave up and temporarily raised the limits in php.ini, I didn’t fancy having to download and then re-upload asset snapshots just for one file!

maxime-rainville:
Here are some random ideas in some random order:
[...]

PR #225, opened 2019-03-18
Add file sync task

bergice:
This task is intended to fix up the filesystem/db in version 4.

It will remove invalid db entries and create db entries for assets that are currently missing db entries.

### DIFF
--- a/src/Dev/Tasks/FilesystemSyncTask.php
+++ b/src/Dev/Tasks/FilesystemSyncTask.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace SilverStripe\Dev\Tasks;
+
+use SilverStripe\Assets\FilesystemSyncTaskHelper;
+use SilverStripe\Dev\BuildTask;
+use SilverStripe\AssetAdmin\Helper\ImageThumbnailHelper;
+
+/**
+ * This task will synchronise the filesystem with the database.
+ *
+ * The execution works like this:
+ *
+ * - if the skip_validate_database argument is not set (default):
+ *  iterate files in db and check they have valid files in the filesystem - if not, remove the db entries
+ *
+ * - files in legacy folders create db entries - the file is then deleted
+ *
+ * - if the delete_broken argument is not set (default):
+ *  files in hashed folders without db entries have db entries created
+ * - if the delete_broken argument is set:
+ *  files in hashed folders without db entries are deleted
+ *  db entry files without assets are deleted
+ *
+ * - delete folders that are now empty
+ *
+ * - generate thumbnails
+ */
+class FilesystemSyncTask extends BuildTask
+{
+    private static $segment = 'FilesystemSyncTask';
+
+    protected $title = 'Filesystem Sync Task';
+
+    protected $description = '
+        Syncs files in the assets folder with the database.
+        Files in the assets folder that do not exist in the database will be added.
+        This is useful if you restore restore a file-only snapshot or want to sync a lot of files.
+		Parameters:
+		- path: Sets path to sync. Defaults to the root assets path.
+		- draft: Files will not be published if set.
+		- delete_broken: Hashed files without a matching database entry will be removed
+		  and database file records will be removed if their asset is missing.
+    ';
+
+    public function run($request)
+    {
+        $results = FilesystemSyncTaskHelper::singleton()->run($request->getVars());
+
+        // Generate Thumbnails
+        ImageThumbnailHelper::singleton()->run();
+    }
+}
--- a/src/File.php
+++ b/src/File.php
@@ -296,9 +296,9 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     }
 
     /**
-     * Find a File object by the given filename.
+     * Find a File object by the given name.
      *
-     * @param string $filename Filename to search for, including any custom parent directories.
+     * @param string $filename Name to search for, including any custom parent directories.
      * @return File
      */
     public static function find($filename)
--- a/src/FilesystemSyncTaskHelper.php
+++ b/src/FilesystemSyncTaskHelper.php
@@ -1,0 +1,349 @@
+<?php
+
+namespace SilverStripe\Assets;
+
+use SilverStripe\Assets\Flysystem\FlysystemAssetStore;
+use SilverStripe\Assets\Flysystem\PublicAssetAdapter;
+use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Core\Environment;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\Tasks\FilesystemSyncTask;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\ORM\ValidationException;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injectable;
+
+/**
+ * Helper class that syncs the filesystem assets folder with the database
+ * @see FilesystemSyncTask
+ */
+class FilesystemSyncTaskHelper
+{
+    use Injectable;
+    use Configurable;
+
+    /**
+     * @var array allowed extensions, only files with these extensions will be migrated
+     */
+    private $allowedExtensions;
+
+    /**
+     * @var bool whether files should be published or not
+     */
+    private $publish;
+
+    /**
+     * @var int[] results data
+     */
+    private $results;
+
+    /**
+     * @var array recursive array of directories to remove.
+     * Directories are only removed if empty once syncing is complete.
+     */
+    private $dirsToRemove;
+
+    /**
+     * @var string root path the sync task is ran on
+     */
+    private $path;
+
+    /**
+     * @var bool whether to remove files without a db entry and vice versa
+     */
+    private $delete_broken;
+
+    /**
+     * @var FlysystemAssetStore
+     */
+    private $flysystemAssetStore;
+
+    /**
+     * @var PublicAssetAdapter
+     */
+    private $publicAssetAdapter;
+
+    /**
+     * @param array $options
+     * @return array An array containing results
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    public function run(array $options)
+    {
+        $this->prepare($options);
+
+        // Check for db files with missing filesystem files and remove those db entries
+        $this->deleteBroken();
+
+        // Sync Files
+        $this->syncFiles();
+
+        // Remove Directories
+        $this->removeDirectories();
+
+        return $this->results;
+    }
+
+    /**
+     * Prepare for execution
+     * @param array $options
+     * @throws ValidationException
+     */
+    private function prepare(array $options)
+    {
+        // Set max time and memory limit
+        Environment::increaseTimeLimitTo();
+        Environment::increaseMemoryLimitTo();
+
+        // Prepare for execution
+        $this->allowedExtensions = File::getAllowedExtensions();
+        $this->path = isset($options['path']) ? $options['path'] : '';
+        $draft = isset($options['draft']) ? $options['draft'] : false;
+        $this->publish = class_exists(Versioned::class) && !$draft;
+        $this->delete_broken = isset($options['delete_broken']) ? $options['delete_broken'] : null;
+        $this->dirsToRemove = [];
+        $this->results = [
+            'filesSyncedWithDb' => 0,
+            'dirsSyncedWithDb' => 0,
+            'filesRemovedFromDb' => 0,
+            'filesRemovedFromFilesystem' => 0,
+            'filesSkippedFromFilesystem' => 0
+        ];
+        $assetStore =  Injector::inst()->get(AssetStore::class);
+        if (!$assetStore instanceof FlysystemAssetStore) {
+            throw new ValidationException('FlysystemAssetStore is unavailable!');
+        }
+        /** @var FlysystemAssetStore $flysystemAssetStore */
+        $this->flysystemAssetStore = $assetStore;
+        /** @var PublicAssetAdapter $adapter */
+        $this->publicAssetAdapter = $this->flysystemAssetStore->getPublicFilesystem()->getAdapter();
+    }
+
+    private function deleteBroken()
+    {
+        if ($this->delete_broken) {
+            $this->deleteFilesWithMissingAssets();
+        }
+
+        DB::alteration_message(
+            $this->results['filesRemovedFromDb'] . ' files removed from the database (because they are missing files on the filesystem)',
+            "info"
+        );
+    }
+
+    /**
+     * @throws ValidationException
+     */
+    private function syncFiles()
+    {
+        foreach ($this->publicAssetAdapter->getFileGenerator($this->path, function ($file) {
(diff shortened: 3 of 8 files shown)


src/File.php
```diff
@@ -318,30 +318,6 @@ public static function find($filename)
         return $item;
     }
 
-    /**
-     * Find a File object by the given filename.
-     *
-     * @param string $filename Filename to search for, including any custom parent directories.
-     * @return File
-     */
-    public static function findByFilename($filename)
```
ScopeyNZ: Removing this would be a breaking (major) change. Did you mean to deprecate?

src/FilesystemSyncTaskHelper.php
```diff
@@ -56,20 +57,51 @@ class FilesystemSyncTaskHelper
...
+
+        // Remove Directories
+        $this->removeDirectories();
+
+        return $this->results;
+    }
+
+    /**
+     * Prepare for execution
+     * @param $options
```
ScopeyNZ: Missing type hint. If it's an array you can put the typehint in the function arguments too.

src/FilesystemSyncTaskHelper.php
```diff
@@ -56,20 +57,51 @@ class FilesystemSyncTaskHelper
...
+     */
+    private function prepare($options)
     {
         // Set max time and memory limit
         Environment::increaseTimeLimitTo();
         Environment::increaseMemoryLimitTo();
 
         // Prepare for execution
         $this->allowedExtensions = File::getAllowedExtensions();
-        $this->path = isset($options['path']) ? $options['path'] : ASSETS_PATH;
```
ScopeyNZ: Could this break someone's setup (who's not running on a managed platform with Flysystem)? Maybe we can't make this change in a minor version...

src/FilesystemSyncTaskHelper.php
```diff
@@ -238,24 +295,26 @@ private function canSyncFileToDatabase(SplFileInfo $splFileInfo)
      * @param SplFileInfo $splFileInfo
      * @throws \SilverStripe\ORM\ValidationException
      */
-    private function syncFileToDatabase(SplFileInfo $splFileInfo)
+    private function syncFileToDatabase(array $splFileInfo)
```
ScopeyNZ: Can I recommend that you change the variable name? `$splFileInfo` was already a pretty bad name and now it means even less considering the type change.

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
 
         return $result;
     }
+
+    /**
+     * @param $directory
```
ScopeyNZ: Type hint

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
...
+     * @return \Generator Returns a generator for recursively iterating through a directory.
+     * This is intended to be less memory intensive than `listContents`
+     * as it yields per directory instead of returning the whole directory tree as one array.
+     * @see listContents
+     */
+    public function getFileGenerator($directory, $filter = null)
+    {
+        $files = $this->listContents($directory, false);
+        foreach ($files as $file) {
+            if ($filter && !$filter($file)) {
```
ScopeyNZ: `is_callable($filter)` is safer here.

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
...
+
+            yield $file;
+            if ($file['type'] == 'dir') {
+                yield from $this->getFileGenerator($file['path']);
+            }
+        }
+    }
+
+    /**
+     * @param $relativePath
```
ScopeyNZ: Type hint. Should this be `public` API?

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -996,4 +997,30 @@ protected function createErrorResponse($code)
...
+
+
+    /**
+     * Find a File object by the given filename.
+     *
+     * @param string $filename Filename to search for, including any custom parent directories.
+     * Should include hashed folder the file is located inside.
+     * @return File
+     */
+    public static function getFileByFilename($filename)
```
ScopeyNZ: Perhaps in the function that was removed above you can just replace all the code with a call to this function (and deprecate it)?

tests/php/FilesystemSyncTaskTest.php
```diff
@@ -71,26 +71,15 @@ public function tearDown()
...
-        // Prior to sync, check that each file has empty Filename / Hash properties
-        /** @var File $file */
-        foreach (File::get()->exclude('ClassName', Folder::class) as $file) {
-            $filename = $file->generateFilename();
-            $this->assertNotEmpty($filename, "File {$file->Name} has a filename");
-            $this->assertEmpty($file->File->getFilename(), "File {$file->Name} has no DBFile filename");
-            $this->assertEmpty($file->File->getHash(), "File {$file->Name} has no hash");
-            $this->assertFalse($file->exists(), "File with name {$file->Name} does not yet exist");
-            $this->assertFalse($file->isPublished(), "File is not published yet");
-        }
```
ScopeyNZ: Is this no longer relevant?

src/File.php
bergice: This is me just refactoring my own stuff from the previous commit.

src/FilesystemSyncTaskHelper.php
bergice: This task requires Flysystem. There is an exception further down that gets thrown if it can't find it.

src/FilesystemSyncTaskHelper.php
bergice: Haha, fair enough.

src/File.php
ScopeyNZ: Ah yep - sorry.

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+
+        return false;
+    }
+
+    /**
+     * Whether the provided file should be synced with the database or not.
+     * @param array $fileTuple
+     * @return bool
+     */
+    private function canSyncFileToDatabase(array $fileTuple)
```
maxime-rainville: Big chunks of this method might need to be rewritten, once I'm done with my permalink jobs.

I'll leave some comments on thing, but maybe don't action them right away.

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+                return true;
+            }
+
+            $file = File::find($fileTuple['path']);
+            if ($file) {
+                $this->results['filesSkippedFromFilesystem']++;
+                return false;
+            } elseif ($this->delete_broken) {
+                // file exists in filesystem but not DB and is in a hash folder
+                $contents = file_get_contents($realPath);
```
maxime-rainville: Any specific reason why we're accessing the file directly here, rather than using flysystem?

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+            if (!in_array($extension, $this->allowedExtensions)) {
+                return false;
+            }
+
+            // check if parent is a hash
+            $path = dirname($fileTuple['path']);
+            $exploded = explode(DIRECTORY_SEPARATOR, $path);
+            $parent = array_pop($exploded);
+            $table = DataObject::singleton(File::class)->baseTable();
+            $parentExists = DB::prepared_query("SELECT COUNT(*) FROM \"$table\" WHERE \"FileHash\" LIKE ?", ["$parent%"])->value();
```
maxime-rainville: Same as before, you can use the ORM for this.

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+    private function syncFileToDatabase(array $fileTuple)
+    {
+        $realPath = $this->publicAssetAdapter->relativeToRealPath($fileTuple['path']);
+        $relativePath = substr($realPath, strlen($this->publicAssetAdapter->getPathPrefix()));
+        $relativeParentPath = dirname($relativePath);
+
+        // Create the file
+        if ($fileTuple['type'] == 'dir') {
+            /** @var Folder $file */
+            $file = Injector::inst()->create(Folder::class);
```
maxime-rainville:
```suggestion
            $file = Folder::create();
```

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+            $file->setFilename(basename($fileTuple['path']));
+        } else {
+            $fileClass = File::get_class_for_file_extension(pathinfo($fileTuple['path'], PATHINFO_EXTENSION));
+            /** @var File $file */
+            $file = Injector::inst()->create($fileClass);
```
maxime-rainville: `Folder` is also a subclass of `File`.

I would have initialised my `$file` variable before going into the `if` statement.

```php
/** @var Folder $file */
$file = null;
```

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+        $parentFile = File::find($relativeParentPath);
+        if ($parentFile) {
+            $file->ParentID = $parentFile->ID;
+        }
+
+        $file->write();
+
+        // Publish the file
+        if ($this->publish) {
+            $file->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
```
maxime-rainville: I think that will only publish the DB record, not the file.

src/FilesystemSyncTaskHelper.php
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        // Remove the original file
+        if ($fileTuple['type'] == 'dir') {
+            $this->results['dirsSyncedWithDb']++;
+        } else {
+            unlink($realPath);
```
maxime-rainville: Why not FlySystem?

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
...
+    {
+        $files = $this->listContents($directory, false);
+        foreach ($files as $file) {
+            if (is_callable($filter) && !$filter($file)) {
+                continue;
+            }
+
+            yield $file;
+            if ($file['type'] == 'dir') {
+                yield from $this->getFileGenerator($file['path']);
```
maxime-rainville: Didn't know about `yield from`. Looks pretty cool.

[But it doesn't work in PHP 5.6](https://www.php.net/manual/en/language.generators.syntax.php#control-structures.yield.from).

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
 
         return $result;
     }
+
```
maxime-rainville: Maybe define those in an interface methods and fall back to the non-generator method when not implemented.

src/Flysystem/AssetAdapter.php
```diff
@@ -195,4 +195,36 @@ protected function renderTemplate($template)
...
+    /**
+     * @param string $directory
+     * @param callable $filter
+     * @return \Generator Returns a generator for recursively iterating through a directory.
```
maxime-rainville: Maybe define the return statement as `Generator|array[]`. That way the `foreach` statement will have an idea of what it's getting back.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -771,7 +772,7 @@ protected function cleanFilename($filename)
      * @param string $fileID
      * @return array
      */
-    protected function parseFileID($fileID)
+    public function parseFileID($fileID)
```
maxime-rainville: We'll have a better way of doing this when the permalink job is merge.

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: We'll have a better way of doing this when the permalink job is merged.

src/Dev/Tasks/FilesystemSyncTask.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * - generate thumbnails
+ */
+class FilesystemSyncTask extends BuildTask
+{
+    private static $segment = 'FilesystemSyncTask';
+
+    protected $title = 'Filesystem Sync Task';
+
+    protected $description = '
```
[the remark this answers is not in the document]
bergice: This is just so it's more readable for us.

src/Flysystem/AssetAdapter.php
bergice: What do you mean?

src/Flysystem/FlysystemAssetStore.php
bergice: Okay, leaving a note in https://github.com/silverstripe/silverstripe-assets/issues/219

bergice:
> Just did a deeper code review this time. There's still a lot of native file system code used, maybe confirm that there's no easy equivalent to flysystem for those.
> 
> The unit test looks very light. It's not testing with `legacy_flenames` and `keep_archive_assets` flags. I'll have a go at running it for real this afternoon.

I've added a couple of more tests for this now.

Looks like the namespaces for permalinks needs to be fixed up though:

```
There was 1 error:

1) SilverStripe\Assets\Tests\FilesystemSyncTaskTest::testSyncKeepArchivedAssets
SilverStripe\Core\Injector\InjectorNotFoundException: ReflectionException: Class SilverStripe\Assets\FilenameParsing\FileResolutionStrategy.public does not exist in /var/www/mysite/www/vendor/silverstripe/framework/src/Core/Injector/InjectionCreator.php:17
```

sminnee:
I'm somewhat concerned about reintroducing this feature. We dropped it in the SS4 filesystem implementation because it was difficult to make robust and so caused follow on issues. 

 * If you're looking to update the database to re-list files that were previously uploaded to SilverStripe but the database records have been deleted, you risk data corruption. One example of why you might want to do this is if you're looking to restore filesystem assets but not the database. The source of corruption is that the IDs of the File records will be different, and so any other data objects that reference those files is going to be wrong. It would be more honest to say "if you want to restore assets you probably have to restore a database too".

 * If you're looking to update the database to include files that were added to the assets/ folder out-of-band, I guess you're implementing a bulk upload feature. If that's the goal I question whether this is the best way of achieving this, especially for users who don't have FTP access to their servers.

So I'd ask the question: what problem are we trying to solve by re-adding this feature, and is this the best solution to the problem?

sminnee:
If the goal is to bulk-add new assets to a locally-running project, something like this might be safer, simpler, and more useful:

```
vendor/bin/import-assets ~/Downloads/stuff-from-the-customer ./Uploads/StuffFromTheCustomer
```

Which would add the files to the filesystem underneath a  root folder ./Uploads/StuffFromTheCustomer

chillu:
Commented on the card: https://github.com/silverstripe/silverstripe-assets/issues/175#issuecomment-496746479

emteknetnz:

maxime-rainville:
I think we decided to drop this approach in the end. I don't think there's anything that can be salvage from this PR.

PR #227, opened 2019-03-29
Add image tag to shortcode task

bergice:
Finds old `img`/`a` tags referencing assets and rewrites them to use shortcodes

Depends on https://github.com/silverstripe/silverstripe-framework/pull/8892

# Extra tests to add
* [x] Test that live content gets updated as well
* [x] Test that other dataobjects
* [x] Test DB fields not called content
* [x] Test an HTMLVarchar
* [x] Test scenarios where files have been migrated
* [x] Test scenarios where files are protected


src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+
+        return $propertyValue;
+    }
+
+    /**
+     * Extracts an array of attributes from a tag.
+     * @param string $tag
+     * @param string $tagType
+     * @param $id
+     * @param $newSrc
```
maxime-rainville:
```suggestion
     * @param string $newSrc
```

`$id` is not on the method definition

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+            if ($height = static::getAttribute($tag, 'height')) {
+                $imgAttributes['height'] = $height;
+            }
+            if ($class = static::getAttribute($tag, 'class')) {
+                $imgAttributes['class'] = $class;
+            }
+            if ($alt = static::getAttribute($tag, 'alt')) {
+                $imgAttributes['alt'] = $alt;
+            }
+
```
maxime-rainville: Do we not care about other attributes that might be attached to our image tag?

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        return $attributes;
+    }
+
+    /**
+     * @param string $tagType
+     * @return string
+     */
+    private static function getShortCodeTypeFromTagType(string $tagType)
```
maxime-rainville:
```suggestion
    private static function getShortCodeTypeFromTagType($tagType)
```

maxime-rainville: Why does this method need to be private?

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+    {
+        return static::VALID_TAGS[$tagType];
+    }
+
+    private function getParsedFileIDFromSrc(string $src)
```
maxime-rainville:
```suggestion
    private function getParsedFileIDFromSrc($src)
```

Add PHP doc.

maxime-rainville: This method needs to be rethink a bit.

* What happens if the file hasn't been publish yet? You're not going to find it.
* You can probably get the expected file strategy directly from the FlysystemAssetStore
* URL always use forward slash.
* SoftResolve is meant to find files that use to exists, but have been archived. Is that really what you are going for here?

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+        $fileID = preg_replace($pattern, '', $src);
+
+        return $fileIDHelperResolutionStrategy->softResolveFileID($fileID, $filesystem);
+    }
+
+    /**
+     * @param string $tag
+     * @return string|null Returns the new tag or null if the tag does not need to be rewritten
+     */
+    private function getNewTag(string $tag)
```
maxime-rainville:
```suggestion
    private function getNewTag($tag)
```

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+    {
+        list('tagType' => $tagType, 'src' => $src) = $this->getTagTuple($tag);
```
maxime-rainville: Don't think this syntax works on PHP 5.6.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+
+        // Search for a File object containing this filename
+        $parsedFileID = $this->getParsedFileIDFromSrc($src);
+        /** @var File $file */
+        if ($parsedFileID && ($file = File::get()->filter('FileFilename:PartialMatch', $parsedFileID->getFilename())->first())) {
+            // Set the new src that points to the correct destination
+            $newSrc = DIRECTORY_SEPARATOR . ASSETS_DIR . DIRECTORY_SEPARATOR . $parsedFileID->getFileID();
```
maxime-rainville: URL forward slash are universal.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+     * @return string|null Returns the new tag or null if the tag does not need to be rewritten
+     */
+    private function getNewTag(string $tag)
+    {
+        list('tagType' => $tagType, 'src' => $src) = $this->getTagTuple($tag);
+
+        // Search for a File object containing this filename
+        $parsedFileID = $this->getParsedFileIDFromSrc($src);
+        /** @var File $file */
+        if ($parsedFileID && ($file = File::get()->filter('FileFilename:PartialMatch', $parsedFileID->getFilename())->first())) {
```
maxime-rainville: Why are you using a partial match here? Files are normally identified by their FileFilename and FileHash combination.

Also you may need to check for live files as well.

src/Dev/Tasks/TagsToShortcodeTask.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace SilverStripe\Assets\Dev\Tasks;
+
+use SilverStripe\Dev\BuildTask;
+
+/**
+ * Class TagsToShortcodeTask
+ * @package SilverStripe\Assets\Dev\Tasks
```
maxime-rainville: Update this please.

src/Dev/Tasks/TagsToShortcodeTask.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class TagsToShortcodeTask extends BuildTask
+{
+    private static $segment = 'TagsToShortcode';
+
+    protected $title = 'Rewrite tags to shortcodes';
+
+    protected $description = "
```
maxime-rainville: I think this only gets shown on the web interface, so all your content will be rendered on a single line.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+     * @throws \ReflectionException
+     */
+    public function run()
+    {
+        Environment::increaseTimeLimitTo();
+
+        $this->validTagsPattern = implode('|', array_keys(static::VALID_TAGS));
+        $this->validAttributesPattern = implode('|', static::VALID_ATTRIBUTES);
+
+        $classes = ClassInfo::getFieldMap($this->baseClass, $this->includeBaseClass, 'HTMLText');
```
maxime-rainville: There's an `HTMLVarchar` as well.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+            $updateSQL->execute();
+            DB::alteration_message('Updated page with ID ' . $row['ID'], 'changed');
+        }
+    }
+
+    /**
+     * @param string $content
+     * @return string
+     */
+    private function getNewContent(string $content)
```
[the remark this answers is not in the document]
maxime-rainville: I would make this method public, so arbitrary string of HTML can be converted. It will also make testing the class easier.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+
+        $this->validTagsPattern = implode('|', array_keys(static::VALID_TAGS));
+        $this->validAttributesPattern = implode('|', static::VALID_ATTRIBUTES);
+
+        $classes = ClassInfo::getFieldMap($this->baseClass, $this->includeBaseClass, 'HTMLText');
+        foreach ($classes as $class => $tables) {
+            foreach ($tables as $table => $fields) {
+                foreach ($fields as $field) {
+                    try {
+                        $records = DB::query("SELECT \"ID\", \"$field\" FROM \"{$table}\" WHERE \"$field\" RLIKE '<(".$this->validTagsPattern.")'");
```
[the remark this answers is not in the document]
bergice: How would you write this with `SQLSelect`?

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+                        $this->rewriteFieldForRecords($records, $table, $field);
+                    } catch (DatabaseException $exception) {
```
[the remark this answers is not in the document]
bergice: When running tests, a bunch of non-existing tables are included, this just ignores those.

src/Dev/Tasks/TagsToShortcodeHelper.php
bergice: How would you do that here using the ORM?

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+    /**
+     * Get all tags within some page content and return as array.
+     * @param $content
+     * @return array
+     */
+    private function getTagsInContent($content)
+    {
+        $resultTags = [];
+
+        preg_match_all('/<('.$this->validTagsPattern.').*?('.$this->validAttributesPattern.')\s*=.*?>/', $content, $matches, PREG_SET_ORDER);
```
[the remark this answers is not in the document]
bergice: I already considered this but opted for the regex route, mostly for performance reasons.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+        return $matches;
+    }
+
+    /**
+     * Gets the value of a given attribute from a tag.
+     * @param $tag
+     * @param $attribute
+     * @return null
+     */
+    private static function getAttribute($tag, $attribute)
```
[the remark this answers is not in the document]
bergice: Because they don't depend on local context.

src/Dev/Tasks/TagsToShortcodeHelper.php
bergice: It doesn't, it's just a habit I have :D

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,263 @@
...
+            foreach ($tables as $table => $fields) {
+                foreach ($fields as $field) {
+                    try {
+                        $sqlSelect = SQLSelect::create(
+                            ['ID', $field],
+                            $table
+                        );
+                        $whereAnys = [];
+                        foreach (array_keys(static::VALID_TAGS) as $tag) {
+                            $whereAnys[]= "\"$table\".\"Content\" LIKE '%<$tag%'";
```
bergice: This needs to work with square brackets too

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+    /**
+     * TagsToShortcodeHelper constructor.
+     * @param string $baseClass The base class that will be used to look up HTMLText fields
+     * @param bool $includeBaseClass Whether to include the base class' HTMLText fields or not
+     */
+    public function __construct($baseClass = null, $includeBaseClass = false)
+    {
+        $flysystemAssetStore = singleton(AssetStore::class);
+        if (!($flysystemAssetStore instanceof FlysystemAssetStore)) {
+            throw new InvalidArgumentException("FlysystemAssetStore missing");
```
maxime-rainville:
```suggestion
            throw new LogicException("TagsToShortcodeHelper requires your default asset store to extends FlysystemAssetStore.");
```

maxime-rainville: Juts making the message a little more clear for developers running the task.

.travis.yml
```diff
@@ -37,7 +37,7 @@ before_script:
 # Install composer dependencies
   - composer validate
   - composer install --prefer-dist
-  - composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev
+  - composer require --prefer-dist --no-update silverstripe/recipe-cms:4.4.x-dev
```
unclecheese: Do we really need the whole recipe here? Seems like it could just be `versioned` and `cms`, right? I'd rather not slow down the builds and keep the tests as isolated as possible.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+            $newContent = $this->getNewContent($content);
+            if ($content == $newContent) {
+                continue;
+            }
+
+            $updateSQL = SQLUpdate::create($updateTable)->addWhere(['"ID"' => $row['ID']]);
+            $updateSQL->addAssignments(["\"$field\"" => $newContent]);
+            $updateSQL->execute();
+            if ($this->logger) {
+                $this->logger->info("Updated page with ID {$row['ID']}");
```
chillu: Pages aren't the only things with HTMLText fields - this needs to mention the record type

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -0,0 +1,290 @@
...
+        if ($parsedFileID && $file) {
+            if ($tagType == 'img') {
+                $find = [
+                    '/(<|\[)img/',
+                    '/src\s*=\s*".*?"/',
+                    '/href\s*=\s*".*?"/',
+                    '/(>|\])/',
+                ];
+                $replace = [
+                    '[image',
```
chillu: Images can have further attributes (e.g. a width/height)

maxime-rainville: I've added unit tests for this use case.

.travis.yml
maxime-rainville: We just need CMS I think because we're using SiteTree objects for testing.

tests/php/TagsToShortcodeHelperTest.php
```diff
@@ -0,0 +1,85 @@
...
+        return ASSETS_PATH . '/TagsToShortcodeHelperTest';
+    }
+
+    public function setUp()
+    {
+        parent::setUp();
+        $this->setupAssetStore();
+    }
+
+    private function setupAssetStore()
```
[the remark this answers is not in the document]
maxime-rainville: I've switch the test to use the AssetStore to create the files.

.travis.yml
unclecheese: Right, that's what I'm saying. Why use the whole recipe, rather than just `silverstripe/cms`?

maxime-rainville: Build is complaing that we don't have a `Page` class, which we'll only get if using the recipe.

We can untangle this later.

maxime-rainville:
Test won't be green until https://github.com/silverstripe/silverstripe-framework/pull/8960 is merge.

chillu:
Was it an explicit decision not to rewrite `_versions` tables here? It means the problem will be reintroduced whenever somebody rolls back to an older version. And there's some inconsistency between the "latest version" for draft and live stages, and the content in the draft and live tables.

If we're not planning to fix that, it's definitely a gotcha we should mention in the upgrade guide. This is modifying people's data, it's absolutely essential that they understand what's happening.

tests/php/Dev/Tasks/TagsToShortcodeHelperTest.php
```diff
@@ -0,0 +1,429 @@
...
+        Filesystem::removeFolder($this->getBasePath());
+        parent::tearDown();
+    }
+
+    public function testRewrite()
+    {
+        $tagsToShortcodeHelper = new TagsToShortcodeHelper();
+        $tagsToShortcodeHelper->run();
+
+        /** @var SiteTree $newPage */
```
chillu: Just as a sidenote: I think this isn't great usage of fixtures - it makes it quite hard to see the before/after state. I generally try to avoid fixtures where the data of those records is relevant to understanding the tests, and default to creating them inline. No need to change here, just thought it's worth mentioning.

maxime-rainville:
@chillu We asked the questions in a stand up a wee while ago and you said it wasn't worth it, so we didn't do it (that will be my go-to excuse from now on ... "we told you about in some unspecified meeting a few months ago").

If you think it's worthwhile, it would not be a big deal to apply it to previous versions as well, although it becomes a bit more murky because some of the files might not be there any more or the DB structure could have changed somewhere in-between. Also, depending on how many previous versions they are, it could take a lot longer to apply the changes.

tests/php/Dev/Tasks/TagsToShortcodeHelperTest.php
```diff
@@ -0,0 +1,429 @@
...
+        $newPage = $this->objFromFixture(SiteTree::class, 'page1');
+        $documentID = $this->idFromFixture(File::class, 'document');
+        $imageID = $this->idFromFixture(Image::class, 'image1');
+
+        $this->assertContains(
```
chillu: Each of these assertions should spell out what they're testing

maxime-rainville:
Grrr ... we still have test failings!!! I'll fix those.

maxime-rainville:
Those unit test should be working now.

PR #228, opened 2019-04-03
Add periodic logging every 100th file iterated

bergice:


src/FileMigrationHelper.php
```diff
@@ -56,27 +61,40 @@ public function run($base = null)
         Environment::increaseTimeLimitTo();
         Environment::increaseMemoryLimitTo();
 
+        $this->logger = Injector::inst()->get(LoggerInterface::class);
```
dnsl48: I guess if we move this one to the constructor, then other methods of the class should be able to reuse the logger. Otherwise we don't have any benefits from keeping the logger as a property and could just hold in within a local variable.

src/FileMigrationHelper.php
```diff
@@ -56,27 +61,40 @@ public function run($base = null)
...
-        foreach ($this->getFileQuery() as $file) {
+        $query = $this->getFileQuery();
+        $total = $query->count();
+        foreach ($query as $file) {
             // Bypass the accessor and the filename from the column
             $filename = $file->getField('Filename');
 
             $success = $this->migrateFile($base, $file, $filename);
+
+            if ($i % 100 == 0) {
```
dnsl48: Should we go for `===` in sake of hygiene? :)

src/FileMigrationHelper.php
```diff
@@ -56,27 +61,40 @@ public function run($base = null)
...
+                /** @var LoggerInterface $logger */
+                if ($this->logger != null) {
```
dnsl48: Same here. Either `!== null`, or `if ($this->logger) {`

src/FileMigrationHelper.php
```diff
@@ -56,27 +61,40 @@ public function run($base = null)
...
+                    $this->logger->info("Iterated $i out of $total files. Migrated $migrated files.");
```
ScopeyNZ: Personally I think this line could just read `Injector::inst()->get(LoggerInterface::class)->info([...]);`. Injector throws an exception if the service cannot be found (and the earlier call wasn't handling that either). This saves adding a private variable and the checks.

dnsl48: nice catch!
though, I think we don't want to abort the whole thing in case the logger hasn't been registered

src/FileMigrationHelper.php
dnsl48: I would imagine, the constructor could be receiving a custom logger and go for the global one as a fallback if null passed. inversion of control ftw

src/FileMigrationHelper.php
robbieaverill: Chiming in to say that currently there'll always be a LoggerInterface defined (by core). In future (hopefully) we'll have it separated so that LoggerInterface by default is a user-defined logger, which will also exist (see https://github.com/silverstripe/silverstripe-framework/pull/8241). When that PR gets merged it should have a NullHandler by default on the LoggerInterface.

I'm fine with having the logger injected via Injector DI, you shouldn't need the truthy check though, but I'm mostly just mentioning this for reference - don't consider this a review =)

src/FileMigrationHelper.php
```diff
@@ -34,6 +36,9 @@ class FileMigrationHelper
      */
     private static $delete_invalid_files = true;
 
+    /** @var LoggerInterface */
+    private $logger;
```
dnsl48: no need to keep it in the properties then?

src/FileMigrationHelper.php
```diff
@@ -56,27 +58,37 @@ public function run($base = null)
...
+
+            if ($i % 100 === 0) {
+                $logger->info("Iterated $i out of $total files. Migrated $migrated files.");
+            }
+
+            $i++;
             if ($success) {
-                $count++;
+                $migrated++;
             }
```
dnsl48: could add a log worning about the file that wasn't a success?

bergice: I'm a bit worried this could cause problems if it skips a lot of files. We don't want to overload the log output.

dnsl48: though users would be able to control it with the log level?

bergice: Alright, I'll add a warning.

src/FileMigrationHelper.php
```diff
@@ -56,27 +58,40 @@ public function run($base = null)
...
+            if ($i % 100 === 0) {
+                $logger->info("Iterated $i out of $total files. Migrated $migrated files.");
+            }
+
+            $i++;
             if ($success) {
-                $count++;
+                $migrated++;
+            } else {
+                $logger->warning("$filename was not migrated");
```
ScopeyNZ: Do you think it makes more sense to add this logging in the `migrateFile` task where it returns `false`? Then you can say "File was not migrated: could not locate file on the filesystem" or "File does not have a whitelisted extension"

src/FileMigrationHelper.php
```diff
@@ -56,27 +58,40 @@ public function run($base = null)
         Environment::increaseTimeLimitTo();
         Environment::increaseMemoryLimitTo();
 
+        /** @var LoggerInterface $logger */
+        $logger = Injector::inst()->get(LoggerInterface::class);
+
         // Loop over all files
-        $count = 0;
+        $i = $migrated = 0;
```
ScopeyNZ: Minor detail, I prefer `$count` > `$i`. But `$processedCount` and `$migratedCount` would be even better.

src/FileMigrationHelper.php
dnsl48: That would spread the logging concern around 2 methods, whereas now it seems to be contained in a single place.
I'd guess it would make sense to either move/add both - positive and negative logging in there, or just leave it here so that migrateFile doesn't contain logging concern at all.

dnsl48:
@robbieaverill @ScopeyNZ good to merge?

kinglozzer:
Looks good, I definitely missed this when upgrading a site this week! Could you squash the commits @bergice? ☺️

bergice:
Squashed, and all good to go! 🚀

src/FileMigrationHelper.php
```diff
@@ -57,26 +73,36 @@ public function run($base = null)
...
-        foreach ($this->getFileQuery() as $file) {
+        $query = $this->getFileQuery();
+        $total = $query->count();
+        foreach ($query as $file) {
             // Bypass the accessor and the filename from the column
             $filename = $file->getField('Filename');
 
             $success = $this->migrateFile($base, $file, $filename);
+
+            if ($processedCount % 100 === 0) {
```
chillu: Are we concerned about database record size on a queuedjobs descriptor here? I think it should have a verbose mode where it logs every file move. It'll be an essential asset for later in case you have to diagnose what went wrong during a migration.

src/FileMigrationHelper.php
```diff
@@ -34,6 +35,21 @@ class FileMigrationHelper
      */
     private static $delete_invalid_files = true;
 
+    private static $dependencies = [
```
chillu: This doesn't log on CLI by default - we need to cover *both* use cases

chillu: Actually, three: Queuedjob (through CLI), direct CLI run, and web-based run. This shouldn't be down to the task or task helper to sort out, but unfortunately the toplevel `TaskRunner` isn't great (https://github.com/silverstripe/silverstripe-framework/issues/5542)

src/FileMigrationHelper.php
bergice: It already tells you why a file was not migrated if that happens.

src/FileMigrationHelper.php
```diff
@@ -57,26 +91,36 @@ public function run($base = null)
...
-        foreach ($this->getFileQuery() as $file) {
+        $query = $this->getFileQuery();
+        $total = $query->count();
+        foreach ($query as $file) {
             // Bypass the accessor and the filename from the column
             $filename = $file->getField('Filename');
 
             $success = $this->migrateFile($base, $file, $filename);
+
+            if ($processedCount % static::$log_interval === 0) {
```
ScopeyNZ:
```suggestion
            if ($processedCount % $this->config()->get('log_interval') === 0) {
```

src/FileMigrationHelper.php
```diff
@@ -34,6 +38,29 @@ class FileMigrationHelper
      */
     private static $delete_invalid_files = true;
 
+    /**
+     * Specifies the interval for every Nth file looped at which output will be logged.
+     *
+     * @config
+     * @var bool
```
ScopeyNZ:
```suggestion
     * @var int
```

src/FileMigrationHelper.php
```diff
@@ -34,17 +36,32 @@ class FileMigrationHelper
...
     /**
      * Perform migration
      *
      * @param string $base Absolute base path (parent of assets folder). Will default to PUBLIC_PATH
      * @return int Number of files successfully migrated
+     * @throws \Exception
      */
     public function run($base = null)
     {
+        $this->logger = Injector::inst()->get(LoggerInterface::class);
```
dnsl48: Sorry to be pain, but for consistency properties should be initialised in the constructor.

robbieaverill: Jumping in as devil's advocate - yes, agreed, but SilverStripe's dependency injection (which we should try and use as much as possible) is lazily injected, not at construction. We should be careful not to promote two different design patterns in core

robbieaverill: More specific example of this would be that there's a setter for the logger and you define `private static $dependencies = ['Logger' => LoggerInterface::class]`

dnsl48: As long as we keep the thing in a property it should be initialised in constructor (or through `private static $dependencies`). Otherwise, we end up with inconsistent state of the class, depending on the order of its method invocations.

> there's a setter for the logger and you define private static $dependencies

I guess that's good to have. However, depending on whether we want this class to be only available as DI service or constructable entity as well.

robbieaverill: Yeah that’s what I mean. While I’m not a fan of the DI we promote in framework, it is our recommended solution so I think we should continue to promote it until it gets replaced with something constructor based

dnsl48: yeah. I'm not saying we should stop using Injector :) I'm rather saying that initialisation should be done in constructor (or by DI through a setter), not in the `run` method.

bergice: I was having issues with $dependencies earlier. Seems to work now though, so I've made the necessary changes. :)

dnsl48:
@chillu @ScopeyNZ ping

PR #239, opened 2019-04-10
Travis | add php versions, add postgres tests

dnsl48:


dnsl48:
Actually, let me better do it for 1.0 and then we can merge it up

PR #240, opened 2019-04-10
Travis | fix php versions, code coverage

dnsl48:


.travis.yml
```diff
@@ -11,26 +11,34 @@ matrix:
...
+      env: DB=MYSQL PHPUNIT_TEST=1
+    - php: 7.3
       env: DB=PGSQL PHPUNIT_TEST=1
 
 before_script:
+# Init PHP
   - printf "\n" | pecl install imagick
   - composer self-update || true
   - phpenv rehash
-  - phpenv config-rm xdebug.ini
```
robbieaverill: This is deliberate - xdebug slows phpunit down a LOT, and we use phpdbg for code coverage runs instead in PHP 7

dnsl48: that's actually been the fix I had to do to fix coverage generation in the branch `4`. not sure it's gonna work without it, but can check it in another branch

robbieaverill: PHP 7 has phpdbg bundled with it, which is why we usually run coverage builds on PHP 7 - most core and supported modules already have this config in place if you need a reference

dnsl48: ok, it's working on this branch without xdebug. Will put it back then and we'll keep an eye on codecov when merge it up

dnsl48: > that's actually been the fix I had to do to fix coverage generation in the branch 4

yeah, my bad, I'm clumsy today...

PR #245, opened 2019-04-16
Update File Migration Task to work with FileResolutionStrategy

maxime-rainville:
Follow up PR to #223 

# Parent issue


src/ImageManipulation.php
```diff
@@ -871,6 +871,8 @@ function (AssetStore $store, $filename, $hash, $variant) use ($callback) {
                         $tuple = $store->setFromStream($result->getStream(), $filename, $hash, $variant);
                         return [$tuple, $result];
                     } finally {
+                        // Unload the Intervention Image resource so it can be garbaged collected
+                        $res = $backend->setImageResource(null);
```
maxime-rainville: This bit improves memory usages when generating lots of thumbnails with the GD image driver.

It will however alter the behaviour everywhere else, so you can't test it just on the migration script.

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -0,0 +1,351 @@
...
+        $filename = $file->generateFilename();
+        $results = $this->store->normalisePath($filename);
+
+        // Move file if the APL changes filename value
+        $file->File->Filename = $results['Filename'];
+        $file->File->Hash = $results['Hash'];
+
+
+        // Save and publish
+        $file->write();
```
unclecheese: Should we wrap this in a try/catch? In my testing, there were files that didn't meet the allowed_extension requirements that had gotten into the system one way or another. Recommend we capture the validation exception and log it out.

Issue #235, opened 2019-04-09
File Migration: Reuse existing resized images


chillu:
## Overview

 * As a developer migrating from 3.x to 4.x, I want to minimise the time spent on migration tasks, in order to de-risk my go-live plan
 * As a sysadmin migrating from 3.x to 4.x, I want to keep the system requirements at the same level in order to avoid provisioning more costly resources (e.g. more disk space)
 * As a CMS author using an upgraded 4.x CMS, I want to manage assets without being slowed down by system tasks

## Acceptance Criteria

 * The file migration task reuses existing resized images on the filesystem (and moves/renames them as required)
 * The file migration task generates new resized images where required (CMS thumbnails), but allows developers to opt out of this
 * The file migration task does not increase the disk space required for assets

## Notes

 * Assumes that missing CMS thumbnails are lazy generated on demand
 * We'll need to update the [file migration reference](https://docs.silverstripe.org/en/4/developer_guides/files/file_migration/#migration-of-thumbnails) to thumbnail behaviour

## Pull Requests

 * [x] https://github.com/silverstripe/silverstripe-framework/pull/8924
 * [x] https://github.com/silverstripe/silverstripe-assets/pull/242

chillu:
Testing approach for this. The frameworktest module can generate thumbnails in 3.x.

**On a 3.x install**

- `composer config repositories.frameworktest vcs https://github.com/open-sausages/silverstripe-frameworktest.git`
- `composer require --dev "silverstripe/frameworktest:dev-pulls/backport-filemakertask as 3.x-dev"`
- `framework/sake dev/tasks/FTFileMakerTask`
- Open a page, insert an image, resize it and publish
- Open a page, insert a link to a file, and publish
- Export SQL

**On a 4.x install**

- Delete current database
- Delete current `public/assets` folder
- Create empty new database
- Import SQL into new database
- Copy 3.x `assets/` folder into `public/assets`
- Run `dev/build`
- Run `vendor/bin/sake dev/tasks/MigrateFileTask`

At the moment, you can see the `_resampled` folders with the original resized files in their original location. If you re-publish the page with the image, it keeps this old location. It has lost the reference to the original image, which we're fixing with https://github.com/silverstripe/silverstripe-assets/issues/167

chillu:
Some more details on how I've manually tested this. One combined test for the opt-in for `legacy_thumbnails=true`, and testing that newly generated thumbnails actually work.

In 3.x:

Perform all steps above

[...]

PR #250, opened 2019-04-30
Add `FixFilePermissionsTask` to fix up `NULL` file permissions from SS3

bergice:
This just resets the view enum to `Inherit` for records that have it set to `NULL`.

This task will be combined with other file migration tasks at some point so we don't have too many of them floating around.

This should also just be necessary after upgrading from SS3 as it is a datafix.


chillu:
This should be a subtask of `MigrateFileTask`, for the same reasons as thumbnail and shortcode migration subtasks. This is part of the overall "4.x file migration", but in some cases you just want to run those tasks in isolation. I don't want people doing 3.x -> 4.x upgrades to start off with running a whole bunch of arbitrary CLI commands to get started :) Can you please rewrite this to be a subtask, based on the work in https://github.com/silverstripe/silverstripe-assets/issues/235

src/Dev/FixFilePermissionsTask.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace SilverStripe\Dev\Tasks;
+
+use SilverStripe\Assets\File;
+use SilverStripe\Dev\BuildTask;
+use SilverStripe\ORM\Queries\SQLUpdate;
+use SilverStripe\Security\InheritedPermissionFlusher;
+
+/**
+ * Files imported from SS3 might end up with broken permissions if there is a case conflict.
```
chillu: Should mention that this fixes issues with https://github.com/silverstripe/silverstripe-secureassets

src/Dev/FixFilePermissionsTask.php
```diff
@@ -0,0 +1,39 @@
...
+    Reset the permission of files that have a `NULL` CanViewType value.
+    You need to flush your cache after running this task via CLI.
+    ';
+
+    public function run($request)
+    {
+        SQLUpdate::create(
+            File::singleton()->baseTable(),
+            ['CanViewType' => 'Inherit'],
+            ['ISNULL(CanViewType)']
```
chillu: Should filter by `Folder` class name

src/Dev/FixFilePermissionsTask.php
```diff
@@ -0,0 +1,39 @@
...
+    private static $segment = 'FixFilePermissionsTask';
+
+    protected $title = 'Fix File Permissions';
+
+    protected $description = '
+    Reset the permission of files that have a `NULL` CanViewType value.
+    You need to flush your cache after running this task via CLI.
+    ';
+
+    public function run($request)
```
chillu: Should have some logging - e.g. `Repaired X folders with broken CanViewType settings`. Running two raw SQL queries (one count, one update) is fine for that

tests/php/FixFilePermissionsTaskTest.yml
```diff
@@ -0,0 +1,31 @@
...
+  parent:
+    Name: ParentFolder
+    CanViewType: NULL
+  subfolder:
+    Name: SubFolder
+    Parent: =>SilverStripe\Assets\Folder.parent
+SilverStripe\Assets\Image:
+  image1:
+    Name: myimage.jpg
+    CanViewType: LoggedInUsers
```
chillu: Only folders can have `CanViewType`, at least when they're using https://github.com/silverstripe/silverstripe-secureassets. Files can't. While there could be other ways in which files have data in this column, it's not in scope here - we're only fixing something which a supported module broke, and then had downstream effects on core file migrations.

bergice:
> This should be a subtask of `MigrateFileTask`, for the same reasons as thumbnail and shortcode migration subtasks. This is part of the overall "4.x file migration", but in some cases you just want to run those tasks in isolation. I don't want people doing 3.x -> 4.x upgrades to start off with running a whole bunch of arbitrary CLI commands to get started :) Can you please rewrite this to be a subtask, based on the work in #235

@chillu should we merge https://github.com/silverstripe/silverstripe-framework/pull/8924/files first or do you want me to cherry-pick/rebase on your branch?

chillu:
I'd say rebase on that PR, rather than cherry pick. Don't let the https://github.com/silverstripe/silverstripe-framework/pull/8924 merge block you for now.

PR #251, opened 2019-04-30
More verbose file migration logging

chillu:
The log interval was useful to see errors in the noise,
but it's also masked what the task actually does.
Since this is moving and modifying project data,
it should be verbose by default. This will help answer
question later on, e.g. "why wasn't this file migrated".

A new message at the bottom which spells out how many files failed (if any)
makes it clear that you need to review the logs in detail (even though the error lines might hide between a lot of success lines).

### DIFF
--- a/_config/migration.yml
+++ b/_config/migration.yml
@@ -1,0 +1,11 @@
+---
+Name: file-migration
+---
+# Remove the HTTPOutputHandler from the logger on the file migration.
+# It logs a lot of >= NOTICE errors that pollute the output with backtrace.
+SilverStripe\Core\Injector\Injector:
+  Psr\Log\LoggerInterface.quiet:
+    type: singleton
+    class: Monolog\Logger
+    constructor:
+      - "file-migration"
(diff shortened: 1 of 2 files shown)


unclecheese:
There are default logging handlers (StreamHandler). But by not calling `pushHandler [ HTTPOutputHandler ]` on this implementation, we can limit a lot of the noise.

robbieaverill:
Oh, TIL. /me goes and updates a module where it shouldn't have any handlers by default to have a NullHandler =D

unclecheese:
Yeah it's a bit of magic in `Monolog\Logger`  ¯\_(ツ)_/¯

```php
    public function addRecord($level, $message, array $context = array())
    {
        if (!$this->handlers) {
            $this->pushHandler(new StreamHandler('php://stderr', static::DEBUG));
        }
```

robbieaverill:
@unclecheese FYI the default LoggerInterface implementation in core is now a user logger rather than a core logger, and doesn't have any handlers attached by default. Do you see any issue with having a StreamHandler for it? 

Context: https://github.com/silverstripe/silverstripe-framework/blob/4/_config/logging.yml#L19-L25

PR #254, opened 2019-04-30
NEW Clean up secureassets module artefacts (fixes #231)

### LINKED ISSUE #231: Secureassets migration does not remove .htaccess files
## Overview

As a developer migrating a site with [secure assets](https://github.com/silverstripe/silverstripe-secureassets) to SilverStripe 4.x, I want to be confident that files can be accessed and restricted in the same way

## Acceptance Criteria

 * Files in folders with specific viewing permissions are still access restricted after the upgrade (should be the case already)
 * Files in folders with specific viewing permissions can be accessed with the right permissions after the upgrade (that's currently not working)
 * Can be run as one file migration task if upgrading from 3.x
 * Can be run as individual task if upgrading from a 4.x minor
 * Custom htaccess files are kept in place

## Notes

 * The [upgrade instructions](https://github.com/silverstripe/silverstripe-secureassets/commit/bcc33303d41ed8cf3620f827fdc3e9b4b3df8d32) on the module are misleading there
 * I think the issue here is that the folder specific `.htaccess` file created through [onAfterWrite()](https://github.com/silverstripe/silverstripe-secureassets/blob/master/code/SecureFileExtension.php#L191) isn't removed through the [FileMigrationHelper](https://github.com/silverstripe/silverstripe-assets/blob/1/src/FileMigrationHelper.php)
 * Since CWP 1.x recipes all include secureassets by default, and authors can choose to enforce it through the UI, devs upgrading a site have to assume that it includes access restricted folders
 * Check through `.htaccess` template history in secureassets to see if that template content was changed

## Pull Requests

 * [x] https://github.com/silverstripe/silverstripe-framework/pull/8948
 * [x] https://github.com/silverstripe/silverstripe-assets/pull/254

chillu:
See htaccess file contents at https://github.com/silverstripe/silverstripe-secureassets/blob/f9d54e7b419574e583f3b958a2f21db163877971/_config.php#L9

Merge after https://github.com/silverstripe/silverstripe-framework/pull/8948

~Rebased on commits from https://github.com/silverstripe/silverstripe-assets/issues/235 since that's required to split up the migration tasks~

This task can't go into the secureassets module, since it needs to be run *after* dev/build on a 4.x compatible codebase. The secureassets module isn't 4.x compatible, so we can't execute code from it. Also, the current `MigrateFileTask` isn't extensible on that level, and I'm not really keen to do that

### DIFF
--- a/src/Dev/Tasks/SecureAssetsMigrationHelper.php
+++ b/src/Dev/Tasks/SecureAssetsMigrationHelper.php
@@ -1,0 +1,175 @@
+<?php
+
+namespace SilverStripe\Assets\Dev\Tasks;
+
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use League\Flysystem\Filesystem;
+use SilverStripe\Assets\Flysystem\FlysystemAssetStore;
+use SilverStripe\Assets\Folder;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\ORM\Queries\SQLSelect;
+
+/**
+ * Removes stray .htaccess files created through the silverstripe/secureassets module
+ * on a 3.x-based site. The 4.x protections work differently:
+ * One central assets/.htaccess file routes non-existent paths through SilverStripe,
+ * which can choose to return a file from assets/.protected.
+ * Any additional .htaccess files in folders can interfere with this logic.
+ *
+ * Note that this task does not migrate file metadata added/managed through silverstripe/secureassets.
+ * The metadata fields are the same in 4.x (File.CanViewType etc).
+ *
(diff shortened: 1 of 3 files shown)


src/Dev/Tasks/SecureAssetsMigrationHelper.php
```diff
@@ -0,0 +1,172 @@
...
+    ];
+
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct()
+    {
+        $this->logger = new NullLogger();
+
+        $this->htaccessRegexes = [
```
chillu: I've compared the [history](https://github.com/silverstripe/silverstripe-secureassets/commits/f9d54e7b419574e583f3b958a2f21db163877971/_config.php) of the original htaccess config, and that's the only format used by secureassets

PR #260, opened 2019-05-07
BUG Escape table name and column with double quote to fix PostgreSQL

maxime-rainville:
PostgreSQL test were failing because of some missing double quotes.

I've also remove a try/catch to avoid hiding errors.

I'm still looking at how to fix the Coverage memory failure, so maybe hold back approving until then.

### DIFF
--- a/src/Dev/Tasks/TagsToShortcodeHelper.php
+++ b/src/Dev/Tasks/TagsToShortcodeHelper.php
@@ -21,6 +21,7 @@ use SilverStripe\ORM\Connect\DatabaseException;
 use SilverStripe\ORM\Connect\Query;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectSchema;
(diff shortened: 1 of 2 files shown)


src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -123,15 +127,18 @@ public function run()
                         continue;
                     }
 
-                    try {
-                        // Update table
-                        $this->updateTable($table, $field);
+                    if (!isset($tableList[strtolower($table)])) {
```
maxime-rainville: That's just for PHPUnit. Not all tables get created when running tests, but the task will still try to run queries against them.

Originally we had a try/catch for this but that was hidding errors.

maxime-rainville:
I just increased the memory available to the build from 2G to 3G. That seems that have shut up PHPUNIT_COVERAGE_TEST.

PR #261, opened 2019-05-07
BUG Avoid locking framework version when installing postgresql module

maxime-rainville:
The travis config was installing silverstripe/postresql right away which would install and lock framework at 4.3 and would break the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_script:
   - composer validate
   - composer install --prefer-dist
   - composer require --prefer-dist --no-update silverstripe/recipe-core:1.1.x-dev silverstripe/versioned:1.1.x-dev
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -38,7 +38,7 @@ before_script:
   - composer validate
   - composer install --prefer-dist
   - composer require --prefer-dist --no-update silverstripe/recipe-core:1.1.x-dev silverstripe/versioned:1.1.x-dev
-  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.0.x-dev --prefer-dist; fi
+  - if [[ $DB == PGSQL ]]; then composer require --prefer-dist --no-update silverstripe/postgresql:2.0.x-dev --prefer-dist; fi
```
dnsl48: `--prefer-dist` is repeated twice

PR #262, opened 2019-05-07
BUGFIX: File::canView() not considerate of ADMIN privileges

unclecheese:
The `canView()` API had no test coverage and was allowing for scenarios where non-admin users could set permissions that made it impossible for admins to view or edit their files in the CMS.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -366,6 +366,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
             return $result;
         }
 
+        if (Permission::checkMember($member, 'ADMIN')) {
+        	return true;
+        }
+
         // Check inherited permissions
         return static::getPermissionChecker()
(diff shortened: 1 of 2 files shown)


chillu:
So this is fixing the case where non-admins restricted access to a specific group via the CMS UI? Yeah that makes sense, users with ADMIN permissions shouldn't be affected by those restrictions. Presumably that brings the logic in line with existing permission checks for SiteTree?

dnsl48:
Since ADMIN has complete access to everything by definition, shall we move this check up to the permission checker implementation rather than keeping it here in this local place?

unclecheese:
I think that's a far too broad based change to put into a patch release. That could in theory break people's code. 

We have all kinds of imperative `ADMIN` checks all over the place that would become redundant. It's a good idea, but feels like a separate story to me.

PR #263, opened 2019-05-07
BUG Make sure findVariants works for hashless tuples and always return an hash with the variants

maxime-rainville:
When normalising files via their path, sometimes the file hash can not be inferred from their file ID.

This lead to 2 problems:
* Some file ID Helper require an Hash to build a URL, which will lead to an InvalidArgumentException
* The targeted file ID may require an Hash to be generated.

This PR address these problems by:
* catching the invalid argument exception when trying to build a file ID with an helper that doesn't support hashless URLs.
* Compiling the hash of the file on the fly if it wasn't provided initially.


unclecheese:
LGTM

PR #270, opened 2019-05-15
Do not run legacy thumbnail migration if site does not have `File.Filename` DB column

bergice:
Without this, an exception will be thrown when running file migrations on new 4.x sites.


maxime-rainville:
@bergice Can you update your commit message to use the BUG prefix?

Issue #269, opened 2019-05-15
File Migration Tasks confuses 10-char-long folders for hash folder


maxime-rainville:
If you are migrating a file such as `multimedia/video.mp4`, the file migration script will get confused and think that `multimedia` is the hash for `video.mp4`.

# Subtasks
* [x] Update `HashFileIDHelper` to only accept hexa-decimal character in the hash folder
* [x] Update file resolution mechanism to validate that file matches hash ... this may be opening a can of worms as it might require us to read the entire file every time we serve it to make sure its actual hash matches the expected hash
* [x] Look back at the rest of the uni test and add some extra edge cases to cover ambiguous folder names.

# Parent PR
* https://github.com/silverstripe/silverstripe-assets/pull/274

maxime-rainville:
The changes introduce by https://github.com/silverstripe/silverstripe-assets/pull/267 will completely negated this issue for SS3 migration. When migrating SS3 files, the FileMigrationHelper will only use the LegacyFileIDHelper now.

I'll add an extra bit of logic to make sure SS4.3 to SS4.4 migration works as well.

bergice:
Can confirm this works.

PR #274, opened 2019-05-21
Don't confuse 10-char folder with hash

maxime-rainville:
This PR tweaks the logic that resolve file ID to make sure that when the hash is found from the file ID, it is matches the actual hash of the file.

# Parent issue 
https://github.com/silverstripe/silverstripe-assets/issues/269

### DIFF
--- a/src/FilenameParsing/HashFileIDHelper.php
+++ b/src/FilenameParsing/HashFileIDHelper.php
@@ -79,7 +79,7 @@ class HashFileIDHelper implements FileIDHelper
 
     public function parseFileID($fileID)
     {
-        $pattern = '#^(?<folder>([^/]+/)*)(?<hash>[a-zA-Z0-9]{10})/(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
+        $pattern = '#^(?<folder>([^/]+/)*)(?<hash>[a-f0-9]{10})/(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
 
         // not a valid file (or not a part of the filesystem)
         if (!preg_match($pattern, $fileID, $matches)) {
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -40,6 +40,13 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
      */
     private $versionedStage = Versioned::DRAFT;
 
+    /**
+     * todo: needs documenting
+     *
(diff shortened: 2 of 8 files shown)


maxime-rainville:
A thing to note is that this will slow down some request because we'll have to recompile the hash of files pretty much every time we want to serve a protected file.

dnsl48:
>  we'll have to recompile the hash of files pretty much every time 

why don't we keep once precalculated hash in the filename by the variant?

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -79,7 +79,7 @@ public function cleanFilename($filename)
 
     public function parseFileID($fileID)
     {
-        $pattern = '#^(?<folder>([^/]+/)*)(?<hash>[a-zA-Z0-9]{10})/(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
+        $pattern = '#^(?<folder>([^/]+/)*)(?<hash>[a-f0-9]{10})/(?<basename>((?<!__)[^/.])+)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#';
```
dnsl48: if we kept hash in the file ID by the variant, we could make the pattern even narrower, by comparing the hash through a back reference. E.g.:
```php
'#^(?<folder>([^/]+/)*)(?<hash>[a-f0-9]{10})/(?<basename>((?<!__)[^/.])+)(?:__\1)(__(?<variant>[^.]+))?(?<extension>(\..+)*)$#'
```

maxime-rainville:
> @maxime-rainville why don't we keep once pre-calculated hash in the filename by the variant?

The core issue I'm trying to fix here is that you could in some context the folder name can be confused for an hash. 

e.g.: Let's say you have the following path: `decade1980/document.doc`. In this case, `decade1980` could be a valid hash but it could also be a folder name.

I was trying to fix this for file migration, but this could also affect other operations.

dnsl48:
> e.g.: Let's say you have the following path: decade1980/document.doc. In this case, decade1980 could be a valid hash but it could also be a folder name.

yes, understood. I'm thinking that we wouldn't have to calculate the file hash every time if it had it precalculated in the name.
Might not worth the effort, but the idea is
```
decade1980/document__decade1980.doc  - looks suspicious, so we can recalculate its hash to check if it's actually protected file

decade1980/document__aoeu.doc - hashes don't match, don't have to recalculate hash

decade1980/document.doc - doesn't have cache, not protected, don't have to recalculate its hash
```

maxime-rainville:
That wouldn't be possible to achieve in time for the 4.4.0 release. The file hash caching PR could help here ... I've only enabled caching for migration, because I didn't want to wrestle with everything thing else at this time, but we could try enabling in 4.5.0

Issue #273, opened 2019-05-20
FileMigrationHelper: Issue with case sensitive folder paths


ichaber:
## Affected Version

This issue was found on SilverStripe 4.4-rc1 (4a0f62fafd13bfc90a1498a632a6fec9ea27d039). 

## Description

Files in a folder with an uppercase name and an entry in the database with a lowercase folder will not be migrated when running the `MigrateFileTask`. That is, because the look up in a case-sensitive file system (e.g. `ext4`) will not find the file in a lower case folder. 

Furthermore this leads to a collision, if two folder have the same name with only the difference being the case and two different entries in the database. 

## Steps to Reproduce

1. Place an image in a folder with an uppercase name (e.g. assets/**P**lanning/diagram.jpg)
2. Create a corresponding entry in the `File` table with a `Filename` having the folder in a lowercase format (e.g. assets/**p**lanning/diagram.jpg)
3. Running the `/dev/tasks/MigrateFileTask` task.
3. Observe, that the file will not be found and hence not migrated.

## Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-assets/pull/277

maxime-rainville:
So I did a quick PR that adds some test case around file names with conflictual/contradicatory filenames. https://github.com/silverstripe/silverstripe-assets/pull/277 

Here's what I'm testing and the end results.

DB | Filesystem | End results
-- | ---------- | -----------
wrong-case.txt | wRoNg-CaSe.tXt | Fails to import
uploads/good-case-bad-folder.txt | uploads/good-case-bad-folder.txt | Fails to import
Uploas/good-case-good-folder.txt | Uploas/good-case-good-folder.txt | Success
MIXED-CASE-FILE.TXT | MIXED-CASE-FILE.TXT | This is the first file, it imports successfully
mixed-case-file.txt | mixed-case-file.txt | Seems to import correctly, but it probably should be renamed mixed-case-file-v2.txt to avoid conflicts with the previous file.

This was on a linux box. The first 2 cases might have worked on a Windows system because they are case insensitive. Not sure about `mixed-case-file.txt` and if that scenario is even possible on Windows.

maxime-rainville:
I had a quick look to see if there's a way to do a case insensitive file existence check on Flysystem ... it doesn't look like there is.

The only way I could do this is to load the entire file system store tree and lowercase it and do my own comparison to the expected file name from the DB.

chillu:
[...]

PR #277, opened 2019-05-23
Update file migration to work with mis-match case filenames

maxime-rainville:
In some weird and convoluted situation, the physical file might use a different cased filename than what's in the Database.

This PR adds extra check to make a best effort attempt to resolve those missing files.

Basically we check for alternative spelling of the filename. If we find an alternative file, we assume it's meant to match our entry in the DB under these condition:
* There's only one possible match (if there's more than one possible match, we bail and curse your wicked name)
* No other DB records (migrated or not) match the exact spelling of our alternative file

# Still not workging perfectly

In a perfect world, the migration script would rename files that have clashing filename to avoid having any case-insenstive clash.

That's kind of difficult right now because the `File::onBeforeWrite()` logic right now is somewhat convoluted and I'm not overly eager to mess around with it the day before a release.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/273

### DIFF
--- a/src/Dev/Tasks/FileMigrationHelper.php
+++ b/src/Dev/Tasks/FileMigrationHelper.php
@@ -91,6 +91,15 @@ class FileMigrationHelper
             $base = PUBLIC_PATH;
         }
 
+        // Set max time and memory limit
+        Environment::increaseTimeLimitTo();
+        Environment::setMemoryLimitMax(-1);
+        Environment::increaseMemoryLimitTo(-1);
+        
+        /** @var FileHashingService $hasher */
+        $hasher = Injector::inst()->get(FileHashingService::class);
+        $hasher::flush();
+
         $this->logger->notice('Step 1/2: Migrate 3.x legacy files to new format');
         $ss3Count = $this->ss3Migration($base);
 
@@ -332,8 +341,8 @@ class FileMigrationHelper
             $this->logger->warning(sprintf(
                 '* SS3 file %s converted to SS4 format but was renamed to %s',
                 $dbFilename,
-                $file->getFilename())
-            );
+                $file->getFilename()
+            ));
         }
 
         if (!empty($results['Operations'])) {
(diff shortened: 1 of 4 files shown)


src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -118,7 +118,7 @@ private function resolveWithHash(ParsedFileID $parsedFileID)
         /** @var File $file */
         $file = Versioned::withVersionedMode(function () use ($parsedFileID) {
             Versioned::set_stage($this->getVersionedStage());
-            return File::get()->filter(['FileFilename' => $parsedFileID->getFilename()])->first();
+            return File::get()->filter(['FileFilename:case' => $parsedFileID->getFilename()])->first();
```
maxime-rainville: Changes in this file ar not really related to this PR. I realised that those check were getting confused in some circumstances because the Filename check was not case sensitive.

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -327,6 +351,96 @@ protected function migrateFile($base, File $file, $legacyFilename)
         return true;
     }
 
+    /**
+     * Look for a file by `$legacyFilename`. If an exact match is found return true. If no match is found return false.
+     * If a match using a different case is found use the path of that file.
+     * @param string $base Location of of the assets folder, usually equals to ASSETS_PATH
```
bergice:
```suggestion
     * @param string $base Location of the assets folder, usually equals to ASSETS_PATH
```

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -327,6 +351,96 @@ protected function migrateFile($base, File $file, $legacyFilename)
...
+                $this->logger->error(sprintf(
+                    '%s could not be migrated because no matching file was found. ' .
+                    'Other files with alternative casing exists: %s',
+                    $legacyFilename,
+                    implode(', ', $files)
+                ));
+                return false;
+        }
+
+        // Make sure we do not have a unmigrated or migrated DB entry for or alternative file.
```
bergice:
```suggestion
        // Make sure we do not have an unmigrated or migrated DB entry for our alternative file.
```

PR #278, opened 2019-05-23
API Bump minimum PHP version to 7.1

robbieaverill:
See https://github.com/silverstripe/silverstripe-framework/issues/8556


robbieaverill:
Self merging since the others have been merged already

PR #281, opened 2019-05-28
Better formatting of file migration tasks

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/8962

### DIFF
--- a/src/Dev/Tasks/FileMigrationHelper.php
+++ b/src/Dev/Tasks/FileMigrationHelper.php
@@ -94,30 +94,34 @@ class FileMigrationHelper
         $this->logger->notice('Step 1/2: Migrate 3.x legacy files to new format');
         $ss3Count = $this->ss3Migration($base);
 
-        $this->logger->notice('Step 2/2: Migrate <4.4 files to new format');
+        $this->logger->notice('Step 2/2: Migrate files in <4.4 format to new format');
 
         $ss4Count = 0;
         if (class_exists(Versioned::class) && File::has_extension(Versioned::class)) {
             Versioned::prepopulate_versionnumber_cache(File::class, Versioned::LIVE);
             Versioned::prepopulate_versionnumber_cache(File::class, Versioned::DRAFT);
 
-            $ss4Count += Versioned::withVersionedMode(function () {
-                Versioned::set_stage(Versioned::LIVE);
-                return $this->normaliseAllFiles('on the live stage');
-            });
+            $stages = [Versioned::LIVE => 'live', Versioned::DRAFT => 'draft'];
+            foreach($stages as $stageId => $stageName) {
+                $this->logger->info(sprintf('Migrating files in the "%s" stage', $stageName));
+                $count = Versioned::withVersionedMode(function () use ($stageName) {
+                    Versioned::set_stage(Versioned::LIVE);
+                    return $this->normaliseAllFiles(sprintf('on the "%s" stage', $stageName));
+                });
+                if ($count) {
+                    $this->logger->info(sprintf('Migrated %d files in the "%s" stage', $count, $stageName));
+                } else {
+                    $this->logger->info(sprintf(
+                        'No files required migrating in the "%s" stage',
+                        $stageName
+                    ));
+                }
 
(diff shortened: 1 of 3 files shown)


src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -93,26 +90,21 @@ public function run($base = null)
...
 
-        $this->logger->info('NORMALISE SILVERSTRIPE 4 FILES');
+        $this->logger->notice('Step 2/2: Migrate <4.4 files to new format');
+
         $ss4Count = 0;
         if (class_exists(Versioned::class) && File::has_extension(Versioned::class)) {
             Versioned::prepopulate_versionnumber_cache(File::class, Versioned::LIVE);
             Versioned::prepopulate_versionnumber_cache(File::class, Versioned::DRAFT);
 
-            $this->logger->info('Looking at live files');
```
bergice: Not sure why this is omitted, it could be useful.

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -93,26 +90,21 @@ public function run($base = null)
...
             $ss4Count += Versioned::withVersionedMode(function () {
                 Versioned::set_stage(Versioned::LIVE);
                 return $this->normaliseAllFiles('on the live stage');
             });
 
-            $this->logger->info('Looking at draft files');
```
bergice: Not sure why this is omitted, it could be useful.

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -317,10 +310,10 @@ protected function migrateFile($base, File $file, $legacyFilename)
             return false;
         }
 
-        $this->logger->info(sprintf('* SS3 file %s converted to SS4 format', $file->getFilename()));
+        $this->logger->debug($file->getFilename());
```
bergice: Is this supposed to be on the same line as the next debug messages?

src/Dev/Tasks/FileMigrationHelper.php
chillu: I found it a bit weird to have those two messages when no files needed migrating. I've made it a bit clearer now what the outcome of each of those steps is, and re-introduced the messages. You're right as in there's some value in flagging that both live/draft are looked at separately, otherwise people might get confused why files show up twice in the logs. Made this a bit more verbose and reintroduced those messages.

src/Dev/Tasks/FileMigrationHelper.php
chillu: No, it's the "section headline" so to speak - each file can have one or more thumbnails related to it, which are their own indented log messages.

chillu:
FYI test failure is the same as upstream: https://travis-ci.org/silverstripe/silverstripe-assets/builds/541466837

robbieaverill:
They're also showing up in the CWP kitchen sink. `FileIDHelperResolutionStrategy` uses SilverStripe dependency injection to create its "hasher" service, but in unit tests it's created directly via `new FileIDHelperResolutionStrategy()` which doesn't use SS DI, so there's a null pointer exception.

`FileIDHelperResolutionStrategy` isn't `Injectable`, which means you need to instantiate it either as a dependency of something else or directly via `Injector::inst()->create(FileIDHelperResolutionStrategy::class)` - the latter is a quick fix for unit tests, or add Injectable and do `FileIDHelperResolutionStrategy::create()` or `FileIDHelperResolutionStrategy::singleton()`.

robbieaverill:
The tests were red and this PR was merged anyway, can you please make a follow up to resolve that?

chillu:
The tests were red because of different reasons this morning: https://travis-ci.org/silverstripe/silverstripe-assets/builds/542001598

Then they were fixed this arvo upstream: https://travis-ci.org/silverstripe/silverstripe-assets/builds/542052665 and https://github.com/silverstripe/silverstripe-assets/pull/287

Then the merge went through without conflicts. Turns out that the originally broken tests *masked* other actually broken tests in this build. So yes, we should've picked up that the build was fixed upstream, and re-run the test. But my fault for not running tests locally as the PR author.

Working on a fix now.

PR #284, opened 2019-05-30
BUG Unset memory limit max so we can increase memory limit

maxime-rainville:
Apparently `increaseMemoryLimitTo` doesn't do anything unless you unset memory limit max, which sounds kind of weird.

I think we should look into that as well, but short term, let's just get the file migration to use as much memory as it wants.

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/9029


chillu:
Note that I've moved this to framework as part of https://github.com/silverstripe/silverstripe-framework/issues/8962

PR #287, opened 2019-06-06
BUG Make FileIDHelperResolutionStrategy Injectable.

maxime-rainville:
Sometimes the hasher on FileIDHelperResolutionStrategy is null because `FileIDHelperResolutionStrategy` was not injectable.


robbieaverill:
You'll also need to update the tests that instantiate this class to use `TheClass::singleton()`, or mock its dependencies where it's used

chillu:
@maxime-rainville There's some warnings in the tests: https://travis-ci.org/silverstripe/silverstripe-assets/jobs/542052671

PR #291, opened 2019-06-13
FIX Folder names no longer allow dots, and are replaced with dashes

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/830

### DIFF
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -53,7 +53,7 @@ class Folder extends File
 
         $parentID = 0;
         $item = null;
-        $filter = FileNameFilter::create();
+        $filter = FolderNameFilter::create();
         foreach ($parts as $part) {
             if (!$part) {
                 continue; // happens for paths with a trailing slash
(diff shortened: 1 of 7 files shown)


src/Folder.php
```diff
@@ -109,6 +109,25 @@ public function getIcon()
...
+    /**
+     * Use default File filter rules, but also remove dots and replace with dashes.
+     *
+     * @param string $name
+     * @return string
+     */
+    protected function filterFilename($name)
+    {
+        // Fix illegal characters
+        $filter = FileNameFilter::create();
```
tractorcow: Would you be better off with a FileNameFilter subclass (e.g. named service) custom to folders? This is un-injectable.

robbieaverill: True, fair enough

robbieaverill: I've done this now =)

robbieaverill:
We really need a FilterInterface for these classes, so I've added one for future consumption here: https://github.com/silverstripe/silverstripe-framework/pull/9064

PR #306, opened 2019-08-01
Allow the temporary store for intervention to be configured

adrhumphreys:
The goal for this is to allow devs to configure if they want to use `TEMP_PATH` or another location. 

The idea here being that when you're dealing with a large amount of assets on a small disk you can run out of space quickly which can be elevated by using a different path

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -65,6 +65,11 @@ class InterventionBackend implements Image_Backend, Flushable
     const FAILED_UNKNOWN = 'unknown';
 
     /**
+     * @config - used to set where cached intervention files will be stored
+     */
+    private static $local_temp_path = TEMP_PATH;
+
+    /**
      * @var AssetContainer
      */
     private $container;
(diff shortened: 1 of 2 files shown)


src/InterventionBackend.php
```diff
@@ -64,6 +64,11 @@ class InterventionBackend implements Image_Backend, Flushable
      */
     const FAILED_UNKNOWN = 'unknown';
 
+    /**
+     * @config - used to set where cached intervention files will be stored
```
ScopeyNZ:
```suggestion
     * Configuration for where cached intervention files will be stored
     *
     * @config
     * @var string
```

src/InterventionBackend.php
```diff
@@ -240,7 +248,9 @@ public function getImageResource()
         try {
             // write the file to a local path so we can extract exif data if it exists.
             // Currently exif data can only be read from file paths and not streams
-            $path = tempnam(TEMP_PATH, 'interventionimage_');
+            $tempPath = $this->config()->get('local_temp_path');
+            $tempPath = $tempPath !== null ? $tempPath : TEMP_PATH;
```
ScopeyNZ: For future reference - we're accepting PHP7 syntax in minor branches now. You could have written `$tempPath = $this->config()->get('local_temp_path') ?? TEMP_PATH;`.

adrhumphreys: I did that originally and then changed it to maintain compat as I thought it'd be required 😂

PR #309, opened 2019-08-08
Allow the temporary store for intervention to be configured

adrhumphreys:
Could we also back port this to v 1.3.3 and onwards if possible (constrained by being stuck on certain prior versions due to recipes)?

This uhhh, is basically this but a bit better... https://github.com/silverstripe/silverstripe-assets/pull/306

Reason being is that on certain envs you won't know the path of the temp dir until you have code running and for perf you might not want to set it each time through _config.php

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -65,6 +65,14 @@ class InterventionBackend implements Image_Backend, Flushable
     const FAILED_UNKNOWN = 'unknown';
 
     /**
+     * Configure where cached intervention files will be stored
+     *
+     * @config
+     * @var string
+     */
+    private static $local_temp_path = TEMP_PATH;
+
+    /**
      * @var AssetContainer
      */
     private $container;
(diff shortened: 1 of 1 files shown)


src/InterventionBackend.php
```diff
@@ -835,4 +844,14 @@ protected function isStreamReadable($stream)
...
     }
+
+    /**
+     * This gets the temp path config usually set through local_temp_path
+     *
+     * @return string
+     */
+    public function getTempPathConfig(): string
+    {
+        return $this->config()->get('local_temp_path') ?? TEMP_PATH;
```
robbieaverill: The fallback doesn't seem strictly necessary here, since the default config value is TEMP_PATH already. I'd also suggest casting the return from config as a string, since it can return null as well (e.g. if you haven't flushed yet but pull this code in)

kinglozzer: Does `private static $local_temp_path = TEMP_PATH;` definitely work as expected? It’s a const we define ourselves in framework, so it strikes me as potentially fragile... I think my preference would be to remove it from the config definition and keep it here

ScopeyNZ: (Not in SS3 but yes in SS4)

adrhumphreys: Would you not want it to throw an exception if config returns null rather than trying to create a file in a random path?

robbieaverill: The problem I'm trying to highlight is that it can be null while you're flushing, if that method is called during the flush process

adrhumphreys: Fair enough, have updated it to cast to a string 👍

robbieaverill: @adrhumphreys could you also address @kinglozzer's comment? You'll need to rebase and fix the conflict as well =)

PR #312, opened 2019-08-14
NEW add extension point to AssetAdaptor for allowing additional extensions

scott1702:
So developers don't need to override the entire PublicAssetAdapter_HTAccess in order to allow access to files that aren't in File's `allowed_extensions` config

### DIFF
--- a/src/Flysystem/AssetAdapter.php
+++ b/src/Flysystem/AssetAdapter.php
@@ -6,6 +6,7 @@ use Exception;
 use League\Flysystem\Adapter\Local;
 use League\Flysystem\Config as FlysystemConfig;
 use SilverStripe\Assets\File;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Assets\Filesystem;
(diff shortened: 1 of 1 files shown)


src/Flysystem/AssetAdapter.php
```diff
@@ -183,6 +185,8 @@ protected function renderTemplate($template)
             }
         }
 
+        $this->extend('updateAllowedExtensions', $allowedExtensions);
```
robbieaverill: I kind of feel like you'd be better to extend `File::getAllowedExtensions()` instead - are you not able to do that because it's static?

ScopeyNZ:
I'm still feeling a little dubious about the use of this new API - considering this is generating an `.htaccess` on flush and the value already comes from config.

scott1702:
@ScopeyNZ I have a workaround for the case that I needed this for, this would have just made it a little bit nicer/cleaner. I'll close this as not necessary

Issue #313, opened 2019-08-15
Early 3.x image thumbnail formats won't be migrated to 4.x


chillu:
3.0 format: `_resampled/<format>-<name>` - see https://github.com/silverstripe/silverstripe-framework/blob/3.0/model/Image.php#L302

3.x format: `_resampled/<format>/<name>` see https://github.com/silverstripe/silverstripe-framework/blob/3/model/Image.php#L760

This means we're not fully migrating thumbnails through `LegacyThumbnailMigrationTask`, effectively causing broken images which also don't redirect any more.

# Sub tasks
* [x] Update legacy file ID parser to understand SS3.0 variant
* [x] Review Ingo's LegacyThumbnailMigrationHelper changes
* [x] Make sure the Tag to short code task works with SS3.0 variant

# Pulls requests
* [x] https://github.com/silverstripe/silverstripe-assets/pull/314
* [x] https://github.com/silverstripe/silverstripe-framework/pull/9202

chillu:
Tracked this down, it changed four years ago in the 3.3.0 release: https://github.com/silverstripe/silverstripe-framework/commit/ea05526e9d1bf3cd38825beee5dbd0a6951b9dbb

chillu:
Here's a comparison of the same file processed by 3.2.0 and 3.x-dev

![image](https://user-images.githubusercontent.com/111025/63127015-70b83280-c005-11e9-923a-c21aad7fcba2.png)

chillu:
- TBC: *PASS* Resize image in code in 3.2, page viewed again in 3.3 or newer: Will recreate thumbnail and update `<img>` reference
- *FAIL* Published page with embedded image in 3.2, never republished in 3.3 or newer: Hardcoded `<img>` reference to old thumbnail, breaks after migration script
- *PASS* Published page with embedded image in 3.2, republished in 3.3 or newer: Creates duplicate thumbnail, updates to new thumbnail format (hardcoded in `<img>` but not an issue since it'll be picked up by the migration script)

Issue #285, opened 2019-06-04
FileMigrationTask doesn't highlight complication when it runs


maxime-rainville:
The File Migration tasks doesn't warn you when it does things out of the ordinary. e.g.: If your SS3 filename is invalid in SS4, it will rename the file without giving you an obvious heads up.

Also when migrating thousands of files, it's not obvious if a dozen of them failed for some reason.

# Acceptance critera
* [ ] As a developer running the file migration task, I can clearly see which files did not get migrated as expected.

maxime-rainville:
For the specific example of renaming an invalid filename, we could just add an extra check that will spit out a log warning instead of a log info, if the filename we get back from `FlysystemAssetStore` is different from what we were expecting.

In term of giving a clear list of what went wrong, we could try adding an extra `LogHandler` that capture any output with "warning" severity or above.

dnsl48:
Ok, I'm confused a little bit. Are we talking about errors or warnings? What do we expect end users to do with this logs?

maxime-rainville:
We're talking about errors and warnings. The issue is you might have thousand of entries in those logs with only a handful of failures. You could easily miss those failures unless you pay attention closely.

dnsl48:
If we got prefixes there that could be a simple matter of `grep` by `ERROR` or `WARN` messages. AFAIK that's how many people read apache and nginx logs for example.

maxime-rainville:
Indeed ... `grep` is the command of the bash Gods. Sadly, many developers still struggle through the dark night, unaware of the exalted righteousness of `grep`. We can only wait until all developers turn towards the magnificent light of `grep` and we can all join together at the brotherhood of the `grep`.

Until such time, we must endeavour to alleviate the torment of those sad souls who have yet to rejoice at the alter of `grep`. 

`grep` sees all and searches all ... filtering out the wicked lines and returning the worthy ones.

dnsl48:
any text editor has search functionality as well :)
also, I believe grep is easier to use than sake and file migration tasks, so I'm not sure that's for gods only ;)

PR #314, opened 2019-08-19
Handle SS3.0 file ID variant

maxime-rainville:
SS3.0 variants had their variant code prefix stored in the file name. That got changed in SS3.3 to have them in folders.

All our migration logic was assuming the SS3.3 kind of variant. This PR:
* Updates LegacyFileIDHelper to understand SS3.0 variant FileID
* Updates LegacyThumbnailMigrationHelper to use FileIDHelpers to parse FileIDs
  * All classes that us FileIDHelpers to understand SS3 variants should magically start working.
* Update unit test to try ss3.0 variant for the following cases:
  * Migrating Legacy thumbnail
  * Migrating SS3 files
  * Redirect old variant URL
  * Convert Tags to short code.

# Note
I didn't actually test any of those cases for "real". It's all unit tests magic.

I'll do an actual test now, but you can start the peer review.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/313

### DIFF
--- a/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
+++ b/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
@@ -143,7 +143,7 @@ class LegacyThumbnailMigrationHelper
         $legacyFileIDParser = new LegacyFileIDHelper($failNewerVariant);
         $naturalFileIDParser = new NaturalFileIDHelper();
 
-
+        $foundError = false;
         // Recurse through folder
         foreach ($filesystem->listContents($resampledFolderPath, true) as $fileInfo) {
             if ($fileInfo['type'] !== 'file') {
@@ -154,10 +154,11 @@ class LegacyThumbnailMigrationHelper
 
             $parsedFileID = $legacyFileIDParser->parseFileID($oldResampledPath);
 
+            // If we can't parse the fileID, let's bail on this file and print out an error
             if (!$parsedFileID) {
-                throw new \LogicException(
-                    'Could not find valid variants in ' . $oldResampledPath
-                );
+                $foundError = true;
+                $this->logger->error('Could not find valid variants in ' . $oldResampledPath);
+                continue;
             }
 
             // Replicate new variant format.
@@ -184,11 +185,11 @@ class LegacyThumbnailMigrationHelper
             $moved[$oldResampledPath] = $newResampledPath;
         }
 
-        // Remove folder and any subfolders.
-        // Assumes all files have been handled in the loop above,
-        // and either deleted (if new location already exists),
-        // or moved out of the folder (if new location didn't exist).
-        $filesystem->deleteDir($resampledFolderPath);
+        // Remove folder and any subfolders. If one or more thumbnail didn't
+        // get migrated leave the folder where it is.
+        if (!$foundError) {
+            $filesystem->deleteDir($resampledFolderPath);
+        }
 
         return $moved;
     }
--- a/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
+++ b/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Assets\Dev\Tasks;
 
+use League\Flysystem\Filesystem;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use SilverStripe\Assets\File;
@@ -132,7 +133,7 @@ class LegacyThumbnailMigrationHelper
         $resampledFolderPath = $folderPath . '_resampled';
 
         // Legacy thumbnails couldn't have been stored in a protected filesystem
-        /** @var \League\Flysystem\Filesystem $filesystem */
+        /** @var Filesystem $filesystem */
         $filesystem = $store->getPublicFilesystem();
 
         if (!$filesystem->has($resampledFolderPath)) {
@@ -188,7 +189,22 @@ class LegacyThumbnailMigrationHelper
         // Remove folder and any subfolders. If one or more thumbnails didn't
         // get migrated leave the folder where it is.
         if (!$foundError) {
-            $filesystem->deleteDir($resampledFolderPath);
+            $files = array_filter(
+                $filesystem->listContents($resampledFolderPath, true),
+                function ($file) {
+                    return $file['type'] === 'file';
+                }
+            );
+            if (empty($files)) {
+                $filesystem->deleteDir($resampledFolderPath);
+            } else {
+                // This should not be possible. If it is, then there's probably a bug.
+                $this->logger->error(sprintf(
+                    'Could not remove folder %s because it still contains files. Please submit a bug report at %s.',
+                    $oldResampledPath,
+                    'https://github.com/silverstripe/silverstripe-assets/issues/new'
+                ));
+            }
         }
 
         return $moved;
--- a/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
+++ b/src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
@@ -5,6 +5,8 @@ namespace SilverStripe\Assets\Dev\Tasks;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use SilverStripe\Assets\File;
+use SilverStripe\Assets\FilenameParsing\LegacyFileIDHelper;
+use SilverStripe\Assets\FilenameParsing\NaturalFileIDHelper;
 use SilverStripe\Assets\Flysystem\FlysystemAssetStore;
 use SilverStripe\Assets\Folder;
 use SilverStripe\Core\Config\Configurable;
@@ -137,6 +139,11 @@ class LegacyThumbnailMigrationHelper
             return $moved;
         }
 
+        $failNewerVariant = false;
+        $legacyFileIDParser = new LegacyFileIDHelper($failNewerVariant);
+        $naturalFileIDParser = new NaturalFileIDHelper();
+
+
         // Recurse through folder
         foreach ($filesystem->listContents($resampledFolderPath, true) as $fileInfo) {
             if ($fileInfo['type'] !== 'file') {
@@ -145,16 +152,11 @@ class LegacyThumbnailMigrationHelper
 
             $oldResampledPath = $fileInfo['path'];
 
-            $variantParts = $this->getVariantParts330($fileInfo);
+            $parsedFileID = $legacyFileIDParser->parseFileID($oldResampledPath);
 
-            // Fall back to <3.3.0 style
-            if(!$variantParts) {
-                $variantParts = $this->getVariantParts300($fileInfo);
-            }
-
-            if (!$variantParts) {
+            if (!$parsedFileID) {
                 throw new \LogicException(
-                    'Could not find valid variants in ' . $fileInfo['path']
+                    'Could not find valid variants in ' . $oldResampledPath
                 );
             }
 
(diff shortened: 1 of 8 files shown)


src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
```diff
@@ -145,9 +152,13 @@ protected function migrateFolder(FlysystemAssetStore $store, Folder $folder)
 
             $oldResampledPath = $fileInfo['path'];
 
-            // Get "variant folders" inside _resampled folder.
-            $variantFolders = preg_replace('#.*_resampled/(.*)#', '$1', $fileInfo['dirname']);
-            $encodedVariants = explode(DIRECTORY_SEPARATOR, $variantFolders);
+            $parsedFileID = $legacyFileIDParser->parseFileID($oldResampledPath);
+
+            if (!$parsedFileID) {
+                throw new \LogicException(
```
maxime-rainville: I thought about replacing this exception with an error log, so the entire job won't crash if it comes across an error. But that would have mean tweaking the directory deletion below.

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -15,6 +17,33 @@
...
 {
     use Injectable;
+    use Configurable;
+
+    /**
+     * List of SilverStripe 3 image method names that can appear in variants. Prior to SilverStripe 3.3, variants were
+     * incoded in the filename with dashes. e.g.: `_resampled/FitW10-sam.jpg` rather than `_resampled/FitW10/sam.jpg`.
+     * @config
+     */
+    private static $ss3_image_variant_methods = [
```
maxime-rainville: By making this configurable, someone who had their own weird variant method can add it to the list.

tests/php/Dev/Tasks/FileMigrationHelperTest.php
```diff
@@ -275,17 +279,23 @@ private function individualFiles(File $file)
...
-        $this->assertFileExists($dir . '/' . $filename);
+        $this->assertFileExists($dir . '/' . $baseFilename);
+
+        // Test that SS3.3 variants have been migrated
+        $variantFilename = preg_replace('#^(.*)\.(.*)$#', '$1__resizeXYZ.$2', $baseFilename);
+        $this->assertFileExists($dir . '/' . $variantFilename);
+        $variantFilename = preg_replace('#^(.*)\.(.*)$#', '$1_scaleABC.$2', $variantFilename);
+        $this->assertFileExists($dir . '/' . $variantFilename);
+
+        // Test that pre SS3.3 variants have been migrated
```
maxime-rainville:
```suggestion
        // Test that pre SS3.0 variants have been migrated
```

maxime-rainville:
@chillu A couple points to note.

I've added some extra unit test around scenarios where you would have a file inside a `_resampled` folder without any actual variant data (e.g.: `resampled/myfile-without-any-variant-prefix.jpg`). This would have been parsed as an empty variant before. Now, `LegacyFileIDHelper::parse()` will return `null` to specify that it can not understand this URL.

When parsing an SS3.3 variant file ID, the PR doesn't make any effort to check that the variant method is valid. The SS3.0 logic needs to be more picky because the convention is more ambiguous. It will do a case insensitive check to see if the stated method is a valid image variant function and it will try to decode the base64 argument to see if it's an array.

sminnee:
I'll look at testing this on the silverstripe.org asset set, where I've been experiencing issues.

src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
sminnee: If people hit this on a large site, you could potentially be adding _days_ to a migration project, as you'd need to run, fix, run, fix a script that takes hours and hours. How likely is it that we'd hit this exception?

maxime-rainville: Any nonsensical file inside a `_resampled` folder would trip this Exception, so I would expect mildly common for big projects that have been going a long time.

However this is not the primary migration tasks. It's meant to clean up any orphan variants after the main migration task is done. In practice, it's an optional task.

Switching to a logger would be a simple process. I just left the note for Ingo to ponder during his peer reviewing.

sminnee:
OK this did _not_ work with my local copy of silverstripe.org so I'll investigate and get back as to why...

src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
chillu: I've flagged a ticket about the wider issue of "defensive coding" during file migration with Bryn, it's part of Exp Debt now: https://github.com/silverstripe/silverstripe-assets/issues/285. So basically: Don't abort task unless there's no other option

sminnee:
The following URLs were not matched

 * assets/2018/logos/_resampled/ResizedImageWzE1MCw1MF0/lenovo-2018.png
 * assets/LogoRoll/_resampled/ResizedImageWzg1LDEwOF0/2D-Logo.png
 * assets/LogoRoll/_resampled/ResizedImageWzkwLDkwXQ/skinny.png
 * assets/LogoRoll/_resampled/ResizedImageWzE0NCwzNl0/Lords-Cricket-Ground-logo.svg.png

In spite of the following file existing:

 * assets/2018/logos/lenovo-2018.png
 * assets/LogoRoll/skinny.png
 * assets/LogoRoll/Lords-Cricket-Ground-logo.svg.png

Note that the resamples were flushed out these ones weren't regenerated by the assets migration. I expect that this is the root cause of the issue but it's an important one to solve.

sminnee:
It's possible that the "real" bug here is that MigrateFileTask doesn't move the ResizedImageWzE0NCwzNl0 to an appropriate place...

maxime-rainville:
> Moving thumbnails happens automatically in 4.4 when running MigrateFileTask. It's marked as a optional task in the release notes, but that's only really the case for devs upgrading from 4.3. We should make this clearer.

As far as I'm concern, migrating Thumbnail is always an optional task, even when upgrading from 3.x. The main value it provides is to move orphan thumbnails to their expected location, even if we've somehow lost their primary file.

The `FileMigrationHelper` should already have moved all the thumbnails that are associated to an existing images to their final SS4 destination.

If `FileMigrationHelper`, somehow didn't manage to move all your thumbnails to their intended destination, then the ThumbnailMigration should allow you to recover them after the fact.

> Because thumbnails are moved, and locations are hardcoded in 3.x style  links, those embeds are now broken unless you also run TagsToShortcode. That's not very obvious from the migration docs.

Out of the box, SS4.4 should be redirecting request for an SS3 variant to it's equivalent SS4 path. e.g.: After running the migration, if you try to access `assets/_resampled/ResizedImageWzE1MCw1MF0/logo.png` or `assets/_resampled/ResizedImageWzE1MCw1MF0-logo.png`, you should automatically be redirected to `assets/logo__ResizedImageWzE1MCw1MF0.png`.

If you try to edit a page that references `assets/_resampled/ResizedImageWzE1MCw1MF0/logo.png`, the WYSIWYG, won't know what the parent image is. This is the main value `TagsToShortcode` provides.

@sminnee Do you have any logs left over from your migration, or a snapshot I could play with? If you run `sake dev/tasks/MigrateFileTask only=move-thumbnails` does finish moving those outstanding variants?

chillu:
Discussed with Max, add this to "file migration" docs and reference it from the 4.4.0 changelog.

> In case your site has existed prior to 3.3, you might be affected by an edge case around thumbnail migrations. This only applies if have run full or incremental migration tasks on SilverStripe 4.4.0 or newer (`dev/tasks/MigrateFileTask` or `dev/tasks/MigrateFileTask only=migrate-thumbnails`). In this case, the migration might have caused orphaned thumbnails. Thumbnails generated through templates or code, these are automatically regenerated. Thumbnails embedded in HTML fields, they are regenerated when the page is republished. In case you want to clean up the orphaned thumbnails, you can find files ending in `__resampled` (e.g. `ScaleWidthWyIxMDAiXQ-image__MyFolder__resampled.jpg`)

maxime-rainville:
I had another look at what the pre-4.4 file migration task did. It completely ignores the variants.

So if you migrate upgrade a site from SS3 to SS4.3 and run the migration task, you'll be left with a bunch of `_resample` folders full of variants. 

If you then upgrade to SS44 and run the file migration task in full, the `move-files` part will not look for those SS3 variants, because the SS3 files will have been marked as migrated. However, the `move-thumbnails` part will find those files and try to move them to their matching position. The SS3.3 thumbnails will land in the correct place. The SS3.0 variants will get mangled names (e.g.: `Uploads/ResizedImageWzIwMCwxMDdd-image-medium__Uploads__resampled.jpg`) because the `LegacyThumbnailMigrationHelper` wasn't code with this kind of variant in mind.

maxime-rainville:
@sminnee Tried replicating your set up, but everything worked fine for me.

maxime-rainville:
I've updated the file migration doc here https://github.com/silverstripe/silverstripe-framework/pull/9202

sminnee:
OK I don't think my concerns relating to ssorg are a blocker to getting this merged. I'll raise a new ticket if I identify any further bugs.

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -102,6 +135,90 @@ public function parseFileID($fileID)
         );
     }
 
+    /**
+     * Try to parse a FileID as a pre-SS33 variant. From SS3.0 to SS3.2 the variants were prefixed in the file name,
+     * rather than ecoded into folders.
```
bergice:
```suggestion
     * rather than encoded into folders.
```

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -102,6 +135,90 @@ public function parseFileID($fileID)
...
+        } else {
+            $pattern = '#^(?<folder>([^/]+/)*?)(_resampled/(?<variant>((((' . $variantPartialRegex . ')[^.-]+))-)+))?((?<basename>([^/.])+))(?<extension>(\..+)*)$#i';
+        }
+
+        // not a valid file (or not a part of the filesystem)
+        if (!preg_match($pattern, $fileID, $matches)) {
+            return null;
+        }
+
+        // Our SS3 variant can be confused with regular filename, let's minise the risk of this by making
```
bergice:
```suggestion
        // Our SS3 variant can be confused with regular filenames, let's minimise the risk of this by making
```

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -102,6 +135,90 @@ public function parseFileID($fileID)
...
+        $possibleVariants = explode('-', $variant);
+        $validVariants = [];
+        $validVariantRegex = '#^(' . $variantPartialRegex . ')(?<base64>(.+))$#i';
+
+        // Loop through the possible variants until we find an invalid one
+        while ($possible = array_shift($possibleVariants)) {
+            // Find the base64 encoded argument attached to the image method
+            if (preg_match($validVariantRegex, $possible, $variantMatches)) {
+                try {
+                    // Our base 64 encoded string always decodes to an string representation of php array
```
bergice:
```suggestion
                    // Our base 64 encoded string always decodes to a string representation of php array
```

src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -102,6 +135,90 @@ public function parseFileID($fileID)
...
+                    // So we're assuming it always start with a `[` and ends with a `]`
```
bergice:
```suggestion
                    // So we're assuming it always starts with a `[` and ends with a `]`
```

src/Dev/Tasks/LegacyThumbnailMigrationHelper.php
```diff
@@ -179,11 +185,11 @@ protected function migrateFolder(FlysystemAssetStore $store, Folder $folder)
...
 
-        // Remove folder and any subfolders.
-        // Assumes all files have been handled in the loop above,
-        // and either deleted (if new location already exists),
-        // or moved out of the folder (if new location didn't exist).
-        $filesystem->deleteDir($resampledFolderPath);
+        // Remove folder and any subfolders. If one or more thumbnail didn't
+        // get migrated leave the folder where it is.
+        if (!$foundError) {
+            $filesystem->deleteDir($resampledFolderPath);
```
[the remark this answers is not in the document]
maxime-rainville: If there's no error, there shouldn't be any files left here. Although, it wouldn't be an horrible idea to check and display and log an error if we find a file.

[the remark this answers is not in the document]
maxime-rainville: I'll add a bit.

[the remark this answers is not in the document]
maxime-rainville: Done.

maxime-rainville:
@bergice I've addressed your last comment and rebased.

Issue #320, opened 2019-08-30
TagsToShortcodeTask breaking content


ichaber:
When running TagsToShortcodeTask on SilverStripe 4.4.0 and assets in version 1.4.3 sometimes the Content is replaced with an invalid markup.

E.g.: The following code in a SilverStripe 3 format
```
<p><a><img class="leftAlone" title="" src="assets/media-releases/img/_resampled/ResizedImageAmNsTLr4AXNQ/20190101-harris-lane-6.jpg" alt="" width="600" height="400"></a></p>
```
will be rewritten to the following markup:
```
<p><a id="2"][image class="leftAlone" src="/assets/44ad6c19ab/image.jpg" width="600" height="400" id="2"]</a></p>
```

# Pull Request
* [ ] https://github.com/silverstripe/silverstripe-assets/pull/321

maxime-rainville:
@bergice I've assigned you for peer review as you wrote the original task and you will have the most context. Let me know if this is problematic in any way.

andreaslindahl:
I'm experiencing this bug when migrating from Silverstripe 3 to 4.5. Has this been fixed?

Cheddam:
@andreaslindahl The [bugfix](https://github.com/silverstripe/silverstripe-assets/commit/37d8923503678e41b277bdfc9537dbaeb8f5ae79) has been included in CMS 4.4.4-4.5.0 so far, so you shouldn't be experiencing issues. Could you provide some more context (which version of CMS 3 you're migrating from, any broken content samples, etc.)?

PR #321, opened 2019-09-02
BUG Update regex matching tags to stop looking after the closing > char

maxime-rainville:
There's a regex that's meant to find the opening `<a>` and `` tags. However, it wouldn't stopping looking after coming across a `>` character.

So when looking at an expression like this `<a></a>`, `<a>` would be match as the opening tag.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/320

### DIFF
--- a/src/Dev/Tasks/TagsToShortcodeHelper.php
+++ b/src/Dev/Tasks/TagsToShortcodeHelper.php
@@ -210,7 +210,8 @@ class TagsToShortcodeHelper
     {
         $resultTags = [];
 
-        preg_match_all('/<('.$this->validTagsPattern.').*?('.$this->validAttributesPattern.')\s*=.*?>/i', $content, $matches, PREG_SET_ORDER);
+        $regex = '/<('.$this->validTagsPattern.')\s[^>]*?('.$this->validAttributesPattern.')\s*=.*?>/i';
+        preg_match_all($regex, $content, $matches, PREG_SET_ORDER);
         if ($matches) {
             foreach ($matches as $match) {
                 $resultTags []= $match[0];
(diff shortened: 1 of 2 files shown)


src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -210,7 +210,8 @@ private function getTagsInContent($content)
     {
         $resultTags = [];
 
-        preg_match_all('/<('.$this->validTagsPattern.').*?('.$this->validAttributesPattern.')\s*=.*?>/i', $content, $matches, PREG_SET_ORDER);
+        $regex = '/<('.$this->validTagsPattern.')\s[^>]*?('.$this->validAttributesPattern.')\s*=.*?>/i';
```
maxime-rainville: This is where the magic happens. Replacing `.*?` for `\s[^>]*?` means that we will stop looking for an `href|src` attribute if our tag doesn't contain a space or if we hit a `>` character.

That way, we won't match the `href` or `src` in a follow up tag.

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -313,20 +314,18 @@ private function getNewTag($tag)
                     '/href\s*=\s*(?:"|\').*?(?:"|\')/i',
                     '/id\s*=\s*(?:"|\').*?(?:"|\')/i',
                     '/\s*(\/?>|\])/',
-                    '/\s?\S+=("")|(\'\')/'
```
maxime-rainville: This line was meant to remove empty attributes. e.g.: `alt=""` would have been stripped from the converted short code.

I don't think this is necessary. Many empty attribute carry meaning ... e.g. a blank `alt` will squash accessibility warning.

@bergice Do you remember if there was any specific use case you had in mind for this?

bergice: @chillu wanted to keep this so that we don't change the structure of the content more than we need to.

I'm okay with removing it though.

chillu: Blank alt tags have semantic meaning (accessibility), so we shouldn't remove them unless it's a lot of work to keep them. But I don't feel strongly about it either way.

PR #325, opened 2019-09-03
BUG Delete the published variant before the publishing the draft version

maxime-rainville:
When we publish a variant, the draft version and all its variants get copied over to the public store, then the existing files get deleted. 

In most cases, this worked fine if your draft version had the same variants as the published file, because the draft files/variants would override the published ones.

However, if your draft store doesn't contain all the variants the live store does, when the main file gets copied over, the AssetStore will start reading the hash of the main file. This is probablematic, because variants of the original file will confused for variants of the newly published file and won't be deleted.

By deleting the published files first, we prevent this weird condition.

# Parent issue
* https://github.com/silverstripe/silverstripe-asset-admin/issues/970

### DIFF
--- a/tests/php/AssetControlExtensionTest.php
+++ b/tests/php/AssetControlExtensionTest.php
@@ -3,7 +3,9 @@
 namespace SilverStripe\Assets\Tests;
 
 use Silverstripe\Assets\Dev\TestAssetStore;
+use SilverStripe\Assets\FilenameParsing\ParsedFileID;
 use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Assets\Storage\DBFile;
 use SilverStripe\Assets\Tests\AssetControlExtensionTest\ArchivedObject;
 use SilverStripe\Assets\Tests\AssetControlExtensionTest\TestObject;
 use SilverStripe\Assets\Tests\AssetControlExtensionTest\VersionedObject;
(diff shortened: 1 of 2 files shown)


tests/php/AssetControlExtensionTest.php
```diff
@@ -222,6 +224,95 @@ public function testReplaceFile()
...
+            'boom'
+        );
+        $this->assertTrue(
+            $store->exists($v1->getFilename(), $v1->getHash(), 'boom'),
+            sprintf('A variant of %s has been created', $v1->getFilename())
+        );
+        
+        // Let's replace the content of the main file and publish it
+        $download->setFromString($v2Content, $v2->getFilename());
+        $object1->write();
```
maxime-rainville: I tried not having the `write()` there and just use the publish, but that created some weird results.

I'll investigate this a bit further.

blueo:
tried this out and it indeed fixes  https://github.com/silverstripe/silverstripe-asset-admin/issues/970 :)

tests/php/AssetControlExtensionTest.php
maxime-rainville: Everywhere, we play with files, we always write them first before publishing them. It seems this has been a constant limitation we've been working around.

It probably would be more intuitive to allow people to publish right away, but that's probably a different bug for another day.

I'll raise an issue for it.

PR #333, opened 2019-09-18
FIX: Chained image manipulations had unexpected behaviour (fixes #332)

### LINKED ISSUE #332: Creating multiple variants from a single variant image is not working properly
Let's start with the use case in order to explain this issue. We do have a module for creating different sized variants of an image for use in a srcset. So typically we do have one call to create an appropriate base image with a certain ratio (i.e. using FillMax) and then create multiple version with different widths (using ScaleMaxWidth). Unfortunately only the first scaled version will be properly sized (respecting the already cropped image from FillMax), the others will have a different ratio. Some example code:

```
use SilverStripe\Assets\Image;

/** @var Image $image given some image, lets assume 1200x800 (ratio 1.5) */
$image;

/** @var Image $filledImage */
$filledImage = $image->FillMax(800, 800);
var_dump([
  $filledImage->getWidth(), 
  $filledImage->getHeight(), 
  $filledImage->getWidth() / $filledImage->getHeight()
]);
// this will be a 800 by 800 (ratio 1) image 
// (given that the original file is larger than this)

$scaledFilledImage = $filledImage->ScaleMaxWidth(500);
var_dump([
  $scaledFilledImage->getWidth(), 
  $scaledFilledImage->getHeight(), 
  $scaledFilledImage->getWidth() / $scaledFilledImage->getHeight()
]);
// this will work properly and create 500 by 500 (ratio 1) image

$scaledFilledImage = $filledImage->ScaleMaxWidth(300);
var_dump([
  $scaledFilledImage->getWidth(), 
  $scaledFilledImage->getHeight(), 
  $scaledFilledImage->getWidth() / $scaledFilledImage->getHeight()
]);
// this will fail and create a 300 by 200 (ration 1.5) image

$scaledFilledImage = $filledImage->ScaleMaxWidth(100);
var_dump([
  $scaledFilledImage->getWidth(), 
  $scaledFilledImage->getHeight(), 
  $scaledFilledImage->getWidth() / $scaledFilledImage->getHeight()
]);
// this will fail as well and create a 100 by 67 (ration ~1.49) image
```

The issue seems to vanish when removing the optimization in [ImageManipulation#manipulate](https://github.com/silverstripe/silverstripe-assets/blob/1/src/ImageManipulation.php#L981).

lozcalver:
Just to explain the patch, given `$Image.Quality(10).Fill(500, 500)`:

Current code:
1. Image resource is loaded
2. Image is resampled to low-quality and written to disk
3. Image resource is passed down to the resampled `DBFile` that’s returned
4. `Fill(500, 500)` is then called on the existing backend + resource, which still has the quality setting stored

The problem with the current code is that in the event that the image resource is garbage collected before another manipulation takes place with the same `Image` record, instead of re-loading the image resource from the resampled copy to work with it’ll load it from the original image. That happens because the `Image_Backend` instance that is currently passed down to the resampled copy has the original `DBFile` tuple stored.

Another problem with the current code that this change fixes is the following example:

```php
// Works as expected
$image->Quality(1)->Fill(200, 200);
// Loads the “Quality1” resampled image created above from the disk, but then saves the
// 500x500 version out with a quality setting of 100, resulting in a large file size
$image->Quality(1)->Fill(500, 500);
```

## Parent issue
- #332

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -947,7 +947,7 @@ trait ImageManipulation
                 $tuple = $result;
                 Deprecation::notice(
                     '5.0',
-                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array 
+                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array
                         containing a tuple and an image backend, i.e. [$tuple, $result]',
                     Deprecation::SCOPE_GLOBAL
                 );
@@ -971,10 +971,11 @@ trait ImageManipulation
         /** @var DBFile $file */
         $file = DBField::create_field('DBFile', $tuple);
 
-        // Pass the manipulated image backend down to the resampled image - this allows chained manipulations
-        // without having to re-load the image resource from the manipulated file written to disk
-        if ($manipulationResult instanceof Image_Backend) {
-            $file->setImageBackend($manipulationResult);
+        // Preserve the current image quality setting, so chained manipulations like
+        // $Quality(1).SetWidth(500) work as expected
+        $currentBackend = $this->getImageBackend();
+        if (method_exists($currentBackend, 'getQuality')) {
+            $file->getImageBackend()->setQuality($currentBackend->getQuality());
         }
 
         return $file->setOriginal($this);
--- a/tests/php/ImageTest.php
+++ b/tests/php/ImageTest.php
@@ -114,16 +114,30 @@ abstract class ImageTest extends SapphireTest
 
         $imageFirst = $image->ScaleWidth(200);
         $this->assertNotNull($imageFirst);
-        $expected = 200;
-        $actual = $imageFirst->getWidth();
-
-        $this->assertEquals($expected, $actual);
+        $this->assertEquals(200, $imageFirst->getWidth());
 
         $imageSecond = $imageFirst->ScaleHeight(100);
         $this->assertNotNull($imageSecond);
-        $expected = 100;
-        $actual = $imageSecond->getHeight();
-        $this->assertEquals($expected, $actual);
+        $this->assertEquals(100, $imageSecond->getHeight());
+
+        // Manipulate a pre-manipulated image twice...
+        // ... this tests that one manipulation on a pre-manipulated image
+        // doesn't affect subsequent manipulations from the same source image
+        $image = $this->objFromFixture(Image::class, 'imageWithoutTitle');
+        $squareImage = $image->Fill(500, 500);
+        $this->assertNotNull($squareImage);
+        $this->assertEquals(500, $squareImage->getWidth());
+        $this->assertEquals(500, $squareImage->getHeight());
+
+        $smallSquareImage = $squareImage->ScaleMaxWidth(200);
+        $this->assertNotNull($smallSquareImage);
+        $this->assertEquals(200, $smallSquareImage->getWidth());
+        $this->assertEquals(200, $smallSquareImage->getHeight());
+
+        $extraSmallSquareImage = $squareImage->ScaleMaxWidth(50);
+        $this->assertNotNull($extraSmallSquareImage);
+        $this->assertEquals(50, $extraSmallSquareImage->getWidth());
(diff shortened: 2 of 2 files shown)


kinglozzer:
I have no idea wtf is going on with Travis here...

kinglozzer:
Okay so the build failures are pre-existing against the 1.3 branch (either from a security fix or merge-up I guess)

src/ImageManipulation.php
```diff
@@ -971,10 +971,11 @@ public function manipulate($variant, $callback)
...
         $file = DBField::create_field('DBFile', $tuple);
 
-        // Pass the manipulated image backend down to the resampled image - this allows chained manipulations
-        // without having to re-load the image resource from the manipulated file written to disk
-        if ($manipulationResult instanceof Image_Backend) {
-            $file->setImageBackend($manipulationResult);
+        // Preserve the current image quality setting, so chained manipulations like
+        // $Quality(1).SetWidth(500) work as expected
+        $currentBackend = $this->getImageBackend();
+        if (method_exists($currentBackend, 'getQuality')) {
```
kinglozzer: I’d add this to the `Image_Backend` interface, but muh BC...

kinglozzer:
Tests are passing after a rebase, I have no idea why codecov is complaining 😆

src/ImageManipulation.php
```diff
@@ -971,10 +971,11 @@ public function manipulate($variant, $callback)
         /** @var DBFile $file */
         $file = DBField::create_field('DBFile', $tuple);
 
-        // Pass the manipulated image backend down to the resampled image - this allows chained manipulations
-        // without having to re-load the image resource from the manipulated file written to disk
-        if ($manipulationResult instanceof Image_Backend) {
-            $file->setImageBackend($manipulationResult);
```
emteknetnz, line 977: shouldn't we keep this line?

tests/php/ImageTest.php
```diff
@@ -114,16 +114,30 @@ public function testMultipleGenerateManipulationCalls()
...
+
+        $smallSquareImage = $squareImage->ScaleMaxWidth(200);
+        $this->assertNotNull($smallSquareImage);
+        $this->assertEquals(200, $smallSquareImage->getWidth());
+        $this->assertEquals(200, $smallSquareImage->getHeight());
+
+        $extraSmallSquareImage = $squareImage->ScaleMaxWidth(50);
+        $this->assertNotNull($extraSmallSquareImage);
+        $this->assertEquals(50, $extraSmallSquareImage->getWidth());
+        $this->assertEquals(50, $extraSmallSquareImage->getHeight());
```
emteknetnz, line 140: shouldn't we be unit testing image quality since that's what the code update fixes?

emteknetnz (changes requested):
Couple of questions

GuySartorelli:
@kinglozzer It's been a wee while since this was opened but I think it's worth taking a look at - do you have some bandwidth to respond to the questions above?

kinglozzer:
Hmm, I can’t remember this anywhere near well enough to answer the questions 😅 I think in the first instance it’s worth someone confirming that the bug(s) still exist, and if the same patch applied to 1.11 fixes them

tests/php/ImageTest.php
GuySartorelli: Having looked into this a bit I can now say: This doesn't _explicitly_ fix images quality, that was just one example of chained image manipulations. This is fixing _all_ chained image manipulations. It would be good to have tests for a few different manipulation chain examples though, so it might be good to add the example from the PR description as a test.

GuySartorelli:
I can confirm that #332 is still a problem with the latest code - and this PR doesn't seem to fix it.
I don't know how to check the quality of the images so I can't check if the specific situation described in this PR's description is a problem, nor if this PR fixes it.

GuySartorelli:
I'm going to close this on the basis that it doesn't fix the issue it is intended to fix, as per my last comment, and it hasn't been touched in a long long time.

Issue #336, opened 2019-09-24
FileHash not invalidated accross multiple servers


maxime-rainville:
# Description
If you're operating in an environment where you have multiple servers, it's possible that he FileHash cache won't be invalidated when you replace a file if that cache is not shared accross all servers.

This leads to weird intermittent behaviour, where files occasionally appear has missing when you hit a server with a stale file hash.

# Possible fix

A simple fix could be to incorporate the file time stamp in the hash key which can be read from the file system and should be touch whenever the physical file is updated. This could have a small performance hit, but retrieving a file timestamp should be a relative quick operation in most situation.

# Pull requests
* [ ]  https://github.com/silverstripe/silverstripe-assets/pull/337

adrhumphreys:
I think the PR should be accepted and then a better solution should be added at a later date. Users whom are upgrading assets are doing so to use the new hashless feature with the idea that they want to retain URL's for assets. This breaks this.

Sure the case can be made that this only occurs on a multi-server setup but that as far as I know is a supported set up for SilverStripe which this is breaking. Either that or I'd expect a warning of some sort to be added to advise users

maxime-rainville:
I don't think this is a very difficult issue to fix. It's probably worth taking an extra day or two to think it thoroughly and come up with long term fix.

NightJar:
Nice work team @adrhumphreys @maxime-rainville :)

PR #337, opened 2019-09-24
Use a timestamp in the hash cache

adrhumphreys:
Caching between two servers is not working as expected. This address that.

Checks to see if File exists before proceeding
https://github.com/silverstripe/silverstripe-assets/blob/1/src/Storage/DBFile.php#L365 
Uses "Flysystem" to check for file existing
https://github.com/silverstripe/silverstripe-assets/blob/master/src/Flysystem/FlysystemAssetStore.php#L1231 
Compares "Hashes" of what is in the database, and what is returned by Flysystem (If they do not match, which they dont, it skips)
https://github.com/silverstripe/silverstripe-assets/blob/master/src/Flysystem/FlysystemAssetStore.php#L327 
Flysystem checks if Cached value of Hash exists (which it does) and returns the cached value instead of checking directly.
https://github.com/silverstripe/silverstripe-assets/blob/master/src/Storage/Sha1FileHashingService.php#L116

This is the point in which a different Hash value is returned by Cache on the offending server (as the cache has not been invalidated / cleared on the server which did not receive the upload).

Cache is non-memory based and is set to infinite lifecycle (meaning it won't expire unless it is specifically removed).
https://github.com/silverstripe/silverstripe-assets/blob/master/_config/assetscache.yml#L17 

By default, all SilverStripe caches are stored in the TempDir (/tmp/silverstripe...). This directory is not shared between servers, so each individual web server contains its own cache values

# Parent issue
* #336

### DIFF
--- a/src/Storage/Sha1FileHashingService.php
+++ b/src/Storage/Sha1FileHashingService.php
@@ -45,7 +45,7 @@ class Sha1FileHashingService implements FileHashingService, Flushable
 
     /** @var Filesystem[] */
     private $filesystems;
-    
+
     public function computeFromStream($stream)
     {
         Util::rewindStream($stream);
@@ -120,7 +120,7 @@ class Sha1FileHashingService implements FileHashingService, Flushable
         $fs = $this->getFilesystem($fs);
         $stream = $fs->readStream($fileID);
         $hash = $this->computeFromStream($stream);
-        
+
         $this->set($fileID, $fs, $hash);
 
         return $hash;
@@ -178,9 +178,12 @@ class Sha1FileHashingService implements FileHashingService, Flushable
      */
     private function buildCacheKey($fileID, $fs)
     {
-        $fsID = $this->getFilesystemKey($fs);
+        $filesystem = $this->getFilesystem($fs);
 
-        return base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        $fsID = $this->getFilesystemKey($fs);
+        $timestamp = $filesystem->getTimestamp($fileID);
+
+        return base64_encode(sprintf('%s://%s://%s', $fsID, $fileID, $timestamp));
     }
 
     public function invalidate($fileID, $fs)
--- a/src/Storage/Sha1FileHashingService.php
+++ b/src/Storage/Sha1FileHashingService.php
@@ -181,7 +181,10 @@ class Sha1FileHashingService implements FileHashingService, Flushable
         $filesystem = $this->getFilesystem($fs);
 
         $fsID = $this->getFilesystemKey($fs);
-        $timestamp = $filesystem->getTimestamp($fileID);
+
+        $timestamp = $filesystem->has($fileID) ?
+            $filesystem->getTimestamp($fileID) :
+            null;
 
         return base64_encode(sprintf('%s://%s://%s', $fsID, $fileID, $timestamp));
     }
--- a/src/Storage/Sha1FileHashingService.php
+++ b/src/Storage/Sha1FileHashingService.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\Assets\Storage;
 
-use Exception;
 use InvalidArgumentException;
 use League\Flysystem\Filesystem;
 use League\Flysystem\Util;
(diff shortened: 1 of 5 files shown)


src/Storage/Sha1FileHashingService.php
```diff
@@ -178,9 +178,21 @@ private function getCache()
...
         $fsID = $this->getFilesystemKey($fs);
 
-        return base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        // If we can't get a timestamp for the file then we'll use the current timestamp so it
+        // invalidates quickly
+        try {
+            $timestamp = $filesystem->has($fileID) ?
+                $filesystem->getTimestamp($fileID) :
+                DBDatetime::now()->getTimestamp();
+        } catch (Exception $e) {
```
maxime-rainville: What `Exception` are we actually expecting here? Using a catch all like `Exception` could lead to legitimate errors being missed, because they will get caught all the time.

maxime-rainville: What `Exception` are we actually expecting here? Using a catch all like `Exception` could lead to legitimate errors being missed, because they will get caught all the time.

src/Storage/Sha1FileHashingService.php
```diff
@@ -178,9 +178,21 @@ private function getCache()
...
+            $timestamp = DBDatetime::now()->getTimestamp();
```
maxime-rainville: Did you check that `DBDatetime::now()->getTimestamp()` and `$filesystem->getTimestamp($fileID)` use exactly the same format?

chillu:
> You don't have a way to invalidate old cache entries using timestamp for files that don't exist any more.

That depends on how you invalidate those caches on a system level. If "fresh" deployments nuke the cache, you're relying on regular deployments before disk space runs out. Or you have a cron job that nukes a cache folder. In both cases, it won't be limited to older caches only, and this makes it infeasible for large or performance sensitive contexts. Specialised cache adapters such as Memcache or Redis have garbage collection built in, it's mostly an issue with the default filesystem caches (constrained by disk space), and the PHP opcache (somewhat constrained by PHP memory).

There's an open ticket on how to handle cache cleanup in 4.x: https://github.com/silverstripe/silverstripe-framework/issues/6678. Even if we implemented this in framework (by calling `prune()` via a cronjob or randomly in process), it won't fix the garbage collection issues with non-expiring cache entries.

src/Storage/Sha1FileHashingService.php
```diff
@@ -179,10 +178,23 @@ private function getCache()
...
+     * @param string $fileID
+     * @param string|Filesystem $fs
+     * @return string
+     */
+    private function getTimestamp($fileID, $fs)
+    {
+        $filesystem = $this->getFilesystem($fs);
+        return $filesystem->has($fileID) ?
+            $filesystem->getTimestamp($fileID) :
+            DBDatetime::now()->getTimestamp();
```
maxime-rainville: They are instances where we'll be generating a hash value for a file that doesn't exists yet, so we need a fallback in case the file doesn't exists.

src/Storage/Sha1FileHashingService.php
maxime-rainville: Had a chat with Adrian, the problem the Mock Filesystem in some test didn't know how to build a timestamp, which was throwing an exception.

src/Storage/Sha1FileHashingService.php
maxime-rainville: It does.

maxime-rainville:
I tweaked the initial implementation, to store the timestamp on that value, instead of the key. This means we'll only ever have one possible entry for each fileID/Filesystem combination.

Before returning the hash, we compare the timestamp in the cache to the one of the physical file. If they don't match, we assume that the hash has been invalidated and recompute it.

Those timestamp are accurate to the second. There's a possible edge case where you would have an outside process that update the file within the same second as SilverStripe is trying to read it's hash. I'm not sure how we could address that however.

src/Storage/Sha1FileHashingService.php
maxime-rainville: Just had a chat with @NightJar about the virtue of return a fake timestamp. We were originally doing this, because we were making it part of the key and we sometime needed to generate hash keys for files that might not exists.

After some consideration, there's a scenario where this is useful. If you're moving a file around, the FysystemAssetStore will try to move the existing hash value to a different hash key. If you decide to move the hash before the physical file, you won't have a file there yet. By generating, a timestamp equal to "now", FlysystemAssetStore won't have to worry about this intricacy. So we reduce complexity a little bit.

Issue #129, opened 2018-04-08
Image's "Title" should not go into title attribute


JorisDebonnet:
When adding an Image into a HTMLEditorField, the dialog will by default put the Image's Title into the "Title text (tooltip)". It should instead go into the "Alternative text (alt)" attribute, as this is what that attribute is meant for (if the file has a descriptive name, or the user has updated the Title into something sensible). `<img>` elements are actually even illegal without an `alt` attribute, so it should at least be set as an empty string. While the `title` attribute is an optional attribute for _any_ html element, so it should not be automatically set.

I strongly suggest moving the auto-fill from `title` to `alt`, or simply not copying the Title to _any_ attribute.

(slightly edited on 12 Nov 2019 after discussion)

maxime-rainville:
Hi @JorisDebonnet 

I just had a look in SS4.1. From what I can see, if you leave the `Alternative text` field empty in the WYSIWYG, an `alt` attribute will be added to the HTML output when your WYSIWYG content is rendered in the front end. The `alt` attribute is populated with the `Title` attached to the image object.

Can you let me know if that's what you are seeing on your end?

However, it is not super clear this is happening from the CMS editor's point of view. Also, if you try to edit an image you have already inserted in the WYSIWYG, the placement information that was provided originally isn't pre-populated into the _Placement_ tab. We should take a look this.

JorisDebonnet:
You're right, the alt field is populated when the page is actually parsed. But it is indeed also a serious issue that the Image dialog does not 'remember' any properties that were filled in before - if I manually resize (drag-and-drop) an image to just the right size, but then open the dialog to change the alignment, it will reset the size (along with everything else). Perhaps an "edit image" dialog is not really implemented yet?

dhensby:
@JorisDebonnet those edit image issues sound like a bug; care to open a new issue?

maxime-rainville:
@dhensby @JorisDebonnet I've created a separate issue (#131) for loosing the placement.

JorisDebonnet:
I could give this one (`title` vs `alt`) a go, but haven't worked on any WYSIWYG integrated stuff yet. Which file would I start in? (not even sure it's in this repository o.O)

mllngtn:
Hi, I've just run into this issue today.

I think that the most standards-compliant thing to do would be to add an empty 'alt' attribute (alt="") by default, if the alt-text field is left empty.

---

[...]

PR #338, opened 2019-09-24
FIX Force alt attribute to render, remove fallback to image Title

Cheddam:
Some images are presentational, and therefore shouldn't have alt content. Others have captions, and therefore the alt content would be duplicate. Most importantly, the name of a file often doesn't accurately or sufficiently describe its contents, and so it is unfit as alt content.

To improve a11y, the alt attribute should always be rendered, even if it's empty, and it should not have a default value.

Part of [asset-admin#983](https://github.com/silverstripe/silverstripe-asset-admin/issues/983).

Also resolves https://github.com/silverstripe/silverstripe-assets/issues/129

### DIFF
--- a/src/Shortcodes/ImageShortcodeProvider.php
+++ b/src/Shortcodes/ImageShortcodeProvider.php
@@ -84,18 +84,18 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
 
         // Build the HTML tag
         $attrs = array_merge(
-            // Set overrideable defaults
-            ['src' => '', 'alt' => $record->Title],
+            // Set overrideable defaults ('alt' must be present regardless of contents)
+            ['src' => '', 'alt' => ''],
             // Use all other shortcode arguments
             $args,
             // But enforce some values
             ['id' => '', 'src' => $src]
         );
 
-        // Clean out any empty attributes
-        $attrs = array_filter($attrs, function ($v) {
-            return (bool)$v;
-        });
+        // Clean out any empty attributes (aside from alt)
+        $attrs = array_filter($attrs, function ($k, $v) {
+            return (bool)$v || $k === 'alt';
+        }, ARRAY_FILTER_USE_BOTH);
 
         $markup = HTML::createTag('img', $attrs);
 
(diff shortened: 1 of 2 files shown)


src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -84,18 +84,18 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
...
             ['id' => '', 'src' => $src]
         );
 
-        // Clean out any empty attributes
-        $attrs = array_filter($attrs, function ($v) {
-            return (bool)$v;
-        });
+        // Clean out any empty attributes (aside from alt)
+        $attrs = array_filter($attrs, function ($k, $v) {
+            return (bool)$v || $k === 'alt';
```
dnsl48: should we be checking `strlen(trim($v))` rather than `(bool)$v` in here? It feels we don't really want php semantics in here.

Cheddam: `!empty($v)` more accurately tracks the existing behaviour, so I might use that instead? Though your suggestion seems a bit more specific and sensible in this context.

dnsl48: though I reckon the current behaviour is incorrect. E.g. `"0"` is a valid value for an html attribute

Cheddam: Your change wouldn't stop `"0"` from passing this filter, though - and I'd say that's more the work of the HTML parser?

dnsl48:
```php
php > $v = "0";
php > echo (bool)$v;  // == false
php > echo !empty($v);  // == false
php > echo strlen(trim($v));  // == true
```

dnsl48: yeah, my point is we shouldn't be deleting `"0"` from the array

Cheddam: Gotcha - I'll tweak it.

jakxnz:
Should this have also covered https://github.com/silverstripe/silverstripe-assets/blob/1/templates/DBFile_image.ss?

Issue #342, opened 2019-10-07
Can't create folders without Versioned module


p0lar-bear:
This is related to silverstripe/silverstripe-framework#7068

If one disables the `SilverStripe\Versioned\Versioned` extension on the `File` class as described in the 4.0.0 changelog, one cannot create new `Folder` DataObjects.

```
ERROR [Emergency]: Uncaught BadMethodCallException: Object->__call(): the method 'copyVersionToStage' does not exist on 'SilverStripe\Assets\Folder'
IN POST /scpupdate/admin/assets/folderCreateForm/0/
Line 54 in C:\xampp\htdocs\silverstripe\vendor\silverstripe\framework\src\Core\CustomMethods.php

Source
======
  45:       * @throws BadMethodCallException
  46:       */
  47:      public function __call($method, $arguments)
  48:      {
  49:          // If the method cache was cleared by an an Object::add_extension() /
       Object::remove_extension()
  50:          // call, then we should rebuild it.
  51:          $class = static::class;
  52:          $config = $this->getExtraMethodConfig($method);
  53:          if (empty($config)) {
* 54:              throw new BadMethodCallException(
  55:                  "Object->__call(): the method '$method' does not exist on '$class'"
  56:              );
  57:          }
  58:  
  59:          switch (true) {
  60:              case isset($config['callback']): {

Trace
=====
SilverStripe\View\ViewableData->__call(copyVersionToStage, Array)
Folder.php:241

SilverStripe\Assets\Folder->onAfterWrite()
DataObject.php:1548

SilverStripe\ORM\DataObject->write()
AssetAdmin.php:981

SilverStripe\AssetAdmin\Controller\AssetAdmin->createfolder(Array, SilverStripe\Forms\Form, SilverStripe\Control\HTTPRequest, SilverStripe\Admin\LeftAndMainFormRequestHandler)
FormRequestHandler.php:230
```

Looking around line 241 of `Folder`, I see this:

```
    public function onAfterWrite()
    {
        parent::onAfterWrite();

        // No publishing UX for folders, so just cascade changes live
        if (class_exists(Versioned::class) && Versioned::get_stage() === Versioned::DRAFT) {
            $this->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
        }

        // Update draft version of all child records
        $this->updateChildFilesystem();
    }
```

[...]

PR #343, opened 2019-10-07
Check if Versioned extension is applied before using it for folders

michalkleiner:
Fix for #342.

I was thinking whether it should rather use `static::has_extension` for the edge cases where someone keeps Versioned for File but removes it for Folder.

### DIFF
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -237,7 +237,7 @@ class Folder extends File
         parent::onAfterWrite();
 
         // No publishing UX for folders, so just cascade changes live
-        if (class_exists(Versioned::class) && Versioned::get_stage() === Versioned::DRAFT) {
+        if (class_exists(Versioned::class) && $this->hasExtension(Versioned::class) && Versioned::get_stage() === Versioned::DRAFT) {
             $this->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
         }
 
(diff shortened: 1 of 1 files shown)


michalkleiner:
Can squash if needed.

michalkleiner:
Can you please have a look, @dnsl48 or @maxime-rainville? Thanks 👍

src/Folder.php
```diff
@@ -237,7 +237,8 @@ public function onAfterWrite()
         parent::onAfterWrite();
 
         // No publishing UX for folders, so just cascade changes live
-        if (class_exists(Versioned::class) && Versioned::get_stage() === Versioned::DRAFT) {
+        if (class_exists(Versioned::class) && File::has_extension(Versioned::class) && Versioned::get_stage() === Versioned::DRAFT)
```
dnsl48: Should we rather use `$this->hasExtension`?

dnsl48: right, that's not really gonna work here since it has different semantics. I guess static would be more flexible then?

dnsl48: though, maybe that's what we actually need here? @maxime-rainville what do you think?

maxime-rainville: I think it makes some degree of sense to specifically look at whether Files has extensions applied to it rather than Folder. While Folder extends File, there's bits of logic that tries to suppress their Versioned behaviour. In a perfect world Folder wouldn't extends File.

dnsl48:
I've pushed a patch for this and removed redundant newlines.
I reckon it's good to merge after a squash if the build is green.

dnsl48:
## merge checklist

* [x] the target branch is [correct](https://docs.silverstripe.org/en/4/contributing/code/#picking-the-right-version)
* [x] all commits are relevant (e.g. no debugging and linting)
* [x] the commit messages follow [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages)
* [x] the patch follows [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/)
* [x] new features are covered with tests (back-end with unit tests, front-end with Behat)
* [x] the relevant CMS documentation is updated; changelog for impactful changes
* [x] CI is green
* [x] at least one peer reviewer approved; no changes requested

PR #344, opened 2019-10-08
FIX Do not show FileLink's in campaigns

bergice:
Relates to https://github.com/silverstripe/silverstripe-asset-admin/issues/903


emteknetnz:
Looks fine, makes use of this functionality https://github.com/silverstripe/silverstripe-campaign-admin/blob/1.6/src/CampaignAdmin.php#L337

Should retarget this to 1.6

PR #346, opened 2019-10-16
BUGFIX handle files with no extension as if they are invalid

andrewandante:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/9037

Struggling to get the tests to run locally but I've added one in, hopefully Travis can determine that this works.

Aims to gracefully skip over a file if it has no extension rather than implode the task.

### DIFF
--- a/src/Dev/Tasks/FileMigrationHelper.php
+++ b/src/Dev/Tasks/FileMigrationHelper.php
@@ -96,7 +96,7 @@ class FileMigrationHelper
         Environment::increaseTimeLimitTo();
         Environment::setMemoryLimitMax(-1);
         Environment::increaseMemoryLimitTo(-1);
-        
+
         /** @var FileHashingService $hasher */
         $hasher = Injector::inst()->get(FileHashingService::class);
         $hasher::flush();
@@ -319,7 +319,7 @@ class FileMigrationHelper
         } catch (ValidationException $e) {
             if ($this->logger) {
                 $this->logger->error(sprintf(
-                    "File %s could not be migrated due to an error. 
+                    "File %s could not be migrated due to an error.
                     This problem likely existed before the migration began. Error: %s",
                     $legacyFilename,
                     $e->getMessage()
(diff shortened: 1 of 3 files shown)


src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -257,8 +257,14 @@ protected function migrateFile($base, File $file, $legacyFilename)
             return false;
         }
 
-        // Fix file classname if it has a classname that's incompatible with its extention
+        // Ignore extensionless files
         $extension = $file->getExtension();
+        if ($extension === "") {
```
michalkleiner: Possibly use apostrophes instead of quotes for consistency?

andrewandante: I only used doubles as the phpdoc for `File::get_file_extension()` shows:

```
 * Examples:
 * - "myfile" returns ""
```

So not too bothered, I guess we aren't interpolating a string so doubles aren't required

src/Dev/Tasks/FileMigrationHelper.php
```diff
@@ -569,4 +561,30 @@ protected function getFilenameArray()
...
     }
+
+    /**
+     * @param File $file
+     * @param array $messages
+     * @param string $legacyFilename
+     *
+     * return void
+     */
+    protected function handleInvalidFile($file, $messages, $legacyFilename)
```
emteknetnz, line 572: does this need to be a protected? could we make it private instead and changes this to a patch release? (optional for patch release, would require a rebase/retarget to `1.6`)

emteknetnz (changes requested):
Looks great - one thing re the protected method

Happy to merge after that

src/Dev/Tasks/FileMigrationHelper.php
andrewandante: Yeah making it private should be no problem. Happy to point it at 1.6, will have a go this afternoon.

PR #347, opened 2019-11-06
disable subsite filter, ensuring all records are affected by task

scott-nz:
Resolves Issue where TagsToShortcode dev task does not update tags on records that are associated with a subsite.


scott-nz:
@maxime-rainville Here is the PR for the issue we worked on yesterday

maxime-rainville:
Humm ... didn't think about that. The best place to put the hook would probably be in `TagsToShortcodeTask` ... then we would need to do a paralele PR in SubSite to connect into the hook.

emteknetnz:
@scott-nz @maxime-rainville 

Hiya, I guess this one won't be merged as is so perhaps we should close this now?

Are we likely to raise an alternative PR for this one to use extension hooks as well as a parallel extension method on subsites?

PR #348, opened 2019-11-14
BUG Fix linting issue in VersionedFilesMigrationTask and VersionedFilesMigrator

maxime-rainville:
Just fixing some linting issue and hopefully making the build green

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -524,9 +524,12 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
             return true;
         }
         if ($member = Security::getCurrentUser()) {
-            $file = File::get()->byID($fileID);
-            if ($file) {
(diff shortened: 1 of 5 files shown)


maxime-rainville:
I added an extra fix that should resolve the PostgreSQL issue. The issue also affect MySQL only it fails silently.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -524,9 +524,12 @@ protected function isGranted($fileID)
             return true;
         }
         if ($member = Security::getCurrentUser()) {
-            $file = File::get()->byID($fileID);
```
maxime-rainville: This will never return a file because $fileID will be a string like "foldername/xxxxxxxxxxx/filename.txt". `byID` expect a numeric ID.

PostgreSQL throws an exception here, because it doesn't like that we're trying to do a where statement against a numeric column with a string value. But this is equally problematic for a MySQL site.

PR #351, opened 2019-11-19
Intervention files get left on the server and neglected

fonsekaean:
When images are being manipulated it creates a temp file from the stream in the `silverstripe-cache` folder. these images doent get deleted and stays in the cache folders. I've noticed this when using every version of the assets module.


kinglozzer:
These are supposed to be cleaned up in `__destruct()`, but I guess that doesn’t happen in the event of a fatal error during resampling...

I _think_ this should be okay, as I’m pretty sure `intervention/image` loads the image into memory when you instantiate it

bergice:
Resolved with https://github.com/silverstripe/silverstripe-assets/pull/366

PR #356, opened 2019-11-28
FIX: Only allow numeric dimensions in ImageShortcodeProvider (fixes #355)

### LINKED ISSUE #355: Image shortcode with percentage width causes a server error
Cause: users entering `100%` for an image width.

This is helped by the improvements to image insertion in `asset-admin` that will prevent entering non-numeric values in 4.5+, but could be fairly easily solved by adding an `is_numeric()` check in the shortcode provider. I think we should fix it here too, in case someone edits the HTML directly in TinyMCE

kinglozzer:
Fixes #355

### DIFF
--- a/src/Shortcodes/ImageShortcodeProvider.php
+++ b/src/Shortcodes/ImageShortcodeProvider.php
@@ -69,8 +69,8 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
         // Check if a resize is required
         $src = $record->Link();
         if ($record instanceof Image) {
-            $width = isset($args['width']) ? $args['width'] : null;
-            $height = isset($args['height']) ? $args['height'] : null;
+            $width = (isset($args['width']) && is_numeric($args['width'])) ? $args['width'] : null;
+            $height = (isset($args['height']) && is_numeric($args['height'])) ? $args['height'] : null;
             $hasCustomDimensions = ($width && $height);
             if ($hasCustomDimensions && (($width != $record->getWidth()) || ($height != $record->getHeight()))) {
                 $resized = $record->ResizedImage($width, $height);


src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -69,8 +69,8 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
         // Check if a resize is required
         $src = $record->Link();
         if ($record instanceof Image) {
-            $width = isset($args['width']) ? $args['width'] : null;
-            $height = isset($args['height']) ? $args['height'] : null;
+            $width = (isset($args['width']) && is_numeric($args['width'])) ? $args['width'] : null;
```
dnsl48: That would accept some incorrect values, e.g. `+123.45e6`.
Since `ResizedImage` accepts `int` only, shall we rather check for `ctype_digit` and cast to integer afterwards?

E.g.

```php
ctype_digit((string) $args['width']) ? (int) $args['width'] : null;
```

robbieaverill: Are decimal points acceptable in these attributes? Even if they are it's probably not common for them to be used.

dnsl48: Even if they could be, `ResizedImage` method only accepts `int` anyways

madmatt:
This also happened to me after running a migration of content from 3.1 -> 4.x. "Invalid" shortcodes are generated with for example `width="800" height=""` which leads to a height of `""` being passed through to this method.

While I agree that a different check might be stronger, the important thing is that the later call to `ResizedImage()` will call `castDimension()` which makes the same `is_numeric` check and then uses `intval` to cast to an integer. So this PR is consistent, and just moves the check up to a layer where the failure mode is 'drop an obviously invalid input' rather than 'throw an exception for an invalid input'.

dnsl48:
In my opinion, we should not make any assumptions about how `ResizedImage` is implemented and what it calls internally. We should rather be complying with its own API declaration which expects `int` as input values.

emteknetnz:

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

PR #358, opened 2019-12-11
FIX: Fixes #352 with guard for Folder query result

### LINKED ISSUE #352: Task fails with incomplete folder check
On line 103 of `SecureAssetsMigrationHelper.php`, the query for the `Folder` `DataObject` subclass naively assumes a record will always be returned, which of course is rarely always the case. Therefore a guard should be added before calling the method (any method) `getFilename()` on a potentially `null` query result:

FROM:

`$migratedPath = $this->migrateFolder($filesystem, $folder->getFilename());`

TO:

```
/** @var Folder $folder */
if (!$folder = Folder::get()->byID($id)) {
    $this->logger->info('Folder not found, skipping');
    continue;
}

$migratedPath = $this->migrateFolder($filesystem, $folder->getFilename());
```

phptek:
Fixes #352

### DIFF
--- a/src/Dev/Tasks/SecureAssetsMigrationHelper.php
+++ b/src/Dev/Tasks/SecureAssetsMigrationHelper.php
@@ -99,8 +99,13 @@ class SecureAssetsMigrationHelper
 
         foreach ($securedFolders->execute()->map() as $id => $path) {
             /** @var Folder $folder */
-            $folder = Folder::get()->byID($id);
+            if (!$folder = Folder::get()->byID($id)) {
+                $this->logger->warning(sprintf('No Folder record found for ID %d. Skipping', $id));
+                continue;
+            }
+
             $migratedPath = $this->migrateFolder($filesystem, $folder->getFilename());
+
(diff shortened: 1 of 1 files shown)


phptek:
I created the patch off "master". Let me know if that should have come out of "1" instead :-)

Cheddam:
@phptek Thanks for raising this PR! If you want the fix to be released in the next minor release (4.6.0, as 4.5 has already been branched in preparation for release), you'll want to target the `1` branch. If you'd like it to appear in a patch update to a current minor release, you'll want to target it to the corresponding `1.x` branch. `master` tracks the next _major_ version, so you'd be waiting a while for that to be released 😉

My recommendation would be to target `1.5`.

robbieaverill:
I've rebased onto 1.5 👍

phptek:
Cheers @robbieaverill Any chance of a release tag so I can stop running off a fork? Thanks :-)

Cheddam:
@phptek You can specify `1.5.x-dev` to run the development branch until a release is created.

Issue #197, opened 2018-11-30
?flush clears Intervention_Manipulations cache files


christopherdarling:
Running SS 4.3.0-rc1

I have noticed that the first load of pages after deploying, running dev/build?flush=1 pages are incredibly slow on their first load (10-20s! granted, the pages have a lot of large images which are to be resampled). After some debugging, this appears to be down to InterventionImage caches being cleared by ?flush.

Intervention_Manipulations store file dimensions, these are used on every getWidth() / getHeight() call to a Image (including resampled images), this can be quite expensive when a lot of images are included on a page (especially large ones as they're cloned to temp files to extract file size (not quite sure why, perhaps for S3 etc)). IMO Asset filesizes are unlikely to change when a flush is required so not quite sure this is necessary to clear the cache.

christopherdarling:
Still an issue on SS 4.4.3. Looking at `InterventionBackend` it looks as though removing `Flushable` would do the trick but there may be some cases where you would want to empty the cache. It looks as though it stores;
- Dimensions
- Failures

If you don't feel as though it is safe to remove the cache flushing on this class, would you accept a PR with this as a config setting for BC?

sb-relaxt-at:
This has a large performance impact. Having a page with some hundred large images the recreation of the dimensions cache takes some time…

dnsl48:
> If you don't feel as though it is safe to remove the cache flushing on this class, would you accept a PR with this as a config setting for BC?

Yes, please. PRs are welcome.

To make the solution more complete I'd suggest you to consider adding a simple `dev/task` that would clean the cache on-demand (through manual invocation). The task could be as simple as the original `Flushable::flush` implementation of `InterventionBackend`.

christopherdarling:
@dnsl48 PR at #362

dnsl48:
An opt-in disabling of Intervention cache invalidation on flush will be released in `1.6.0` (`Silverstripe CMS 4.6.0`)

sheadawson:
```
SilverStripe\Assets\InterventionBackend:
  flush_enabled: false
```

PR #362, opened 2019-12-17
add config setting + task for flushing IntervetionBackend cache

christopherdarling:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/197 when config is set to false. I have left the default value as true to match the previous behavior allowing users to opt-in to these caches not being cleared.

Was also tempted to change `$cache = Injector::inst()->get(CacheInterface::class . '.InterventionBackend_Manipulations');` to `$cache = (new static())->getCache();` - if you want that including in, or a seperate PR let me know.

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -19,6 +19,7 @@ use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Config\Config;
 
 class InterventionBackend implements Image_Backend, Flushable
 {
@@ -33,6 +34,14 @@ class InterventionBackend implements Image_Backend, Flushable
      * Cache prefix for dimensions
      */
     const CACHE_DIMENSIONS = 'DIMENSIONS_';
+
+    /**
+     * Is cache flushing enabled?
+     *
+     * @config
+     * @var boolean
+     */
+    private static $flush_enabled = true;
 
     /**
      * How long to cache each error type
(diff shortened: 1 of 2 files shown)


src/InterventionBackend.php
```diff
@@ -812,9 +820,11 @@ public function __destruct()
      */
     public static function flush()
     {
-        /** @var CacheInterface $cache */
-        $cache = Injector::inst()->get(CacheInterface::class . '.InterventionBackend_Manipulations');
-        $cache->clear();
+        if (static::$flush_enabled) {
```
dnsl48: I think we may want to fetch this from configs explicitly (e.g. `Config::get`)

christopherdarling: Completed with $this->config() to be consistent with the rest of the code in this file.

Do we want any @deprecated notices adding for v5?

christopherdarling:
@dnsl48 Change request completed above. Do we want any @deprecated notices adding? If so, what version?

dnsl48:
@christopherdarling thanks for the patch! :+1: 
Sorry, I haven't got around this one yet. Will try to have a look later this week.

dnsl48:
> Do we want any @deprecated notices adding? If so, what version?

At this point I think we should rather stick to the safer default, which is to invalidate cache on flush. For the use cases with hundreds of images per page (which feels to be a rather exceptional than common use case, although I'm completely guessing here), I will write some documentation notes about this new feature.

dnsl48:
One more thing before we merge. Could you please rewrite the commit messages to follow our [contribution guides](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages) so they get picked up by our changelog generator? (e.g. add NEW to the beginning of the messages). You could also squash them into a single one if you'd like. Then you could simply force push into this branch.

@christopherdarling

christopherdarling:
@dnsl48 great - thank you. I presume this will be in a 1.5.0 release of silverstripe-assets?  Thanks for reviewing and look forward to this being merged.

dnsl48:
@christopherdarling 1.5.0 has already been released (the latest version now is `1.5.1`)
This is a new feature and merged into the branch `1`, which will be taken as the base for the next minor release (`1.6.0` of silverstripe/assets, included into `4.6.0` release of the silverstripe/installer)

Issue #330, opened 2019-09-18
Multiple "Uploads" folders are being created


adrhumphreys:
### The issue
At some point during normal CMS user activity new uploads folders are being created with the exact same name as prior ones. This poses a large issue as it fixing this means reconciling different trees of assets back into the one Uploads folder.

We are running on `"silverstripe/recipe-cms": "4.4.3"`
We have fluent installed alongside it

We have been able to confirm that the folders are being created with assets inside of them. From actions such as uploading an asset via an upload field or from replacing an asset in the asset-admin.

We have confirmed that the folders should not be treated differently by fluent

We have not been able to reproduce the issue locally, but it has occured on a development server and a staging/testing server so we know it's not a one off with the server.

This has only started showing up since upgrading from `"silverstripe/recipe-cms": "4.3.3"` in which we haven't seen the issue occur before then (site has been running for over a year, with over 800gb in assets and continuous uploads) which makes us think that it's from a recent issue.

We've tried a good chunk of methods in recreating this via code but no matter what we always hit some validation step of some kind that prevents us from having the same folder name

Here's a copy of our `File` table for all records named "Uploads":
```
INSERT INTO trains.File (ID, ClassName, LastEdited, Created, Version, Caption, Credit, VisualLibraryID, Expires, MediaType, OwnerType, CanViewType, CanEditType, Name, Title, ShowInSearch, RegionID, CityID, ParentID, OwnerID, FileHash, FileFilename, FileVariant) VALUES (35, 'SilverStripe\\Assets\\Folder', '2018-09-04 08:44:03', '2018-09-04 08:44:03', 2, null, null, null, null, 'Other', null, 'Inherit', 'Inherit', 'Uploads', 'Uploads', 1, 0, 0, 0, 2, null, null, null);
INSERT INTO trains.File (ID, ClassName, LastEdited, Created, Version, Caption, Credit, VisualLibraryID, Expires, MediaType, OwnerType, CanViewType, CanEditType, Name, Title, ShowInSearch, RegionID, CityID, ParentID, OwnerID, FileHash, FileFilename, FileVariant) VALUES (91610, 'SilverStripe\\Assets\\Folder', '2019-08-20 17:01:02', '2019-08-20 17:01:02', 1, null, null, null, null, 'Other', null, 'Inherit', 'Inherit', 'Uploads', 'Uploads', 1, 0, 0, 0, 73, null, null, null);
[...]

PR #363, opened 2019-12-20
Folder won't find or make in archived mode

adrhumphreys:
We had this issue where the folder was made later than the version of the page we were looking at which meant each time you viewed the archived page a new folder would be made as it wouldn't exist at that time

Should solve: https://github.com/silverstripe/silverstripe-assets/issues/330


robbieaverill:
Closing - if #364 is merged then it will be merged up to this branch as well, we don't need both. Thanks for the PRs!

PR #364, opened 2019-12-20
Folder won't find or make in archived mode

adrhumphreys:
We had this issue where the folder was made later than the version of the page we were looking at which meant each time you viewed the archived page a new folder would be made as it wouldn't exist at that time

Should solve: #330

### DIFF
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -68,6 +68,12 @@ class Folder extends File
                 'Name' => array($partSafe, $part)
             ))->first();
 
+            // In archived mode we would expect for folders to sometimes not exist and therefore shouldn't create them
+            if (class_exists(Versioned::class)
+                && strpos(Versioned::get_reading_mode(), 'Archive.') !== false) {
+                return null;
+            }
+
             if (!$item) {
                 $item = new Folder();
                 $item->ParentID = $parentID;
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Assets;
 
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleResourceLoader;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\ValidationResult;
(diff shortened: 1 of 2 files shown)


unclecheese:
One way could make it less coupled would be to take the previous dataQuery, and check the query param.

```php
$query = Folder::get()->filter(...);
$folder = $query->first();
$mode = $query->getQueryParam('Versioned.mode');
```

That would fail gracefully if Versioned isn't installed.

adrhumphreys:
@maxime-rainville I've added tests which picked up an issue 🎉  so fixed that up

@dhensby / @unclecheese I ended up using Injector to search for Versioned. I feel like that is still coupled by the name of how we register it, but it's now just checking that Versioned is registered rather than a class existing etc

tests/php/FolderTest.php
```diff
@@ -428,4 +428,29 @@ public function testChildrenEnsureParentsPublish()
...
+            Versioned::set_reading_mode('Archive.2000-01-01 12:00:00.Live');
+            $folder = Folder::find_or_make('Folder-that-does-not-exist');
+            // Since this folder doesn't exist we would expect the result to be null
+            $this->assertEmpty($folder);
+
+            // Since this folder doesn't  exist at that date and time we would expect the result to be null
+            $folder = Folder::find_or_make('Folder-that-does-exist');
+            $this->assertEmpty($folder);
+
+            // Hopefully in 3000 years we've found a better date setting
```
maxime-rainville, line 450: Just call `DBDatetime::set_mock_now()` before creating your folder. That way your test will still work on 3000-01-02.

src/Folder.php
```diff
@@ -68,6 +69,16 @@ public static function find_or_make($folderPath)
                 'Name' => array($partSafe, $part)
             ))->first();
 
+            // When in archived mode, find or make should not find folders that don't exist
+            // We check explicitly for Versioned and if it exists then we'll confirm the reading mode isn't archived
+            $versioned = Injector::inst()->get('SilverStripe\Versioned\Versioned');
```
maxime-rainville, line 74:
```suggestion
            $versioned = Injector::inst()->get(Versioned::class);
```

adrhumphreys: The intent here is to decouple it _slightly_ from versioned, this allows us to skip needing Versioned to exist by referencing the FQN for if it does exist. That addition/change was largely driven by the conversation above with the worry of tight coupling

adrhumphreys:
I've added the additional scenarios as required, please have another review

src/Folder.php
maxime-rainville: You won't get an error if you reference a class that doesn't exists. So using `Versioned::class` doesn't increase the coupling between asset and versioned.

brynwhyman:
👏

PR #372, opened 2020-02-14
Fix wording in comment in assets htaccess

aNickzz:
The wording "Block invalid" sent me on a goose chase since all I did was confirm the extension I wanted was _not_ in this list as I assumed it was a blacklist.


Cheddam:
Thanks @aNickzz!

PR #379, opened 2020-03-12
Extensibility improvements

mfendeksilverstripe:
# Extensibility improvements

* Allow the ability to subclass and override some functions
* used in a feature where all file migration tasks can be executed as jobs which indicate progress and process much smaller chunks of work as opposed to processing everything in one go

This is still **Work in progress**


mfendeksilverstripe:
Replaced by https://github.com/silverstripe/silverstripe-assets/pull/396

Issue #380, opened 2020-03-17
Backslash in file name lead to folders in Hash folder


maxime-rainville:
# Description
If you programmatically create a file with a backslash in the filename, assets get confused and tries to create a folder inside the hash folder.

# Steps to reproduce
* Make a call like this `$file->setFromString('hello', 'BadFolder\test.txt');`

Expected results: 
* Either an exception is thrown or 
* ideally all the `\` symbols get swapped for a `/` and the file is saved at `assets/.protected/BadFolder/[hash]/test.txt`

Actual results: No warning are and the file is saved at `assets/.protected/[hash]/BadFolder/test.txt`

# Pull requests
* [ ] https://github.com/silverstripe/silverstripe-assets/pull/381

emteknetnz:
PR's welcome ;-)

maxime-rainville:
Got a PR: https://github.com/silverstripe/silverstripe-assets/pull/381

I don't think think this is a high impact bug. If you're putting backslashes in your filename, you are misusing our the asset system. It's more like a convenience enhencement. So I've only fixed for the next minor.

brynwhyman:
Is this good to close with #381 ? :)

emteknetnz:
^ Yup, PR was merged, can close issue

PR #381, opened 2020-03-18
BUG Update FileIDHelpers to replace backslashes with forward slashes

maxime-rainville:
This simple fix simply updates the file ID to replace backslashes with forward slashes.

The point of this is to fix an issue where if you set the content of a file with backslashes in the filename, you end up with folders inside the hash folder.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/380

### DIFF
--- a/src/FilenameParsing/HashFileIDHelper.php
+++ b/src/FilenameParsing/HashFileIDHelper.php
@@ -73,6 +73,9 @@ class HashFileIDHelper implements FileIDHelper
 
     public function cleanFilename($filename)
     {
+        // Swap backslash for forward slash
+        $filename = str_replace('\\', '/', $filename);
+
         // Since we use double underscore to delimit variants, eradicate them from filename
         return preg_replace('/_{2,}/', '_', $filename);
     }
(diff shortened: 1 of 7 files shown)


src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -73,6 +73,9 @@ public function buildFileID($filename, $hash = null, $variant = null, $cleanfile
 
     public function cleanFilename($filename)
     {
+        // Swap backslash for forward slash
+        $filename = str_replace('\\', '/', $filename);
```
emteknetnz: does this have any impact running on windows?

maxime-rainville: No we're not meant to be using \ even on windows. When you actually end up writing files, flysystem will swap backslashes for forward slashes. However a lot of our own logic won't.

This is why the current logic doesn't fail outright, but places the file in an unexpected location.

emteknetnz: ok sounds good

PR #383, opened 2020-03-29
PHPUnit 8 upgrade

dhensby:
This depends on https://github.com/silverstripe/silverstripe-framework/pull/9443

It also includes a merge of the 1 branch into master

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /artifacts/
+/vendor/
+/silverstripe-cache/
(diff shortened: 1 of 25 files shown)


dhensby:
This PR is now in the same state as the current master (failing tests when trying to call `allVersions()` on files). I'm not entirely sure why this is failing - maybe Files are not versioned by default any more and the test suite never got updated?

dhensby:
GREEN?

PR #386, opened 2020-04-09
(old) Add canViewAnonymous()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1066 (includes links to related PRs)

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -501,7 +501,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      */
     public function isUserDefinedFormUpload(): bool
     {
-        if (!class_exists('SilverStripe\\UserForms\\Model\\Submission\\SubmittedFileField')) {
+        if (!class_exists('SilverStripe\UserForms\Model\Submission\SubmittedFileField')) {
             return false;
         }
         return SubmittedFileField::get()->find('UploadedFileID', $this->ID) ? true : false;
--- a/src/File.php
+++ b/src/File.php
@@ -11,6 +11,7 @@ use SilverStripe\Assets\Storage\AssetContainer;
 use SilverStripe\Assets\Storage\AssetNameGenerator;
 use SilverStripe\Assets\Storage\DBFile;
 use SilverStripe\Control\Director;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleLoader;
@@ -25,6 +26,8 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\HasManyList;
 use SilverStripe\ORM\Hierarchy\Hierarchy;
+use SilverStripe\ORM\Queries\SQLSelect;
+use SilverStripe\ORM\Tests\MySQLDatabaseTest\Data;
 use SilverStripe\ORM\ValidationResult;
 use SilverStripe\Security\InheritedPermissions;
 use SilverStripe\Security\InheritedPermissionsExtension;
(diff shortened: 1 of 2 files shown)


src/File.php
```diff
@@ -485,6 +485,16 @@ public function canDelete($member = null)
             ->canDelete($this->ID, $member);
     }
 
+    /**
+     * Check if the File is viewable to a public user i.e. someone who is not logged in to the CMS
+     */
+    public function canViewAnonymous(): bool
```
emteknetnz, line 491: This should possibly be on DataObject instead

src/File.php
```diff
@@ -485,6 +486,27 @@ public function canDelete($member = null)
...
+            return $this->canView();
+        });
+    }
+
+    /**
+     * Whether the file was originally uploaded in a userform submissin on a FileFiled
+     */
+    public function isUserDefinedFormUpload(): bool
+    {
+        if (!class_exists('SilverStripe\\UserForms\\Model\\Submission\\SubmittedFileField')) {
```
emteknetnz, line 504: can use either `\` or `\\` here, single looks prettier and is a bit more standard throughout the rest of the modules

michalkleiner:
Can't find on which of the other related PRs I've mentioned it... It'd be nice if this didn't need to be tied to user defined forms — a file can be uploaded through a custom user-facing form as well.

emteknetnz:
@michalkleiner how would you identify if the file was uploaded by a custom user form?  I'm not aware of any mechanism / db column that says 'this was uploaded outside of the cms'

michalkleiner:
@emteknetnz I don't know exactly. An idea might be a flag on the actual File class that becomes managed by the assets-admin and if it's empty, and a dev didn't populate it specifically via other code, the file would be treated as CMS uploaded. Otherwise it'd be 'foreign'. Or some combination of this and another approach etc. Userforms are just _a_ module and IMHO shouldn't be treated _special_ in other modules. Plenty of sites don't use them and still have front-end upload fields.

src/File.php
kinglozzer: `class_exists(SubmittedFileField::class)` will work too if you’ve already imported the class name 😉

emteknetnz: yeah though what if you haven't actually imported it :-) The `use` statement to import the class will do nothing and not error if the class does not exist.  It'll be pretty normal for an install to not have user-defined-forms

kinglozzer: Yep but PHP will magically resolve it to a string, so `class_exists()` will work without erroring 😄 https://3v4l.org/gf3kr

emteknetnz: well ok, learn something everyday :-)

src/File.php
```diff
@@ -485,6 +489,56 @@ public function canDelete($member = null)
...
+    {
+        return Member::actAs(null, function(): bool {
+            return $this->canView();
+        });
+    }
+
+    /**
+     * Whether the file was originally uploaded in a userform submissin on a FileFiled
+     */
+    public function isUserDefinedFormUpload(): bool
```
maxime-rainville, line 505: I think this would make more sense as an Extension applied from the userforms module. I would rather avoid tying core module with non-core module if possible.

maxime-rainville:
Agree that we should avoid tie-in with other modules. I'm fine with updating core modules to provide hooks for other modules.

FYI there's already ways to define [ownership relations between DataObjects](https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#defining-ownership-between-related-versioned-dataobjects) and to trigger [cascade deletion](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#cascading-deletions). That might be something to keep in mind on this issue or the wider epic.

src/File.php
maxime-rainville: Arguably we should have written the `canView` methods to accept a value that means "Anonymous". Unfortunately, any falsy value gets converted to the current user which is annoying.

maxime-rainville: If we call this method on every file every time you display a list of files, it might end up being very expansive.

You might be able to do something like get the permission of the parent folder and have the front-end or graphql decide if if the file is restricted or not based on each file's CanViewType property.

emteknetnz: hmm yeah not too keen to build a separate canView() set of logic on the front end.  If it does turn out to be too expensive then I'd way rather prefer to look at optimising things in the backend

emteknetnz:
Ownership is an interesting possibility, will keep that in mind

Still to be determined if we'll have time to get the 'is form submission' functionality into this sprint, and if so, whether we calculate it dynamically (easier) or save it into the File table (so if you delete a form submission the File is still aware that it was originally uploaded outside of the cms)

emteknetnz:
Moved to https://github.com/silverstripe/silverstripe-asset-admin/pull/1087

src/File.php
```diff
@@ -485,6 +489,56 @@ public function canDelete($member = null)
...
+     * Whether this Folder has child File's that where uploaded via a UserDefinedForm FileField
+     * This only checks for direct children, it will not check for:
+     * - Grandchild Files
+     * - Child Folders that themselves have child Files that are uploads
+     *
+     * This function is File.php instead of Folder.php as the asset-admin graphql fileFragments.js
+     * seems to treats Folder as a 'FileInterface' i.e the top-level class, instead of as a subclass of File
+     * instead, File is treated as a sub-class of FileInterace
+     */
+    public function hasChildUserDefinedFormUploads(): bool
```
emteknetnz, line 523: can probably remove this now

emteknetnz:
New PR https://github.com/silverstripe/silverstripe-assets/pull/398

PR #391, opened 2020-04-17
API Add new updateResponse hook to allow extension to update the response

maxime-rainville:
This arise from a discussion with @madmatt who needs a way to redirect old asset url to user defined new addressed.

This could be helpful to anyone who might need to highjack the HTTP response to add an extra header or do some other weird thing to it.

# Related PR
* https://github.com/silverstripe/silverstripe-framework/pull/9514

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -23,6 +23,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPStreamResponse;
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Security\Security;
@@ -33,6 +34,7 @@ use SilverStripe\Security\Security;
 class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
 {
     use Configurable;
+    use Extensible;
 
     /**
      * Session key to use for user grants
@@ -1460,6 +1462,23 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
 
     public function getResponseFor($asset)
     {
+        /** @var HTTPResponse $response */
+        /** @var array $context */
+        [$response, $context] = $this->generateResponseFor($asset);
+
+        // Give a chance to extensions to tweak the response
+        $this->extend('updateResponse', $response, $asset, $context);
+
+        return $response;
+    }
+
+    /**
+     * Build a response for getResponseFor along with some context information for the `updateResponse` hook.
+     * @param string $asset
+     * @return array HTTPResponse and some surronding context
+     */
+    private function generateResponseFor(string $asset): array
+    {
         $public = $this->getPublicFilesystem();
         $protected = $this->getProtectedFilesystem();
         $publicStrategy = $this->getPublicResolutionStrategy();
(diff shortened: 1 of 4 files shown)


maxime-rainville:
If we think this is helpful I'll add some docs and unit test around it.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1488,16 +1498,38 @@ public function getResponseFor($asset)
...
+    /**
+     * Wrap an "updateResponse" extend call around the response along with some extra context.
+     * @param HTTPResponse $response Response that could be updated
+     * @param string $asset Original file id of the request
+     * @param array $context Additional context to send along
+     * @return HTTPResponse Updated response
+     */
+    private function extendResponse(HTTPResponse $response, string $asset, array $context = [])
+    {
+        $this->extend('updateResponse', $response, $asset, $context);
```
maxime-rainville, line 1531: Context is just meant to provide extra details about the request and how the response was built. At this stage it could contain a `visibility` attribute to say what store the file was found on and a `parsedFileID`.

michalkleiner:
Interesting idea, I'm definitely for more hooks where it makes sense.

Without diving too much into the detail, the first gut-feel is that the context params should be passed to the `createResponseFor` method and the extension hook added there, instead of wrapping it in an extend method that gets an update hook for itself.

madmatt:
Yeah I like that idea too @michalkleiner, except `createResponseFor` is only used for two of the `return` statements, and there's 3 responses that don't use that method, so I see what Max is going for. I'll be testing this with the `redirectedurls` module, will see how it goes.

maxime-rainville:
What about something like this https://github.com/silverstripe/silverstripe-assets/pull/391/commits/3bee71d4d6df144407611c02eff962ce9d08aa16

It's functionally equivalent but it might be a bit easier to understand.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1459,6 +1461,23 @@ public static function flush()
     }
 
     public function getResponseFor($asset)
+    {
+        /** @var HTTPResponse $response */
+        /** @var array $context */
+        [$response, $context] = $this->generateResponseFor($asset);
```
maxime-rainville, line 1467: That's a shorthand for `list($response, $context) = $this->generateResponseFor($asset);`

madmatt: This seems over-complicated - should we just use `list()` to make it clearer? I'm not fussed either way...

madmatt:
I've added a PR to the silverstripe/redirectedurls module that indicates how I'd expect to consume this here: silverstripe/silverstripe-redirectedurls/pull/80

src/Flysystem/FlysystemAssetStore.php
maxime-rainville: It's a valid PHP syntax that meet our minimum PHP version. 

My general view is that we shouldn't hold back using new PHP feature if they meet our minimum PHP requirements ... otherwise you get into a chicken and egg problem where you can't use new syntax because people aren't familiar with it and people are not familiar with the new syntax because you're not using it.

michalkleiner: Yeah, the array syntax (instead of `list()`) is pretty standard nowadays.

maxime-rainville:
Just added some unit test. I'll draft some doc as well.

@silverstripe/core-team Do you guys have beef with adding this API?

maxime-rainville:
Here's the PR for the DOCs https://github.com/silverstripe/silverstripe-framework/pull/9514

maxime-rainville:
I've fixed the broken unit test and squashed/rebased.

madmatt:
Fantastic, thanks team!

PR #393, opened 2020-04-18
Use short array syntax across the module's codebase

michalkleiner:
As per the title — use `[]` instead of `array()`.


michalkleiner:
Thanks @dhensby, that's a valid point. Rule added.

PR #398, opened 2020-05-07
Methods to support file status icons

emteknetnz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/1066

Old PR: https://github.com/silverstripe/silverstripe-assets/pull/386/files

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -22,7 +22,6 @@ use SilverStripe\Forms\TextField;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\ORM\DataObject;
-use SilverStripe\ORM\DB;
 use SilverStripe\ORM\HasManyList;
 use SilverStripe\ORM\Hierarchy\Hierarchy;
 use SilverStripe\ORM\ValidationResult;
@@ -269,6 +268,12 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      */
     private static $file_types = [];
 
+    /**
+     * @internal
+     * @see hasRestrictedPermissions
+     */
+    private static $has_restricted_permissions_cache = [];
+
     public static function get_shortcodes()
     {
         return 'file_link';
@@ -483,6 +488,61 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         // Check inherited permissions
         return static::getPermissionChecker()
             ->canDelete($this->ID, $member);
+    }
+
+    /**
+     * Check if the File has a CanViewType set or inherited at "LoggedInUsers or above"
+     *
+     * This is a bit different from a canView() check in that it doesn't check anything against a member.
+     * Instead it is a member-less permission check
+     *
+     * @return bool
+     */
+    public function hasRestrictedAccess(): bool
+    {
+        return $this->hasRestrictedPermissions($this);
+    }
+
+    /**
+     * Recursively determine whether a File has, or inherits, restricted permissions.
+     *
+     * @param File $file
+     * @return bool
+     */
+    private function hasRestrictedPermissions(File $file): bool
+    {
+        $id = $file->ID;
+        $parentID = $file->ParentID;
+        $canViewType = $file->CanViewType;
+        if (in_array($canViewType, [InheritedPermissions::LOGGED_IN_USERS, InheritedPermissions::ONLY_THESE_USERS])) {
+            self::$has_restricted_permissions_cache[$id] = true;
+            return true;
+        }
+        if ($canViewType == InheritedPermissions::INHERIT && $parentID != 0) {
+            if (isset(self::$has_restricted_permissions_cache[$parentID])) {
+                return self::$has_restricted_permissions_cache[$parentID];
+            }
+            $parent = $file->Parent();
+            if ($parent->exists()) {
+                $value = $this->hasRestrictedPermissions($parent);
+                self::$has_restricted_permissions_cache[$parentID] = $value;
+                return $value;
+            }
+        }
+        self::$has_restricted_permissions_cache[$id] = false;
+        return false;
+    }
+
+    /**
+     * If the file was uploaded via a form and tracked in the database
+     *
+     * @return bool
+     */
+    public function isTrackedFormUpload(): bool
+    {
+        $value = false;
+        $this->extend('updateTrackedFormUpload', $value);
+        return (bool) $value;
     }
 
     /**
--- a/tests/php/FileTest.php
+++ b/tests/php/FileTest.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Assets\Tests;
 
+use Generator;
 use PHPUnit_Framework_MockObject_MockObject;
 use Silverstripe\Assets\Dev\TestAssetStore;
 use SilverStripe\Assets\File;
@@ -998,5 +999,37 @@ class FileTest extends SapphireTest
         $file->CanEditType = 'Anyone';
         $file->ID = 234;
         $this->assertFalse($file->canEdit());
+    }
+
+    /**
+     * @return Generator
+     * @see testHasRestrictedAccess
+     */
+    public function restrictedAccessDataProvider()
+    {
+        yield ['restricted-test-r', false];
+        yield ['restricted-test-r1', false];
+        yield ['restricted-test-r11', false];
+        yield ['restricted-test-r111', false];
+        yield ['restricted-test-r12', true];
+        yield ['restricted-test-r121', true];
+        yield ['restricted-test-r2', true];
+        yield ['restricted-test-r21', true];
+        yield ['restricted-test-r211', true];
+        yield ['restricted-test-r22', true];
+        yield ['restricted-test-r221', false];
+    }
+
+    /**
+     * @dataProvider restrictedAccessDataProvider
+     *
+     * @param string $fixtureName
+     * @param bool $expected
+     */
+    public function testHasRestrictedAccess(string $fixtureName, bool $expected)
+    {
+        /** @var Folder $folder */
+        $folder = $this->objFromFixture(Folder::class, $fixtureName);
+        $this->assertSame($expected, $folder->hasRestrictedAccess());
     }
 }
--- a/tests/php/FileTest.yml
+++ b/tests/php/FileTest.yml
@@ -64,6 +64,50 @@ SilverStripe\Assets\Folder:
   deep-folder:
     Name: super_deep_folder_with_underscores
     ParentID: =>SilverStripe\Assets\Folder.folder1-subfolder1
+  restricted-test-r:
+    Name: 'restricted-test-r'
+    CanViewType: Anyone
+  restricted-test-r1:
+    Name: 'restricted-test-r1'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r
+    CanViewType: Inherit
+  restricted-test-r11:
+    Name: 'restricted-test-r11'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r1
+    CanViewType: Inherit
+  restricted-test-r111:
+    Name: 'restricted-test-r111'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r11
+    CanViewType: Inherit
+  restricted-test-r12:
+    Name: 'restricted-test-r12'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r1
+    CanViewType: LoggedInUsers
+  restricted-test-r121:
+    Name: 'restricted-test-r121'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r12
+    CanViewType: Inherit
+  restricted-test-r2:
+    Name: 'restricted-test-r2'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r
+    CanViewType: LoggedInUsers
+  restricted-test-r21:
+    Name: 'restricted-test-r21'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r2
+    CanViewType: Inherit
+  restricted-test-r211:
+    Name: 'restricted-test-r211'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r21
+    CanViewType: Inherit
+  restricted-test-r22:
+    Name: 'restricted-test-r22'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r2
+    CanViewType: Inherit
+  restricted-test-r221:
+    Name: 'restricted-test-r221'
+    ParentID: =>SilverStripe\Assets\Folder.restricted-test-r22
+    CanViewType: Anyone
+
 SilverStripe\Assets\File:
   asdf:
     FileFilename: FileTest.txt


src/File.php
```diff
@@ -485,6 +484,49 @@ public function canDelete($member = null)
...
+     * Whether the file was originally uploaded in a userform submission by a FileField
+     *
+     * @see UserFormFileExtension::updateIsUserFormUpload
+     */
+    public function isUserFormUpload(): bool
+    {
+        $value = false;
+        $this->extend('updateIsUserFormUpload', $value);
+        return $value;
+    }
```
Cheddam, line 528: This should be more generic - something like `isTrackedFormUpload()` - so that we aren't referencing a completely optional module in a core one, and so that it's more sensible for other modules / developers to make use of in the future.

src/File.php
```diff
@@ -485,6 +484,49 @@ public function canDelete($member = null)
...
+     */
+    public function hasRestrictedAccess(): bool
+    {
+        return $this->hasRestrictedPermissions($this);
+    }
+
+    /**
+     * @param File $file
+     * @return bool
+     */
```
Cheddam, line 501:
```suggestion
    /**
     * Recursively determine whether a File has, or inherits, restricted permissions.
     *
     * @param File $file
     * @return bool
     */
```

composer.json
```diff
@@ -19,6 +19,7 @@
         }
     ],
     "require": {
+        "php": ">=7.1",
         "silverstripe/framework": "^4.5",
```
Cheddam, line 23: The `silverstripe/framework ^4.5` dependency here already covers this PHP requirement, so this isn't necessary. If you'd like to change the way we handle this, maybe put together an RFC.

src/File.php
```diff
@@ -485,6 +484,49 @@ public function canDelete($member = null)
...
+    private function hasRestrictedPermissions(File $file): bool
+    {
+        $canViewType = $file->CanViewType;
+        if (in_array($canViewType, [InheritedPermissions::LOGGED_IN_USERS, InheritedPermissions::ONLY_THESE_USERS])) {
+            return true;
+        } elseif ($canViewType == InheritedPermissions::INHERIT) {
+            if ($file->ParentID != 0) {
+                $parent = $file->Parent();
```
Cheddam, line 509: This will trigger another query for every traversal, right? I'm wondering what the performance is like once you get a few layers deep 🤔

composer.json
emteknetnz: We seems to be flip-floping on this.  Previously there had been some agreement that we'd put this requirement in where needed.  There has been at least one instance where the 'framework is always included' turned out to be a false assumption and it caused a travis failure.

src/File.php
emteknetnz: That's right, pretty inefficient.  Worse case scenario say subfolder nested 4 deep, 50 paginated records, 200 individual queries.  However I'd expect database caching to kick in as 150 of these queries will be the same as the other 50 queries 'on that level'

Without manually testing performance, I don't know if this is an issue worth solving or not

Options if we need them:
- InheritedPermissions class (I cannot recall my reasoning at the time for choosing not to use this)
- New MyCachingService class that does a few large queries and caches in memory i.e. only for the duration of the request

src/File.php
emteknetnz: I sort of agree, though at the same time it's pretty baked into the admin/asset-admin JS that we're looking a userform upload specifically

In theory this could be extended later to also cover 'regular form' uploads, however we've already decided early in this piece that's out of scope as you `[cannot retroactively identify uploads](https://github.com/silverstripe/silverstripe-userforms/issues/950#issuecomment-616320644)`, so it's pretty unlikely this will ever get extended

src/File.php
emteknetnz: Updated docblock

composer.json
Cheddam: My recollection of those discussions was that we'd add it directly for non-core modules, and for core modules, we'd keep it centralised in `framework` and make sure version requirements were kept in sync between core modules to avoid the issue we faced with the CMS module happening again.

src/File.php
Cheddam: I'm not sure whether `InheritedPermissions` would help here, as it doesn't return specific permission data, just a boolean canView based on a given user.

It'd be great to run some benchmarking against this, as I think it's probably most sensible to leave it as-is, but we'll want to warn users if we expect choppy performance upon initial upgrade.

src/File.php
Cheddam: I still think it's worth naming generically - our tooltips for this feature in the CMS make no reference to userforms, and I could easily see developers adopting this API in custom forms on their own projects. We should try not to reference downstream modules in core to avoid confusion and implied coupling.

emteknetnz: There is a scenario where:
- a brand new website is built with no existing data
- therefore don't need to worry about half the form uploads having missing data - which causes the icon to be displayed inconsistently (this is much worse than no icon at all as your CMS if full of false negatives.)
- developers add their custom logic to MyApp\FileExtensions::updateIsTrackedFormUpload() that somehow registers custom FormField uploads

However, there's still some issues with this:
- We're providing a method to developers to introduce missing data, basically any future dev work (post initial build, think years later) involving a form with uploads MUST do something with this extension method, and when someone forgets they'll end up with same inconsistent icon experience.
- We're essentially highjacking the userform submission icon to do something that it wasn't designed to do.  The reason the tooltips say 'Form submission' instead of 'UserForm submissions' is PUX's view that most cms-users won't know what a 'UserForm' is.  We had a long discussion of the tooltip saying 'Form' vs 'UserForm'.
- This is scope creep.  We specifically excluded catering for custom form uploads and we made no mention of extensibility when doing the specs for this work.

Therefore, my view is still that we should keep this as is.

composer.json
emteknetnz: Have removed php >=7.1

src/File.php
emteknetnz: Ugh, performance is pretty bad, even accounting for my slow local

 /admin/graphql request that gets file list

network panel open 'disable cache' (not that that should make a diff)

Root(Files) / 1 folder / 2 folder / 3 folder / 4 folder / 50 Files
(worst case scenario)

xhprof on the 51 requests to hasRestrictedAccess() (50 files + 1 for the current folder)

microsec 1,000,000 = 1 second

with db lookups
648,588
800,023
838,835
752,186
625,229
avg = 732,972

without db lookups, just return false
90

For comparison

Root(Files) / 1 folder / 50 Files
130,664

emteknetnz: > we expect choppy performance upon initial upgrade

It's straight choppy performance, unlike what we did with the userform lookup where we saved that to the database, we don't save the result of this one because it must stay dynamic

Will investigate `InheritedPermissions` to see if a `Member::actAs(null` will suffice for this (possibly not as we need to test this against draft-content), and failing that will do a custom caching service

src/File.php
emteknetnz:
```
# This doesn't work, doesn't read draft site:
return Versioned::withVersionedMode(function() {
    Versioned::set_reading_mode('Stage.' . Versioned::DRAFT);
    return Member::actAs(null, function() {
        return !$this->canView();
    });
});

# This works, fast performance, but is it risky?
$default = Versioned::get_draft_site_secured();
Versioned::set_draft_site_secured(false);
$value = Member::actAs(null, function() {
    return !$this->canView();
});
Versioned::set_draft_site_secured($default);
return $value;

2 runs:
11,232
35,774
```

src/File.php
emteknetnz: File::canView() actually already uses InheritedPermissions, though it doesn't quite work as I'd like:

```
# This doesn't work, doesn't read draft site:
# Gets caught up on Versioned::canViewVersioned()
return Versioned::withVersionedMode(function() {
    Versioned::set_reading_mode('Stage.' . Versioned::DRAFT); // same result if Versioned::LIVE
    return Member::actAs(null, function() {
        return !$this->canView();
    });
});

# This works, fast performance, but is it risky?
$default = Versioned::get_draft_site_secured();
Versioned::set_draft_site_secured(false);
$value = Member::actAs(null, function() {
    return !$this->canView();
});
Versioned::set_draft_site_secured($default);
return $value;

2 runs:
11,232
35,774
```

emteknetnz: Have instead added some basic memory caching into the function.

Works great, 1 run in xhprof was only 4,316

src/File.php
Cheddam: - Arguably this will be 'inconsistent' whether we explicitly make this API available to developers or not, thanks to the broad description used in the tooltips. By naming this as a pattern suitable for broader use, we at least give them the option to make it more consistent if they have the scope / ability.
- @sachajudd / @clarkepaul - Do you see the form submission indicator as being exclusive to UserForms, or would you be comfortable with it being used more broadly in projects where developers opted to apply it to their own Forms?
- I don't think this change fits the term 'scope creep'; if we went through with it, it'd require a batch of renaming and nothing more. We're not catering for it in terms of actually supporting it - we're catering for it in terms of allowing the possibility of it in the future.

Once this ships, it's locked in - even if we were to leave it undocumented for now and then seek to expose it more widely later, we wouldn't be able to adjust the naming until CMS 5 because it introduces public API surface.

Sensible extensibility is an NFR for features in the Open Source project.

sachajudd: > we wouldn't be able to adjust the naming until CMS 5 because it introduces public API surface.
Agree that this isn't great. Why don't we just batch rename it to 'isTrackedFormUpload' and then we actually have the option of allowing this possibility in the future? I don't see why there is such a big push back on this.

robbieaverill: I agree with Garion - please don't couple core modules to optional ones

robbieaverill: My suggestion would be to find out the reason why this method has to exist, e.g. because we want to distinguish userform uploads differently from regular assets in the assets area (for example, for some reason). If that's the example reason here then design your API around providing extensibility for that - e.g. `getIcon(): string` and have an extension point inside it.

emteknetnz: Have updated naming on all the PR's to `isTrackedFormUpload`

robbieaverill:
Nice work, it’s pretty clear. I’ve left some questions and suggestions. Can you please also provide some unit tests for this, and write a short description of the API you’re adding for your commit message?

emteknetnz:
@robbieaverill 
> Nice work

Chur

> I’ve left some questions and suggestions. 

I'm not seeing these, could you point me in the right direction?

robbieaverill:
Ha, whoops

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
 
+
+    /**
+     * Check if the File has a CanViewType set or inherited at "LoggedInUsers or above"
+     *
+     * @return bool
+     */
+    public function hasRestrictedAccess(): bool
+    {
+        return $this->hasRestrictedPermissions($this);
```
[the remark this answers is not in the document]
emteknetnz: `hasRestrictedAccess` has no params.

`hasRestrictedPermissions` has a param and is a recursive function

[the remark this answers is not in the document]
robbieaverill: Not sure you answered my question there

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+    /**
+     * Recursively determine whether a File has, or inherits, restricted permissions.
+     *
+     * This custom permission check is essentially a couldView() if the record was published for an
+     * anonymous (not-logged-in) user
+     *
+     * Ideally we wouldn't need this custom logic and instead use File->canView() which utilises
+     * InheritedPermission::batchPermissionCheck() and we'd wrap this in `Member::actAs(null,` and
+     * `Versioned::set_reading_mode('Stage.Stage')`, however the anonymous user will fail the
+     * `Versioned::canViewVersioned()` check
```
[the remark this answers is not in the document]
emteknetnz: > I think you could say here that you want a canView check but without other modules involved.

It's not actually a canView() check since that involves a `$member`.  This is a `$memberless` permission check, which is something I don't think we do anywhere else.  The docblock explanation is so people understand why we didn't just use existing methods

> did you consider dogfooding this new API?

By this do you mean updating the existing File::canView() to use this instead of the  InheritedPermissions logic it uses now?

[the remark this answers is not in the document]
robbieaverill: If this logic is new then all good, was just suggesting we reuse this new API if possible

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+        return false;
+    }
+
+    /**
+     * If the file was uploaded via a form and tracked in the database
+     */
+    public function isTrackedFormUpload(): bool
+    {
+        $value = false;
+        $this->extend('updateIsTrackedFormUpload', $value);
```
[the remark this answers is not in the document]
emteknetnz: hmm, i prefer to keep naming as consistent as possible to make it easier to track variable/function in different parts of the codebase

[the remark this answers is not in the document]
robbieaverill: The way I read these is get/set/is/has<Something> and “Is” the prefix and not part of the something

src/File.php
```diff
@@ -269,6 +268,11 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      */
     private static $file_types = [];
 
+    /**
+     * @see hasRestrictedPermissions
+     */
+    private static $has_restricted_permissions_cache = [];
```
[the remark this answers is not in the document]
emteknetnz: hmm :-/  well, this one is a minor release so the semver i'm not too worried about.  One option is I split it off into its own file as a `RestrictedPermissionsCacheService` though that seems kind of spammy.   Any other suggestions? How bad is it if it's left as is in the config layer?

src/File.php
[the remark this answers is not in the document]
robbieaverill: Example: updateCMSFields is not called updateGetCMSFields

src/File.php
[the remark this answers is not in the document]
robbieaverill: I think we can just put `@internal` on it actually

src/File.php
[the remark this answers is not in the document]
emteknetnz: Reason is I don't want to want the public method `hasRestrictedAccess` to have any params.

If it had an optional `($file = null)` then people could misuse this almost like it's a static method e.g. `$fileA->hasRestrictedAccess($fileB)`

I think it's better simply as `$fileA->hasRestrictedAccess()`

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+     * InheritedPermission::batchPermissionCheck() and we'd wrap this in `Member::actAs(null,` and
+     * `Versioned::set_reading_mode('Stage.Stage')`, however the anonymous user will fail the
+     * `Versioned::canViewVersioned()` check
+     *
+     * @param File $file
+     * @return bool
+     */
+    private function hasRestrictedPermissions(File $file): bool
+    {
+        $id = $this->ID;
```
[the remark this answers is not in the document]
emteknetnz: Believe it or not the function still returns the correct result if you incorrectly have `$this->ID` instead of the correct `$file->ID`, though the caching no longer works

Have updated

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+    }
+
+    /**
+     * If the file was uploaded via a form and tracked in the database
+     */
+    public function isTrackedFormUpload(): bool
+    {
+        $value = false;
+        $this->extend('updateIsTrackedFormUpload', $value);
+        return $value;
```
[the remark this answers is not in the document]
emteknetnz: updated

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+     * @return bool
+     */
+    private function hasRestrictedPermissions(File $file): bool
+    {
+        $id = $this->ID;
+        $canViewType = $file->CanViewType;
+        if (in_array($canViewType, [InheritedPermissions::LOGGED_IN_USERS, InheritedPermissions::ONLY_THESE_USERS])) {
+            self::$has_restricted_permissions_cache[$id] = true;
+            return true;
+        } elseif ($canViewType == InheritedPermissions::INHERIT) {
```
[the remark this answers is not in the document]
emteknetnz: updated

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+            $parentID = $this->ParentID;
+            if ($parentID != 0) {
```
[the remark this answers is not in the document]
emteknetnz: updated

src/File.php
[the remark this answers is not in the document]
emteknetnz: I've updated the comment to spell it out clearer that it's a member-less permission check

src/File.php
```diff
@@ -485,6 +489,65 @@ public function canDelete($member = null)
...
+                if (isset(self::$has_restricted_permissions_cache[$parentID])) {
+                    return self::$has_restricted_permissions_cache[$parentID];
+                } else {
+                    $parent = $file->Parent();
+                    if ($parent->exists()) {
```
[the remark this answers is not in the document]
emteknetnz: updated

src/File.php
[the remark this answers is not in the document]
emteknetnz: changed to `updateTrackedFormUpload`

src/File.php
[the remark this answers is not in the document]
emteknetnz: added `@internal`

tests/php/FileTest.php
```diff
@@ -999,4 +999,41 @@ public function testCanEditFallsBackToCheckingDefaultFilePermissions()
         $file->ID = 234;
         $this->assertFalse($file->canEdit());
     }
+
+    public function testHasRestrictedAccess()
```
dhensby, line 1003: This test looks like a prime candidate to use a [data provider](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers)

emteknetnz: (I've never used dataProvider before)

I'm open-minded to refactoring to this, though
- Would all the DataObjects be created properly and all the Parent() associations as well?
- Do we actually want to encourage two ways of doing something instead of just using one? i.e. Creating DataObjects yml fixtures (de facto standard) vs dataProvider (new and maybe better standard)?

robbieaverill: You can't really scaffold anything dynamic in dataproviders well. I think here Dan's suggestion would be to pass the fixture name and the expectation:

```php
public function restrictedAccessProvider()
{
    yield ['restricted-test-r', false];
    yield ['restricted-test-r1', false];
    yield ['restricted-test-r11', false];
}

/**
 * @dataProvider restrictedAccessProvider
 */
public function testHasRestrictedAccess(string $fixtureName, bool $expected)
{
    $fixture = $this->objFromFixture(Folder::class, $fixtureName);
    $this->assertSame($expected, $fixture->hasRestrictedAccess());
}
```

etc

emteknetnz: cool, have implemented

PR #400, opened 2020-05-28
Target 4.6 minor branches in build

maxime-rainville:


.travis.yml
```diff
@@ -8,8 +8,7 @@ services:
 
 env:
   global:
-    - COMPOSER_ROOT_VERSION=1.x-dev
-    - CORE_RELEASE=master
```
maxime-rainville, line 12: This doesn't look like it serves any purpose so I blasted it.

PR #403, opened 2020-06-09
FIX Ensure session grant config is respected on uncached file shortcodes

### LINKED ISSUE #220: File::canView() not checked when a File is linked from a Page
Splitting out of https://github.com/silverstripe/silverstripe-framework/issues/8493 which is bigger and more amorphous.

If you have Page that embeds a File, the File's canView() permission is _not_ checked before granting access to the File. This breaks the security model, although requires that an author links to a protected file from a non-protected page in the CMS, so I would rate it as low severity (which is why I've posted it to the public issue list).

The relevant code is here: https://github.com/silverstripe/silverstripe-assets/blob/1/src/Shortcodes/FileShortcodeProvider.php#L72

This check should be checking File::canView(). If we do this, that should address Ingo's concern that custom canView() implementations wouldn't be respected.

**Edit:** This issue has been reopened. See [comment](https://github.com/silverstripe/silverstripe-assets/issues/220#issuecomment-571856130) below

## PRs
 - [x] https://github.com/silverstripe/silverstripe-assets/pull/402
- [x] https://github.com/silverstripe/silverstripe-assets/pull/403

Cheddam:
Supplementary fix for CVE-2019-14273, adjusting the logic to ensure that the `FileShortcodeProvider::$allow_session_grant config` is still applied when a shortcode is uncached.

`File::Link()` should not be used unless the user _must_ be guaranteed access to the asset and the security implications are sufficiently considered and mitigated.

Fixes #220.

### DIFF
--- a/src/Dev/TestAssetStore.php
+++ b/src/Dev/TestAssetStore.php
@@ -18,9 +18,11 @@ use SilverStripe\Assets\Storage\DBFile;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Folder;
 use SilverStripe\Assets\Storage\GeneratedAssetHandler;
+use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\TestOnly;
+use SilverStripe\Security\Security;
 use SilverStripe\View\Requirements;
 
 /**
@@ -192,6 +194,17 @@ class TestAssetStore extends FlysystemAssetStore implements TestOnly
     public function getDefaultConflictResolution($variant)
     {
         return parent::getDefaultConflictResolution($variant);
+    }
+
+    /**
+     * Allows testing of grant status
+     *
+     * @param string $filename
+     * @param string $hash
+     */
+    public function isGranted($fileID)
+    {
+        return parent::isGranted($fileID);
     }
 
     protected function isSeekableStream($stream)
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -73,17 +73,19 @@ class FileShortcodeProvider implements ShortcodeHandler, Flushable
      */
     public static function handle_shortcode($arguments, $content, $parser, $shortcode, $extra = [])
     {
+        $allowSessionGrant = static::config()->allow_session_grant;
+
         /** @var CacheInterface $cache */
         $cache = static::getCache();
         $cacheKey = static::getCacheKey($arguments, $content);
 
         $item = $cache->get($cacheKey);
         if ($item) {
-            /** @var AssetStore $store */
-            $store = Injector::inst()->get(AssetStore::class);
             if (!empty($item['filename'])) {
-                $grant = static::config()->allow_session_grant;
-                $store->getAsURL($item['filename'], $item['hash'], null, $grant);
+                /** @var AssetStore $store */
+                $store = Injector::inst()->get(AssetStore::class);
+
+                $store->getAsURL($item['filename'], $item['hash'], null, $allowSessionGrant);
             }
             return $item['markup'];
         }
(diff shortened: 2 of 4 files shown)


src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -73,17 +73,19 @@ public static function get_shortcodes()
...
         if ($item) {
-            /** @var AssetStore $store */
-            $store = Injector::inst()->get(AssetStore::class);
             if (!empty($item['filename'])) {
-                $grant = static::config()->allow_session_grant;
-                $store->getAsURL($item['filename'], $item['hash'], null, $grant);
+                /** @var AssetStore $store */
+                $store = Injector::inst()->get(AssetStore::class);
+
+                $store->getAsURL($item['filename'], $item['hash'], null, $allowSessionGrant);
```
maxime-rainville, line 88:
```suggestion
                $store->grant($item['filename'], $item['hash']);
```

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -73,17 +73,19 @@ public static function get_shortcodes()
...
+
         /** @var CacheInterface $cache */
         $cache = static::getCache();
         $cacheKey = static::getCacheKey($arguments, $content);
 
         $item = $cache->get($cacheKey);
         if ($item) {
-            /** @var AssetStore $store */
-            $store = Injector::inst()->get(AssetStore::class);
             if (!empty($item['filename'])) {
```
maxime-rainville, line 84:
```suggestion
            if ($allowSessionGrant && !empty($item['filename'])) {
```

src/Shortcodes/FileShortcodeProvider.php
maxime-rainville: You can grant access to a user without needlessly generating a URL.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -52,9 +54,8 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
             /** @var AssetStore $store */
             $store = Injector::inst()->get(AssetStore::class);
             if (!empty($item['filename'])) {
-                $grant = static::config()->allow_session_grant;
                 // Initiate a protected asset grant if necessary
-                $store->getAsURL($item['filename'], $item['hash'], null, $grant);
+                $store->getAsURL($item['filename'], $item['hash'], null, $allowSessionGrant);
```
maxime-rainville, line 58:
```suggestion
                $store->grant($item['filename'], $item['hash']);
```

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -52,9 +54,8 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
             /** @var AssetStore $store */
             $store = Injector::inst()->get(AssetStore::class);
             if (!empty($item['filename'])) {
```
maxime-rainville, line 56:
```suggestion
            if ($allowSessionGrant && !empty($item['filename'])) {
```

Cheddam:
@maxime-rainville I've addressed your feedback and added some unit testing.

src/Dev/TestAssetStore.php
```diff
@@ -194,6 +204,65 @@ public function getDefaultConflictResolution($variant)
...
+        self::$grantedFiles[$member ? $member->ID : 0][$fileID] = false;
+    }
+
+    /**
+     * Fetches grants directly from TestAssetStore to avoid Session dependency in tests
+     *
+     * @param string $filename
+     * @param string $hash
+     */
+    public function isGranted($fileID)
```
maxime-rainville, line 241: Why not just call the protected method with the same name?

maxime-rainville, line 241: Why not just call the protected method with the same name?

Cheddam: You're totally right 🤦 I've dropped out the other overloaded methods and adjusted `isGranted()` to call its parent.

PR #411, opened 2020-07-19
Allow file variant with different extensions POC

maxime-rainville:
This is just a proof-of-concept for now. Matching RFC https://github.com/silverstripe/silverstripe-assets/issues/441

# Background

Our asset system allows you to define "variants" of a file. Variants are are variation of the parent file. e.g.: A thumbnail of an image, or a cropped version of the original image.

This is quite a useful feature and allows any operation on the main file to be reflected on its variants. e.g: Deleting, publishing, unpublishing a file will do the same thing to all its variant. If you are granted access to a file, you automatically get access to its variants.

Currently variants are tracked by creating file with a variation on the name of the original file. e.g.: If you have an image called "Upload/logo.png", all it's variant will be called something like "Upload/logo__someVariantstring.png"

# Problem

Currently, variants must have the same extensions as the parent file. This foreclose a lot of interesting scenarios and in-effect means that variants are only useful for resizing/cropping images.

If we could have variants with extension that differ from their parents we could open this kind of possibilities:
* provide thumbnails for more file types (e.g.: We could take thumbnails for videos or documents)
* convert files to alternative formats (e.g.: .docx document could be converted to PDF on the fly or we could convert a PNG to JPG)

We probably wouldn't want to provide all these features out of the box in core, but we could provide a simple API that third parties could hook into to provide conversion features.

# Approach

This PR builds on some of our previous work with `FileIDHelper` concept. FileIDHelpers are responsible for:
* converting file tuple (an array with a file name, file hash, variant) to their equivalent File ID (path as a string),
* parsing File IDs back to their original file tuple,
* matching a file variant back to its original file.

This PR updates `FileIDHelper` classes so they can understand a special `extRewrite` variant. This variant as two parameters: the extension of the parent file and the extension that the new variant should have. 

When building the fileID from a tuple, the extension will swap to the new extension. When parsing a file ID back to its original tuple, the extension in the filename key is restored to the extension of the parent file.

For example, this tuple would be converted to this file ID:
`['Filename' => 'video.mp4', 'Hash' => 'b67fcb469a', 'variant' => 'extRewriteWyJtcDQiLCJqcGciXQ']` becomes `b67fcb469a/video__extRewriteWyJtcDQiLCJqcGciXQ.jpg`. `WyJtcDQiLCJqcGciXQ` is a base64 encoded representation of `['mp4','jpg']`.

# What would the API for this look like

This is based on the existing logic for creating thumbnails and the `manipulateImage` method. We might be able to provide a nicer API to do this.

```php
public function ToImage()
{
    return $file->manipulateExtension(
        'jpg',
        function (AssetStore $store, string $filename, string $hash, string $variant) use ($file) {
            $results = $file->getImageBackend();
            $results->setImageResource(VideoConverter::magicalMethod($file->getStream()));
            $tuple = $results->writeToStore($store, $filename, $hash, $variant, ['conflict' =>     AssetStore::CONFLICT_USE_EXISTING]);
            return [$tuple, $results];
        }
    );
}
```

# Extra complications

Most of the logic around generating thumbnails in assets and asset-admin has some safe guard to make sure you don't try generating thumbnails for regular files. Generating images for regular files is kind of the entire point of this, so we'll need to rewrite some of those safe guards or remove them altogether.

We might want to have dedicated API for this purpose as well. This will require some more thoughts.

On a UX consideration, the asset-admin GalleryItem currently only has thumbnails for images. If we move into a direction where PDF, Videos and other types of files can have thumbnails as well, we'll need to have a way to differentiate those files from regular images

# Related PR
* https://github.com/silverstripe/silverstripe-asset-admin/pull/1133

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -1503,4 +1503,10 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     {
         return FileNameFilter::create();
     }
+
+
+    public function ToImage(): ?AssetContainer
+    {
+        return $this->File->ToImage();
+    }
 }
--- a/src/FilenameParsing/AlternativeFileExtensionTrait.php
+++ b/src/FilenameParsing/AlternativeFileExtensionTrait.php
@@ -4,46 +4,77 @@ namespace SilverStripe\Assets\FilenameParsing;
 
 use SilverStripe\Core\Convert;
 
+/**
+ * This trait can be applied to a `FileIDHelper` imlementation to add the baility to encode alternative file extension.
+ *
+ * This allows the creation of file variant with different extension to the original file.
+ */
 trait AlternativeFileExtensionTrait
 {
 
+    /**
+     * Rewrite a filename with a different extension
+     * @param string $filename
+     * @param string $variant
+     * @return string
+     */
     public function rewriteVariantExtension(string $filename, string $variant): string
     {
         return $this->swapExtension($filename, $variant, 1);
     }
 
+    /**
+     * Rewrite a filename to use the original extension for the provided variant.
+     * @param string $filename
+     * @param string $variant
+     * @return string
+     */
     public function restoreOriginalExtension(string $filename, string $variant): string
     {
         return $this->swapExtension($filename, $variant, 0);
     }
 
+    /**
+     * Construct the original or alternative filname extension for the given filename and variant string.
+     * @param string $filename Original filename without variant
+     * @param string $variant Full variant list
+     * @param int $extIndex Wether we want the original extension (0) or the new extension (1)
+     * @return string
+     */
     private function swapExtension(
         string $filename,
         string $variant,
         int $extIndex
-    ): string
-    {
+    ): string {
+        // If there's no variant at all, we can rewrite the filenmane
         if (empty($variant)) {
             return $filename;
         }
 
+        // Split variant string in variant list
         $subVariants = explode('_', $variant);
 
+        // Split our filename into a filename and extension part
         if (!preg_match('/(.+)\.([a-z\d]+)$/i', $filename, $matches)) {
             return $filename;
         }
-
         [$_, $filenameWitoutExtension, $extension] = $matches;
 
+        // Loop our variant list until we find our special file extension swap variant
         foreach ($subVariants as $subVariant) {
-            if (preg_match("/^extRewrite(.+)$/", $subVariant, $matches)) {
+            $extSwapVariant = FileIDHelper::EXTENSION_REWRITE_VARIANT;
+            if (preg_match("/^$extSwapVariant(.+)$/", $subVariant, $matches)) {
                 [$_, $base64] = $matches;
+
+                /**
+                 * This array always contain 2 values: The orignial extension and the new extension
+                 * @var array $extensionData
+                 */
                 $extensionData = Convert::base64url_decode($base64);
                 $extension = $extensionData[$extIndex];
             }
         }
 
         return $filenameWitoutExtension . '.' . $extension;
-
     }
 }
--- a/src/FilenameParsing/AlternativeFileExtensionTrait.php
+++ b/src/FilenameParsing/AlternativeFileExtensionTrait.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+use SilverStripe\Core\Convert;
+
+trait AlternativeFileExtensionTrait
+{
+
+    public function rewriteVariantExtension(string $filename, string $variant): string
+    {
+        return $this->swapExtension($filename, $variant, 1);
+    }
+
+    public function restoreOriginalExtension(string $filename, string $variant): string
+    {
+        return $this->swapExtension($filename, $variant, 0);
+    }
+
+    private function swapExtension(
+        string $filename,
+        string $variant,
+        int $extIndex
+    ): string
+    {
+        if (empty($variant)) {
+            return $filename;
+        }
+
+        $subVariants = explode('_', $variant);
+
+        if (!preg_match('/(.+)\.([a-z\d]+)$/i', $filename, $matches)) {
+            return $filename;
+        }
+
+        [$_, $filenameWitoutExtension, $extension] = $matches;
+
+        foreach ($subVariants as $subVariant) {
+            if (preg_match("/^extRewrite(.+)$/", $subVariant, $matches)) {
+                [$_, $base64] = $matches;
+                $extensionData = Convert::base64url_decode($base64);
+                $extension = $extensionData[$extIndex];
+            }
+        }
+
+        return $filenameWitoutExtension . '.' . $extension;
+
+    }
+}
--- a/src/Image.php
+++ b/src/Image.php
@@ -2,6 +2,9 @@
 
 namespace SilverStripe\Assets;
 
+use SilverStripe\Assets\Storage\AssetContainer;
+use SilverStripe\Assets\Storage\DBFile;
+
 /**
  * Represents an Image
  */
@@ -59,5 +62,16 @@ class Image extends File
         }
         $this->extend('updatePreviewLink', $link, $action);
         return $link;
+    }
+
+    public function ToImage(): ?AssetContainer
+    {
+        $image = parent::ToImage();
+
+        if (empty($image)) {
+            $image = $this->File;
+        }
+
+        return $image;
     }
 }
--- a/src/ImageBackendFactory.php
+++ b/src/ImageBackendFactory.php
@@ -51,7 +51,7 @@ class ImageBackendFactory implements Factory
 
         // Verify file exists before creating backend
         $backend = null;
-        if ($store->exists() && $store->getIsImage()) {
+        if ($store->exists()) {
             $backend = $this->creator->create($service, $params);
         }
 
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -731,7 +731,7 @@ trait ImageManipulation
         }
 
         // Skip files we know won't be an image
-        if (!$this->getIsImage() || !$this->getHash()) {
+        if (!$this->getHash()) {
             return null;
         }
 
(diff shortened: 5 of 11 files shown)


clarkepaul:
@maxime-rainville looking great! are there other variations of icons we would require like doc, pdf etc. ? would be good to have a list so we could ensure iconography would still make sense for both thumb and list view.

maxime-rainville:
As far as I know, these are all the file categories for which we have icons/styles:
- Video
- Audio
- Document (e.g.: DOCX, DOC, PDF, XSLX, XSL, CSV)
- Any other file
- Image
- Archives (e.g.: ZIP, TAR.GZ, 7ZIP)

So we don't currently differentiate between a DOC and a PDF file ... or between a PNG vs a JPG.

So if we want to allow any file type to have a thumbnail, then maybe we need have some generic logic that overlays some file type icon over the thumbnail ... if the thumbnail is provided.

image

sachajudd:
I've created a PDF thumbnail icon to differentiate between a DOC extension icon. This will hopefully help with more clarity about extensions. Note this will need to be exported by myself to a PNG for use in admin as I'm pretty sure this is how they are implemented currently. See [File icons (PDF)](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f6d4dc5152a65699af63443/tab/design?mode=preview) for design sizes in 32 and 92.

I've also had a little think about where we could add an extension type more clearly in the CMS. I was thinking potentially a readonly field with the type but this feels like adding clutter. The original "Filename" does usually include the extension but this can be removed and easy to loose this information. As a quick win I think we could add the extension to the end of the editor specs (Size of image, size of file, **extension type**) e.g 10x10px, 2MB, JPG. Let me know your thoughts, see [Edit file / Details tab](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f152068f6772b33db2e52b0/tab/design?mode=preview) for design.

tractorcow:
I wrote a module that lets you transcode images in templates.

https://github.com/tractorcow/silverstripe-image-formatter

It works but it breaks a lot of core filesystem in order to do so;
 - images are forced to be public
 - hack to disable hash-comparison for file->exists()
 

What you propose here @maxime-rainville would be much nicer :)

maxime-rainville:
I've got 4 people who have expressed that this feature would be helpful in the real world:
- @DorsetDigital on the [forum](https://forum.silverstripe.org/t/thumbnails-for-every-file-types-and-convert-files-between-different-formats/3564)
- @tractorcow on this PR
- @scott-nz and @blueo on the [office slack](https://silverstripeltd.slack.com/archives/CB95TE3MF/p1605651627074600) (Silverstripe LTD only)

I'll spend some time cleaning up the PR and adding a bit of doc. Then see if we can get a bit more more momentum to get this over the line.

I'm thinking we can probably just get the `assets` bit merged initially. If this works well and we don't find any major bugs, we can  have a  look at enabling the `asset-admin` part.

src/File.php
```diff
@@ -1503,4 +1503,10 @@ protected function getFilter()
     {
         return FileNameFilter::create();
     }
+
+
+    public function ToImage(): ?AssetContainer
```
maxime-rainville, line 1508: Strictly speaking, this method is not needed to introduce the ability to create variants with other extension. If we want to minimise the risk, we could ship the PR without it and still ship something useful.

So you could still do something like `$file->ToJPG()`. However, you couldn't perform image manipulation on the result.

src/Image.php
```diff
@@ -60,4 +63,15 @@ public function PreviewLink($action = null)
         $this->extend('updatePreviewLink', $link, $action);
         return $link;
     }
+
+    public function ToImage(): ?AssetContainer
```
maxime-rainville, line 67: See https://github.com/silverstripe/silverstripe-assets/pull/411#discussion_r609476073

src/ImageBackendFactory.php
```diff
@@ -51,7 +51,7 @@ public function create($service, array $params = [])
 
         // Verify file exists before creating backend
         $backend = null;
-        if ($store->exists() && $store->getIsImage()) {
+        if ($store->exists()) {
```
maxime-rainville, line 54: See https://github.com/silverstripe/silverstripe-assets/pull/411#discussion_r609476073

src/ImageManipulation.php
```diff
@@ -731,7 +731,7 @@ public function getImageBackend()
         }
 
         // Skip files we know won't be an image
-        if (!$this->getIsImage() || !$this->getHash()) {
+        if (!$this->getHash()) {
```
maxime-rainville, line 735: See https://github.com/silverstripe/silverstripe-assets/pull/411#discussion_r609476073

src/ImageManipulation.php
```diff
@@ -853,10 +853,19 @@ function (AssetStore $store, $filename, $hash, $variant) use ($callback) {
                 $backend = $this->getImageBackend();
 
                 // If backend isn't available
-                if (!$backend || !$backend->getImageResource()) {
+                if (!$backend) {
                     return null;
                 }
 
+                // If we don't have an image resource
```
maxime-rainville, line 861: See https://github.com/silverstripe/silverstripe-assets/pull/411#discussion_r609476073

src/Storage/DBFile.php
```diff
@@ -628,4 +628,12 @@ public function copyFile($newName)
             ->getStore()
             ->copy($this->Filename, $this->Hash, $newName);
     }
+
+    public function ToImage(): ?AssetContainer
```
maxime-rainville, line 632: See https://github.com/silverstripe/silverstripe-assets/pull/411#discussion_r609476073

src/FilenameParsing/AlternativeFileExtensionTrait.php
```diff
@@ -0,0 +1,80 @@
...
+namespace SilverStripe\Assets\FilenameParsing;
+
+use SilverStripe\Core\Convert;
+
+/**
+ * This trait can be applied to a `FileIDHelper` imlementation to add the baility to encode alternative file extension.
+ *
+ * This allows the creation of file variant with different extension to the original file.
+ */
+trait AlternativeFileExtensionTrait
```
maxime-rainville, line 12: I'm thinking maybe this should be some sort of service instead. There's not really any value in making that logic snipet part of the FIleIDHelper.

Also, there's bit of logic in `ImageManipulate::manipulateExtension()` and `InterventionBackend` that might make more sense in a centralised service.

src/ImageManipulation.php
```diff
@@ -984,6 +994,18 @@ public function manipulate($variant, $callback)
...
 
+    /**
+     * @param string $newExtension
+     * @param callable $callback
+     * @return DBFile The manipulated file
+     */
+    public function manipulateExtension(string $newExtension, callable $callback)
+    {
+        $pathParts = pathinfo($this->getFilename());
+        $variant = $this->variantName(FileIDHelper::EXTENSION_REWRITE_VARIANT, $pathParts['extension'], $newExtension);
```
maxime-rainville, line 1005: Maybe this bit should be in a centralised service

src/InterventionBackend.php
```diff
@@ -364,7 +365,9 @@ public function writeToStore(AssetStore $assetStore, $filename, $hash = null, $v
             }
 
             // Save file
-            $extension = pathinfo($filename, PATHINFO_EXTENSION);
+            $url = $assetStore->getAsURL($filename, $hash, $variant, false);
+            $extension = pathinfo($url, PATHINFO_EXTENSION);
```
maxime-rainville, line 369: Maybe this bit should be in a centralised service.

maxime-rainville:
Just looking at this again after a while, a couple question pop up:
- Maybe `AlternativeFileExtensionTrait` should be some sort of service instead
- Do we want to ship the tweaks to allow Image Manipulation of variants right away? This enable things like convert a PDF to an image and then resize that image.
- There's some edge case where you chain up file conversion which I'm not sure how to handle e.g.:
  - Maybe you start with a XLS file, then get it converted to a CSV, then get it converted to a TXT file
  - Or maybe you have a JPG that you convert to a PNG that you then convert back to a JPG
- Right now the way you create a new converters is you write an a new DataExtension for Image/File. That extension has a method that calls `manipulateExtension` with a callback. Maybe we could be more prescriptive than that: define a `FileConverter` interface. Then you register your new converter somewhere. Then, Image/File could have a `convert` method that looks for a suitable converter for your use case.

tractorcow:
question, wouldn't the extRewrite variant only need to include the "from" variant? (since the "to" variant is the actual current extension). This would be simpler in terms of url length :)

The only thing that you need is the ability to generate the "original" file ID from any current tuple.

maxime-rainville:
@tractorcow Possibly ... I remember pondering this question when I started working on this. From memory, it made the logic more difficult to follow because you had to decode the `extRewrite` variant to know if you were working with the original file name or the the one with the rewritten extension.

I don't recall spending hours trying to get it working, so it's probably just a shortcut I took to try to ship something in one hackday. Probably worth having a second though about it.

kinglozzer:
https://github.com/silverstripe/silverstripe-assets/issues/486 just reminded me of this PR. This API would definitely be valuable in the context of webp images, because most content authors we’ve mentioned webp to have absolutely no idea the format exists, or how to create images for it. The ability to upload JPEGs (and other formats, but our content authors typically deal with JPEGs) and have the CMS handle conversion for you would be awesome.

nathanbrauer:
> I've got 4 people who have expressed that this feature would be helpful in the real world:
> 
> * @DorsetDigital on the [forum](https://forum.silverstripe.org/t/thumbnails-for-every-file-types-and-convert-files-between-different-formats/3564)
> * @tractorcow on this PR
> * @scott-nz and @blueo on the [office slack](https://silverstripeltd.slack.com/archives/CB95TE3MF/p1605651627074600) (Silverstripe LTD only)

You can add me and @Marketera to the list.  We have several projects/clients that need similar functionality. Especially generating thumbnails for video assets.

Attn: @muskie9

Reference: https://forum.silverstripe.org/t/thumbnails-for-every-file-types-and-convert-files-between-different-formats/3564

GuySartorelli:
Closed in favour of https://github.com/silverstripe/silverstripe-assets/pull/585

Issue #414, opened 2020-07-28
Files can't be published, recursive listContents causes out of memory


adrhumphreys:
We found that when trying to publish an image (DataObject) that had it's file replaced the request would timeout and the file would not be replaced. The image (DataObject) would not be updated with the new file.

We were able to trace through the code to find where this was occurring. When you publish a new file, first a call is made to `AssetControlExtension::deleteAll` to delete the existing variants of that file. This loops over all assets passed to it and calls `delete` on the `AssetStore`.

The `AssetStore` is `FlysystemAssetStore`. In the delete code it calls `deleteFromFileStore` to delete the files from the file system. The strategy which is used is the `SilverStripe\Assets\FilenameParsing\FileIDHelperResolutionStrategy` and `findVariants` is called on it.

In `FileIDHelperResolutionStrategy::findVariants` it first gets the helpers that are resolvable and then loops through those helpers to try and find the variants. At the begging of these helpers it first calls `$filesystem->listContents($folder, true)`. This translates to `League\Flysystem\Adapter\Local::listContents`.

This function will then recursively list the directories and files based on the directory that is passed through to it. While looping through that directory it will get the meta information for each file.

Usually at this point you will slowly be building a large array and get an exception like:
```
Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 20480 bytes) in /var/www/mysite/releases/046d2d2565825741941ddfa598701a937f48dafd/vendor/league/flysystem/src/Adapter/Local.php on line 508
```

From my understanding we shouldn't be [recursively searching](https://github.com/silverstripe/silverstripe-assets/blob/3d842a160163366456ba504dba00920d1f10fa20/src/FilenameParsing/FileIDHelperResolutionStrategy.php#L439) for variants and should be calling `$filesystem->listContents($folder, false)` to only get the files that are in the current directory the asset is stored in. Otherwise any asset that is in the root asset directory will on publish need the file system to loop through every asset that exists to compare if it is a variant of the uploaded asset or not

## Recreating the issue in a new install:
- First we create a new project from the recipe:  
  `composer create-project silverstripe/recipe-cms ./asset `
- Then login to that new site and add an asset to the file system at the top level
- Save and publish that file
- In the terminal go into the assets directory:
  `cd asset/public/assets`
- Run `composer create-project silverstripe/recipe-cms ./example` to create a large amount of files in the folder
[...]

PR #415, opened 2020-07-28
414: listContents no longer called recursively

adrhumphreys:
This assumes that variants for files are always in the same directory as the asset itself which will probably require some validation on what happens when you move a file around with its variations etc

Hopefully resolves: https://github.com/silverstripe/silverstripe-assets/issues/414

### DIFF
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -53,4 +53,11 @@ interface FileIDHelper
      * @return string
      */
     public function lookForVariantIn(ParsedFileID $parsedFileID);
+
+    /**
+     * Specify if this File ID Helper stores variants in subfolders and require a recursive look up to find all
+     * variants.
+     * @return bool
+     */
+    public function lookForVariantRecursive(): bool;
 }
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -436,7 +436,7 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
 
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariants = $filesystem->listContents($folder, true);
+            $possibleVariants = $filesystem->listContents($folder, false);
 
             // Flysystem returns array of meta data abouch each file, we remove directories and map it down to the path
             $possibleVariants = array_filter($possibleVariants, function ($possibleVariant) {
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -436,7 +436,7 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
 
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariants = $filesystem->listContents($folder, false);
+            $possibleVariants = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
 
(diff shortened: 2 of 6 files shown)


src/FilenameParsing/FileIDHelper.php
```diff
@@ -53,4 +53,11 @@ public function isVariantOf($fileID, ParsedFileID $parsedFileID);
      * @return string
      */
     public function lookForVariantIn(ParsedFileID $parsedFileID);
+
+    /**
+     * Specify if this File ID Helper stores variants in subfolders and require a recursive look up to find all
+     * variants.
+     * @return bool
+     */
+    public function lookForVariantRecursive(): bool;
```
maxime-rainville, line 62: Technically this is a breaking API change because a class implementing this interface will get an error if it doesn't implement this method. However this interface is marked as `internal` so we're allowed do to breaking changes to it.

emteknetnz, line 62: What's the existing unit test coverage for this like?

emteknetnz: In practical terms I wouldn't expect there to be any custom implementations of this out in the wild, so this is probably OK

maxime-rainville: `lookForVariantIn` and the logic for finding variants has pretty solid coverage. 

`lookForVariantRecursive` is a new method so it won't have any coverage. I can add some unit test of it, but the expectation is that each file ID helper will always return the same value for this method.

emteknetnz: Could do it non-breaking by just adding this function to the classes themselves instead of the interface and doing a method_exists() check before calling, and if it doesn't exist, default to true

maxime-rainville: We could, but I expect that at some stage we'll make this interface part of our official API. I would rather have a nice clean API when we do that. The entire point of marking these classes as internal was that we could do something like this.

emteknetnz: Sweet, I'm going to presume `lookForVariantIn` means that all the variants are in sane places e.g. hash and natural only have variants in the same folder as the main file

It's almost pointless adding unit tests to `lookForVariantRecursive` except to make the the codecov level nicer.  Probably should just for the sake of completionism, though I'm honestly not fussed

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -436,7 +436,7 @@ public function findVariants($tuple, Filesystem $filesystem)
 
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariants = $filesystem->listContents($folder, true);
+            $possibleVariants = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
```
emteknetnz, line 439: Prevent it from being a potentially breaking change
```
$recursive = method_exists($helper, 'lookForVariantRecursive') ? $helper->lookForVariantRecursive() : true;
$possibleVariants = $filesystem->listContents($folder, $recursive);
```

maxime-rainville: Even if we do this change, PHP will throw a fit if you have a class that implement the `FileIDHelper` interface without the `lookForVariantRecursive` method.

I would rather keep the API for FileIDHelper clean and concise ... even if it causes an internal API breakage.

src/FilenameParsing/FileIDHelper.php
emteknetnz: Oh right gotcha, sorry I didn't realise all these classes were marked internal.

emteknetnz:
Ran this through travis cwp-recipe-kitchen-sink - is green
https://travis-ci.org/github/creative-commoners/cwp-recipe-kitchen-sink

Tested locally, with protected + public images in root folder, sub-folder and sub-sub-sub-folder.

PR #420, opened 2020-09-09
Merge up 1.5 into 1.6

maxime-rainville:


.travis.yml
```diff
@@ -30,15 +30,8 @@ matrix:
...
-
-    # Warning: phpunit with --coverage-clover takes too much time, so travis build fails after 10
-    # minutes with no output from the task.
-    # We have to disable it until we come up with another (faster) solution to gather coverage
-    #
-    # See https://github.com/silverstripe/silverstripe-assets/pull/407
-    #
-    # - php: 7.3
-    #   env: DB=MYSQL RECIPE_VERSION=4.6.x-dev PHPUNIT_COVERAGE_TEST=1
+    - php: 7.3
```
dnsl48, line 33: we already have `php: 7.3` above in the matrix. Might be worth adding coverage there instead

dnsl48 (changes requested):

.travis.yml
maxime-rainville: Historically, we tend to have code coverage on its own matrix entry ... although there's no specific reason for doing this.

dnsl48: true. ok then, let's do it

dnsl48 (approved):

PR #424, opened 2020-09-30
NEW Introduce PHP 8 and prefer-lowest builds

Cheddam:
Notes:

- Imagick has PHP 8 support on their development branch, but this hasn't shipped yet, so I've disabled it for now. I can confirm the relevant tests pass in my local environment running the development build of Imagick.
- ~The build is failing due to its use of Prophecy, so it's in a [similar conundrum to silverstripe/config](https://github.com/silverstripe/silverstripe-config/pull/46#issuecomment-698574874). I'm now leaning towards @sminnee's suggestion of refactoring away from Prophecy.~ I've successfully refactored the affected tests into native PHPUnit mocks.
- The prefer-lowest build was broken by a since-fixed issue in Flysystem. I've added Flysystem as a direct dependency and bumped the minimum version in the process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
   fast_finish: true
   include:
     - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPUNIT_TEST=1 PHPCS_TEST=1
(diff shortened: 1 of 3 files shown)


sminnee:
In other repos I've added some require-dev dependencies to fix the prefer-lowest build. 4.0 wasn't accurate in its minimum requirement for a few packages, thanks mainly to the lack of a prefer lowest build in the past ;-)

Bumping the minimum framework requirement works too, but it does narrow the combination of packages that could otherwise be legitimately installed together.

robbieaverill (approved):
LGTM

PR #427, opened 2020-11-30
MNT Travis shared config

emteknetnz:
The codecov isn't working properly, however we may need to just keep it this way for the time being

I added PHPUNIT_COVERAGE_SUITE to resolve the issue where
a) it would normally time out after 10 minutes
b) the old solution of using `then pecl install pcov ; composer require pcov/clobber --dev ; vendor/bin/pcov clobber` was causing a segmentation fault

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,60 +1,15 @@
-language: php
+version: ~> 1.0
 
-dist: xenial
-
-services:
-  - mysql
-  - postgresql
-  - xvfb
-
-apt:
-  packages:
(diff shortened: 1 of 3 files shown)


Cheddam:
Agree that we can address the coverage issue separately - hopefully Xdebug 3 will make things smoother, but it's out of scope for this change. Raised [product-issues#316](https://github.com/silverstripeltd/product-issues/issues/316) to tackle this internally in the new year.

Issue #429, opened 2021-01-28
Filesystem::removeFolder will skip folders whose names evaluate to a falsy value


blueo:
SilverStripe\Assets\Filesytem::removeFolder attempts to recursively delete everything in folder. It does this recursively via a while loop:

```php
public static function removeFolder($folder, $contentsOnly = false)
    {

        // remove a file encountered by a recursive call.
        if (is_file($folder) || is_link($folder)) {
            unlink($folder);
        } else {
            $dir = opendir($folder);
            while ($file = readdir($dir)) {
                if (($file == '.' || $file == '..')) {
                    continue;
                } else {
                    self::removeFolder($folder . '/' . $file);
                }
            }
            closedir($dir);
            if (!$contentsOnly) {
                rmdir($folder);
            }
        }
    }
```

We ran into a situation where `$file = readdir($dir)` was `"0"` (zero). This evaluates to false and terminates the loop. Consequently, when `rmdir` is called it fails as the parent directory is not empty. 

This function might be a bit outdated with the introduction of flysystem - we switched to using the flysystem methods (`deleteDir`) and it works fine.

PR #430, opened 2021-01-28
429: Ensure remove folder removes all exepcted files/folders/links

adrhumphreys:
In some instances the where the directory could be `0` or `false` then the while loop would exit due to the function `readdir` returning the entry name on success or false on failure.

This new method will ensure that if passed a folder it recursively deletes the directory and throws an exception if the item is not readable to ensure we inform users that something wrong has occurred.

I've also added test cases for the scenarios to ensure they all work as expected, the second falsie case was failing with the previous code.

This should fix: #429

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "silverstripe/versioned": "^1@dev",
         "sminnee/phpunit-mock-objects": "^3.4.5",
         "sminnee/phpunit": "^5.7.29",
-        "squizlabs/php_codesniffer": "^3"
+        "squizlabs/php_codesniffer": "^3",
+        "mikey179/vfsstream": "^1.6"
     },
     "suggest": {
         "ext-exif": "If you use GD backend (the default) you may want to have EXIF extension installed to elude some tricky issues"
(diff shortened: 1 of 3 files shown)


maxime-rainville (commented):
I'm thinking maybe we should just require `symfony/filesystem` (which is already installed indirectly by framework) and wrap our call around that lib.

adrhumphreys:
@maxime-rainville I've updated it to use the `symfony/filesystem`. Not 100% sold on it just because it needs to recreate the folder to maintain support with the `contentsOnly` flag which means it's changing functionality a little bit and could mess up permissions

maxime-rainville:
You could adjust the call a tiny bit so it deletes each individual item one at a time when the `contentOnly` lag is provided.

In the long term, my instinct would be to rewrite the entire class to use symphony. So this PR could be a start.

Or even better, deprecate the entire class and tell people to use symphony directly.

FYI, we'll need to explicitly require symphony in the `composer.json` file

adrhumphreys:
Sweet updated to require symfony and also support contents only

maxime-rainville (approved):
@adrhumphreys Looks good.

After having a second look at the tests, I decided I had some beef with all the manual folder and file creation. I rewrote your tests to use a virtual file system abstraction, which I think makes the them more readable.

Let me know if you're OK wit the change and we can merge.

PR #431, opened 2021-01-28
Check if Versioned is available

### LINKED ISSUE #422: Exception in `Folder::find_or_make()` if Versioned not installed
As of silverstripe/assets v1.6.1 it is now impossible to `Folder::find_or_make()` if Versioned is not installed, caused by:
```
$versioned = Injector::inst()->get('SilverStripe\Versioned\Versioned');
```
Injector returns an `InjectorNotFoundException()`.

My guess is that this part of the code should potentially have a wrapping `if (class_exists(Versioned::class))...`.

axllent:
Allow assets to work without `Versioned::class`. 

Without this check assets errors out with `Uncaught SilverStripe\Core\Injector\InjectorNotFoundException: ReflectionException: Class SilverStripe\Versioned\Versioned does not exist in....`.

Resolves #422

### DIFF
--- a/src/Folder.php
+++ b/src/Folder.php
@@ -71,12 +71,14 @@ class Folder extends File
 
             // When in archived mode, find or make should not find folders that don't exist
             // We check explicitly for Versioned and if it exists then we'll confirm the reading mode isn't archived
-            $versioned = Injector::inst()->get('SilverStripe\Versioned\Versioned');
-            if ($versioned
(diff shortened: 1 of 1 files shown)


emteknetnz (approved):
Seems legit, technically Versioned is only a [dev dependency as at 1.7.0](https://github.com/silverstripe/silverstripe-assets/blob/68a6a820a5f99b23f784903e135655743c69c483/composer.json#L28)

PR #432, opened 2021-01-28
#426: Ensure cache key doesn't contain invalid characters

### LINKED ISSUE #426: InvalidArgumentException thrown by File::setFromLocalFile()
## Problem
When I tried to use setFromLocalFile() to import a file to the SilverStripe database, an Symfony\Component\Cache\Exception\InvalidArgumentException is thrown.

The reason for this is the method buildCacheKey() in the class Sha1FileHashingService, which returns a base64 string. According to PSR-6 a cache key should not contain the characters "{}()/\@:.", as they are reserved.

A base64 string can contain a forward slash character, which produces the named exception, because it is not a valid character in a cache key.

## Steps to reproduce
Try to save a file using File::setFromLocalFile(), which leads to a '/' in the base64 string. See screenshot.

## Environment
PHP 7.3, Assets-Module: 1.6.1, OS: Debian 9, Webserver: Apache2 2.4

adrhumphreys:
The cache key when converted to base64 can contain `/` which is an invalid character for the symphony [CacheItem](https://github.com/symfony/cache/blob/a7a14c4832760bd1fbd31be2859ffedc9b6ff813/CacheItem.php#L166)

The fix is to just replace that one character with the letter `a`, not 100% sure if that's appropriate or we should use some method to decide on a specific character?

This should fix #426

### DIFF
--- a/src/Storage/Sha1FileHashingService.php
+++ b/src/Storage/Sha1FileHashingService.php
@@ -178,7 +178,11 @@ class Sha1FileHashingService implements FileHashingService, Flushable
     private function buildCacheKey($fileID, $fs)
     {
         $fsID = $this->getFilesystemKey($fs);
-        return base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        $cacheKey = base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        // base64_encode can contain `/` , but that is not a valid character in cache keys
+        // @see CacheItem::validateKey. We therefore replace it with the url encoded equivalent
+        // from https://tools.ietf.org/html/rfc4648#page-8 which is `_`
+        return strtr($cacheKey, ['/' => '_']);
     }
 
     /**
--- a/tests/php/Storage/Sha1FileHashingServiceTest.php
+++ b/tests/php/Storage/Sha1FileHashingServiceTest.php
@@ -6,6 +6,7 @@ use Exception;
 use InvalidArgumentException;
 use League\Flysystem\FileNotFoundException;
 use League\Flysystem\Filesystem;
+use ReflectionMethod;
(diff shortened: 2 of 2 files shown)


src/Storage/Sha1FileHashingService.php
```diff
@@ -178,7 +178,11 @@ private function getCache()
     private function buildCacheKey($fileID, $fs)
     {
         $fsID = $this->getFilesystemKey($fs);
-        return base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        $cacheKey = base64_encode(sprintf('%s://%s', $fsID, $fileID));
+        // base64_encode can contain `/` , but that is not a valid character in cache keys
+        // @see CacheItem::validateKey. Therefore we translate it to `a` (effectively a random character)
+        // without actually being random since the key needs to be consistent
+        return strtr($cacheKey, ['/' => 'a']);
```
kinglozzer, line 185: Should we just preg_replace all the reserved characters `{}()/\@:` while we’re here?

adrhumphreys: From what I can see `base64_encode` will use the characters outlined in [rfc4648](https://tools.ietf.org/html/rfc4648#page-6) of which are two outside of alpha-numeric values: `+` and `/`, and `+` is not forbidden.

Happy to err on the side of caution though and add them all in, did we want to assign each character a letter or just replace all instances with the same character?

dhensby: Why not use the base64 URL encoding standard rather than replacing / with an a (which has a base64 meaning)

adrhumphreys: Oh good point, I should've scrolled further down in the RFC, updated it to use `_`

emteknetnz (approved):

Issue #290, opened 2019-06-11
Enhancement: Improve InterventionBackend error logging


kinglozzer:
Currently it’s _really_ hard to track down the cause of issues when resampling images, exceptions are swallowed and replaced with a generic “unknown” message which is stored in a (non-human-readable) cache: https://github.com/silverstripe/silverstripe-assets/blob/1.4.1/src/InterventionBackend.php#L270-L278

I eventually managed to track down my issue by hacking in some logging on live (😎). Example of exception that was being swallowed:

```
Unable to read image from path (<path>/silverstripe-cache/www-data/interventionimage_QoP2Kw.jpg).
at <path>/vendor/intervention/image/src/Intervention/Image/Imagick/Decoder.php:26,
ImagickException(code: 445): cache resources exhausted
'<path>/silverstripe-cache/www-data/interventionimage_QoP2Kw.jpg'
@ error/cache.c/OpenPixelCache/3984 at
<path>/vendor/intervention/image/src/Intervention/Image/Imagick
```

I’m not sure how best to approach this, silently swallowing exceptions is a good default but it should be possible to opt in to exposing them. Perhaps something like:

```php
if (Injector::inst()->has(LoggerInterface::class.'.InterventionBackend')) {
    Injector::inst()->get(LoggerInterface::class.'.InterventionBackend')->error($e->getMessage());
}
```

sminnee:
Recategorising as medium priority bug as "eats errors" feels like a bug and we've seen this cause some headaches. CC @indygriffiths

kinglozzer:
FWIW, this is what I’ve been using to help us debug issues in production: https://gist.githubusercontent.com/kinglozzer/a9334ae31a7e544f4ba397888503fcb1/raw/6e22099c6f40ec4d7f262279572bb96b13bd44d8/patch.patch

That patch is opt-in however, so perhaps we should tweak it to something like:

```php
if (Injector::inst()->has(LoggerInterface::class.'.InterventionBackend')) {
    Injector::inst()->get(LoggerInterface::class.'.InterventionBackend')->error($e->getMessage(), ['exception' => $ex]);
} elseif (Director::isDev()) {
    Injector::inst()->get(LoggerInterface::class)->error($e->getMessage(), ['exception' => $ex]);
    // or just...
    throw $e;
}
```

PR #433, opened 2021-01-28
290: Enable logging for intervention exceptions

adrhumphreys:
This adds in some logging for the InterventionBackend to help aid when looking at issues with assets
This comes from the enhancement: https://github.com/silverstripe/silverstripe-assets/issues/290

If this gets approved, I can then look to getting a PR added to the docs to document that this exists

### DIFF
--- a/src/InterventionBackend.php
+++ b/src/InterventionBackend.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Assets;
 
 use BadMethodCallException;
+use Exception;
 use Intervention\Image\Constraint;
 use Intervention\Image\Exception\NotReadableException;
 use Intervention\Image\Exception\NotSupportedException;
@@ -13,9 +14,11 @@ use Intervention\Image\Size;
 use InvalidArgumentException;
 use LogicException;
 use Psr\Http\Message\StreamInterface;
+use Psr\Log\LoggerInterface;
 use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\Storage\AssetContainer;
 use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Control\Director;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
@@ -289,6 +292,8 @@ class InterventionBackend implements Image_Backend, Flushable
             // Handle unsupported image encoding on load (will be marked as failed)
             // Unsupported exceptions are handled without being raised as exceptions
             $error = self::FAILED_INVALID;
+            // Attempt to log the exception
+            self::logException($ex);
         } finally {
             if ($error) {
                 $this->markFailed($hash, $variant, $error);
@@ -324,6 +329,8 @@ class InterventionBackend implements Image_Backend, Flushable
             // Handle unsupported image encoding on load (will be marked as failed)
             // Unsupported exceptions are handled without being raised as exceptions
             $error = self::FAILED_INVALID;
+            // Attempt to log the exception
+            self::logException($ex);
         } finally {
             if ($error) {
                 $this->markFailed($hash, null, $error);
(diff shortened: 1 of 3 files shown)


emteknetnz:
This seems generally seems like a good idea, though the way the logger is being created looks a bit different from how it's normally done in the codebase

[Random example](https://github.com/silverstripe/silverstripe-assets/blob/1/src/Dev/Tasks/NormaliseAccessMigrationHelper.php#L38)

There's probably no need to have a logger fallback i.e. `$hasLogger || Director::isDev()` - simply have a default logger that's always there which can be swapped via yml if that makes sense for the project

Also the newly added unit test is failing in travis

emteknetnz (changes requested):
This seems generally seems like a good idea, though the way the logger is being created looks a bit different from how it's normally done in the codebase

Random example

There's probably no need to have a logger fallback i.e. $hasLogger || Director::isDev() - simply have a default logger that's always there which can be swapped via yml if that makes sense for the project

Also the newly added unit test is failing in travis

adrhumphreys:
@emteknetnz good point regarding the logging, I've added some checks in though as I don't want this to run in production without explicitly being enabled as it has the potential to eat up a bunch of disk space

src/InterventionBackend.php
```diff
@@ -858,4 +880,24 @@ protected function isStreamReadable($stream)
...
+     * If a file operation is erroring we can quickly eat up disk space with these exceptions.
+     * Therefore we only want to log these errors in production if the logging is enabled.
+     *
+     * We always log when in dev mode to try and alert developers to errors that can exist
+     *
+     * @param Exception $exception
+     */
+    private function logException(Exception $exception)
+    {
+        $isEnabled = Environment::getEnv('InterventionBackend_EnableExceptionLogging')
```
kinglozzer, line 894: I don’t think we need to introduce a new environment variable for this, it could be handled in config (forgive the pseudo-code):

```
if Injector::hasService(Logger.interventionLogger)
    interventionLogger->error()
else
    nullLogger->error()
```

Then the default YAML config could be:

```yml
---
Only:
  environment: dev
---
Injector:
  Logger.interventionLogger:
    logConfigEtc...
```

So on test/live it just hits the null logger and on dev it hits an actual error logger, then projects can opt into logging on test/live with YAML. What do you think?

michalkleiner:
Hi @adrhumphreys, do you mind sharing why you're closing all your issues, removing PRs? It's a great addition to the community even when it's just about describing something not working, or when you took time to create a PR. Is there something we can help with to avoid closing/losing some of the existing issues/PRs/code?

adrhumphreys:
Hey @michalkleiner just a general lack of action on them from both me and the PR's themselves getting reviewed.

Closing this one is more of a time vs value situation for me, the amount of time I would have saved in just running a fork when I run into an issue like this is less than trying to perfect it to get it merged in. The way I see it is, if I move the logic from an env var to it being in the config, will the next person have a different opinion on this change and request another round of changes from me?

As for other issues, we can use https://github.com/silverstripe/silverstripe-admin/issues/1053 as an example, the PR for the change was closed, I figured I'd close the issue to and someone can reopen it or create a new one if we want to progress this

Another example is https://github.com/silverstripe/silverstripe-versioned/issues/252 it had a PR that was hanging around for over a year with no action or feedback which makes me think it's not providing value

michalkleiner:
Thanks for taking the time to reply @adrhumphreys. That's a fair point that I, personally, can't really help with apart from weighing in on some of the issues/PRs. Also not sure who to tag, who to let know it's a potential frustration not just for you, here, but maybe across more modules for more people, or how to improve the status quo.

Could you, perhaps, instead of closing the issue or the PR, give it one last nudge — add a comment whether there's still value in the PR, whether something needs adjust, what can be done to bring it over the line, or so? It just doesn't feel right  seeing good, honest effort goes to waste when it doesn't have to, necessarily. 🙇

Issue #435, opened 2021-02-23
ImageShortcodeProvider doesn't seem to process [img shortcodes with empty title="" or alt=""


micschk:
After running the TagsToShortcode task some images ended up with an empty title or alt attribute, which didn't seem to get processed correctly by the shortcode provider. 

Upon saving, the attributes have four """" instead of two. (Eg `[image ... title="""" alt="""" ]`. Two more get added upon each save...

See also: https://github.com/silverstripe/silverstripe-cms/issues/2489

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-framework/pull/9872

maxime-rainville:
Thanks for taking the time to report this bug.

Can you provide the exact HTML strings that are causing problems, before and after running the `TagsToShortcode` task. This will make it easier for us to debug the issue.

maxime-rainville:
https://github.com/silverstripe/silverstripe-framework/pull/9872 should fix this

PR #436, opened 2021-02-23
Cast image dimensions from shortcode as int

### LINKED ISSUE #434: Valid HTML height/width values cause Exception in [image] shortcode
After running the `TagsToShortcode` task, a valid HTML image size of 50% was copied into the tag attribute. 
```html
<img class="image" src="some-image.png" alt="..." width="50%" height="auto" data-id="87299" data-shortcode="image" />
```

This results in an exception in the `ImageShortcodeProvider`: `InvalidArgumentException: Width must be a numeric value`.

While this is probably an edge case, I do think (editor supplied) shortcode values should be sanitized and/or fallback to sane defaults by the handler so they're unable to cause an exception?

# Pull request
- [ ] https://github.com/silverstripe/silverstripe-assets/pull/436

micschk:
This prevents sizes like '50%' from triggering exceptions.
Fixes #434 (or at least prevents an exception)
Blocked by https://github.com/silverstripe/silverstripe-framework/pull/9872

### DIFF
--- a/tests/php/Shortcodes/ImageShortcodeProviderTest.php
+++ b/tests/php/Shortcodes/ImageShortcodeProviderTest.php
@@ -89,6 +89,24 @@ class ImageShortcodeProviderTest extends SapphireTest
             ))
         );
     }
+    
+    public function testShortcodeHandlerDoesNotResampleToNonIntegerImagesSizes()
+    {
+        $image = $this->objFromFixture(Image::class, 'imageWithoutTitle');
+        $parser = new ShortcodeParser();
+        $parser->register('image', [ImageShortcodeProvider::class, 'handle_shortcode']);
+
+        $this->assertEquals(
+            sprintf(
+                '<img src="%s" alt="" width="50%" height="auto">',
+                $image->Link()
+            ),
+            $parser->parse(sprintf(
+                '[image id="%d" alt="" width="50%" height="auto"]',
+                $image->ID
+            ))
+        );
+    }
 
     public function testShortcodeHandlerFailsGracefully()
     {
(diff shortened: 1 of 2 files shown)


maxime-rainville (changes requested):
That's an awesome contribution. Would you mind adding a test to `tests/php/Shortcodes/ImageShortcodeProviderTest.php` to cover this edge case?

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -89,6 +89,24 @@ public function testShortcodeHandlerAddsDefaultAttributes()
...
+        $this->assertEquals(
+            sprintf(
+                '<img src="%s" alt="" width="50%" height="auto">',
+                $image->Link()
+            ),
+            $parser->parse(sprintf(
+                '[image id="%d" alt="" width="50%" height="auto"]',
+                $image->ID
+            ))
+        );
```
maxime-rainville, line 108: Test is failing because `sprintf` thinks the `%` in `50%` is meant to be another substitution. This should take care of that problem.

```suggestion
        $this->assertEquals(
            sprintf(
                '<img src="%s" alt="" width="50%%" height="auto">',
                $image->Link()
            ),
            $parser->parse(sprintf(
                '[image id="%d" alt="" width="50%%" height="auto"]',
                $image->ID
            ))
        );
```

Test is till failing because we end up with `alt="&quot;&quot;"` in the alt attribute, which must be related to that other issue you raise https://github.com/silverstripe/silverstripe-assets/issues/435

I'll see if I can figure out what's wrong with that one.

maxime-rainville (changes requested):

tests/php/Shortcodes/ImageShortcodeProviderTest.php
maxime-rainville: Once the above suggestion is applied and https://github.com/silverstripe/silverstripe-framework/pull/9872 is merged, this PR should be green.

maxime-rainville (approved):
Blocking PR has been merged and I've committed my tweaks to the unit test which are now green.

PR #439, opened 2021-03-10
Remove disabled extensions from allowed_extensions

### LINKED ISSUE #438: Can't remove entries from file allowed_extensions list
Using Silverstripe in 4.6.

The documented method here to remove existing entries from the `allowed_extensions` list does not work: https://docs.silverstripe.org/en/4/developer_guides/files/allowed_file_types/

If I add, say:

```
SilverStripe\Assets\File:
  allowed_extensions:
    zip: false
```

Looking at the code, all it does is skip that new entry for `zip`. It doesn't check if zip is already in the list and remove it.

If that is the intended behaviour then perhaps the docs need to be clarified.

andrewandante:
Fixes #438 

Currently the docs example doesn't work: https://docs.silverstripe.org/en/4/developer_guides/files/allowed_file_types/#file-extensions-validation

It just skips over the value and leaves the original in the list. This change removes it from the core list.

### DIFF
--- a/tests/php/FileTest.php
+++ b/tests/php/FileTest.php
@@ -801,7 +801,7 @@ class FileTest extends SapphireTest
     public function testGetAllowedExtensions($allowedExtensions, $expected)
     {
         Config::modify()->set(File::class, 'allowed_extensions', $allowedExtensions);
-        $this->assertSame($expected, $expected);
+        $this->assertSame(array_values($expected), array_values(File::getAllowedExtensions()));
     }
 
     /**
--- a/src/File.php
+++ b/src/File.php
@@ -1446,12 +1446,17 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
                 $value = true;
             }
 
+            $key = strtolower($key);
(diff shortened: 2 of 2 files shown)


tests/php/FileTest.php
```diff
@@ -801,7 +801,7 @@ public function testRename()
     public function testGetAllowedExtensions($allowedExtensions, $expected)
     {
         Config::modify()->set(File::class, 'allowed_extensions', $allowedExtensions);
-        $this->assertSame($expected, $expected);
```
andrewandante, line 804: This is one of the funnier things I've ever had to fix in a PR 😆

maxime-rainville: :see_no_evil:

maxime-rainville (changes requested):
Good fix. This is probably worth putting into a patch release for order version. Can you retarget to the 1.7 branch?

andrewandante:
Will do, just going to rebase 👍

maxime-rainville (approved):
Awesome!!! Merge on green.

PR #446, opened 2021-05-16
NEW: Ensure logged in members can access draft files

unclecheese:
Unfortunately, this worsens the already leaky abstraction in `isGranted()` check, but I can't see any better way to do this. Curious if you have a better approach.

For clarity, the use case is accessing draft files in a headless setup, where we need the API key in the header to grant access statelessly.

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -27,6 +27,7 @@ use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Security\Security;
+use SilverStripe\Versioned\Versioned;
 
 /**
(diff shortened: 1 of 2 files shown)


maxime-rainville (commented):
Looks sensible. There could be some weird edges where you are trying to access a restricted file who was publish with one name and who has a different name in draft.

I would suggest adding a unit test with a published restricted file and published restricted file with an altered name.

There's also a scenario where the developer has activated `keep_archived_assets` and the user is trying to access an archived file ... not sure what we want to do in that scenario ... probably nothing.

unclecheese:
Have added the extra test for published restricted.

unclecheese:
Have added altered filename tests.

dhensby (approved):

michalkleiner:
Nice! This would be useful to us two weeks ago (as always... ) :-D

PR #448, opened 2021-06-11
Make castDimension method public

micschk:
...so its functionality can also be used by image manipulation extensions applied to the Image class.


emteknetnz:
This would be an API change.  Pretty unlikely scenario, though if someone subclassed File (which ImageManipulation is applied to) and overrode `protected function castDimension()` function then their code will break on update

Recommend that you just copy paste `castDimensions()` to your extension

Issue #452, opened 2021-06-22
Global opt-out of `loading` attribute


brynwhyman:
### Overview
With lazy-loading of images published by the CMS being the global default, with the option to opt-out on a per image basis, there should still be an option to globally disable this. 

One example for this would be for projects who at the time of upgrading to a release with this feature, already rely on a third-party library for lazy-loading images.

Background: https://forum.silverstripe.org/t/browser-level-lazy-loading-from-silverstripe-cms/4218

### Acceptance Criteria
- [x] There is the option for a developer to opt-out of this feature globally

### PRs
- [x] https://github.com/silverstripe/silverstripe-assets/pull/454

Issue #453, opened 2021-06-22
Set your own `loading` value globally


brynwhyman:
### Overview

In addition to the [opt-out functionality](https://github.com/silverstripe/silverstripe-assets/issues/452) we've discussed adding the option to allow developers to set their own global value for the `loading` attribute, i.e changing this from `loading="lazy"` to `loading="eager"`, `loading="auto"`, or any future option. 

Background: https://forum.silverstripe.org/t/browser-level-lazy-loading-from-silverstripe-cms/4218

### Acceptance Criteria

- [ ] There is the option for a Developer to set the default value, opting to retain the loading attribute and changing it to something else
- [ ] The downstream affects of changing this to a value won't be supported by the CMS field used by content editors is clear
- [ ] Images that don't have a predefined value default to using this value in the SSViewer
- [ ] Images that don't have a predefined value default to using this value in the image shortcode

### Notes
- A migration task would be out of scope should this be changed halfway through a project and images across the project be in two different states.

brynwhyman:
I think this idea is a no-go. [Google make recommendations against this](https://web.dev/browser-level-lazy-loading-for-cmss/#allow-per-element-modifications):

> Caution: If an element should not be lazy-loaded, require or encourage skipping the `loading` attribute entirely. While using `loading="eager"` is a supported alternative, this would tell the browser explicitly to always load the image right away, which would prevent potential benefits if browsers implemented further mechanisms and heuristics to automatically decide which elements to lazy-load.

emteknetnz:
Happy to close issue, not supporting this reduces complexity

@clarkepaul would mean you only need to design UX around the scenario where lazy loading is the global default

brynwhyman:
Based on the above quote from Google, I can't see a reason why we should go ahead with this. Closing.

bergice:
> I think this idea is a no-go. [Google make recommendations against this](https://web.dev/browser-level-lazy-loading-for-cmss/#allow-per-element-modifications):
> 
> > Caution: If an element should not be lazy-loaded, require or encourage skipping the `loading` attribute entirely. While using `loading="eager"` is a supported alternative, this would tell the browser explicitly to always load the image right away, which would prevent potential benefits if browsers implemented further mechanisms and heuristics to automatically decide which elements to lazy-load.

I don't see how this has anything to do with changing the default.

If they set it to `eager` we can just ignore rendering.

This has nothing to do with rendering vs not rendering things.

[...]

PR #454, opened 2021-06-23
API Add image lazy load config

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-assets/issues/452

### DIFF
--- a/src/Image.php
+++ b/src/Image.php
@@ -25,6 +25,15 @@ class Image extends File
      */
     private static $plural_name = "Images";
 
+    /**
+     * Globally control whether Images added via the WYSIWYG editor or inserted as Image objects in Silverstripe
+     * templates have the loading="lazy" HTML added by default
+     *
+     * @config
+     * @var bool
+     */
+    private static $lazy_loading_enabled = true;
+
     public function __construct($record = null, $isSingleton = false, $queryParams = [])
     {
         parent::__construct($record, $isSingleton, $queryParams);
@@ -59,5 +68,13 @@ class Image extends File
         }
         $this->extend('updatePreviewLink', $link, $action);
         return $link;
+    }
+
+    /**
+     * @return bool
+     */
+    public static function getLazyLoadingEnabled(): bool
+    {
+        return self::config()->get('lazy_loading_enabled');
     }
 }
--- a/tests/php/ImageTest.php
+++ b/tests/php/ImageTest.php
@@ -547,4 +547,13 @@ abstract class ImageTest extends SapphireTest
         $builder->expects($this->any())->method('getIsImage')->will($this->returnValue(true));
         return $builder;
     }
+
+    public function testGetLazyLoadingEnabled()
+    {
+        $this->assertTrue(Image::getLazyLoadingEnabled());
+        Config::withConfig(function () {
+            Config::modify()->set(Image::class, 'lazy_loading_enabled', false);
+            $this->assertFalse(Image::getLazyLoadingEnabled());
+        });
+    }
 }


src/Image.php
```diff
@@ -25,6 +25,12 @@ class Image extends File
      */
     private static $plural_name = "Images";
 
+    /**
+     * @config
```
bergice, line 29: Need a description here.

src/Image.php
```diff
@@ -25,6 +25,12 @@ class Image extends File
...
+     * @var bool
+     */
+    private static $lazy_load = true;
```
bergice, line 32: Maybe it should say something like `$enable_lazy_loading`, as `$lazy_load` is a bit ambiguous.

maxime-rainville: @emteknetnz Can you walk us through why you chose to put this config option here? What other alternative did you consider?

maxime-rainville (commented):
I think it might make sense to have a method that reads this config value and gets called by the other classes that need it.

Also you need unit test for this PR ... which you can't have because none of the other stuff has been implemented yet.

src/Image.php
emteknetnz: Updated

src/Image.php
emteknetnz: Updated

emteknetnz: @maxime-rainville Both ImageShortcodeProvider.php and DBFile.php live in the assets module, so it needed to be a file within this module

Lazy loading only applies to Images, and not generic files.  Image.php lives in assets module and is the base class of, images :-)  So it seems pretty logical.   Other alternatives would be files in the assets src folder, though none seemed better than Image

emteknetnz:
@maxime-rainville Have added function to read config value + unit test

src/Image.php
maxime-rainville: My instinct would have been to put it on a class where it's actually use like `DBFile`.

emteknetnz: ImageShortCodeProvider.php also uses it, so I'm not sure DBFile makes the most sense

It would look kind of weird putting this on ImageShortCodeProvider and then from DBFile calling `if (ImageShortCodeProvider::lazyLoadingEnabled()) {`

`Image::lazyLoadingEnabled()` reads the cleanest IMO

src/Image.php
```diff
@@ -60,4 +69,12 @@ public function PreviewLink($action = null)
         $this->extend('updatePreviewLink', $link, $action);
         return $link;
     }
+
+    /**
+     * @return bool
+     */
+    public static function lazyLoadingEnabled(): bool
```
bergice, line 76:
```suggestion
    public static function getLazyLoadingEnabled(): bool
```

src/Image.php
```diff
@@ -25,6 +25,15 @@ class Image extends File
      */
     private static $plural_name = "Images";
 
+    /**
+     * Globally control whether Images added via the WYSIWYG editor or inserted as Image objects in Silverstripe
+     * templates have the loading="lazy" HTML added by default
```
bergice, line 30: Did we decide that the default value is not configurable?

If not, then the above statement is incorrect.
Otherwise, where did we decide on that?

emteknetnz: Yes, has been discussed a few times and decision has been made to make the default non-configurable.  See https://github.com/silverstripe/silverstripe-assets/issues/453

src/Image.php
emteknetnz: Updated

maxime-rainville (approved):
I guess it's kind of OK ... but there's nothing to test because by itself, this doesn't do anything.

You'll need to make sure you include unit test in the related PRs to validate that setting this to false is honoured. 

This is why I think this should have been the last thing we did, not first one.

Issue #451, opened 2021-06-22
WYSIWYG image renders `loading` attribute


brynwhyman:
**Blocked by https://github.com/silverstripe/silverstripe-asset-admin/issues/1214**

### Overview

Modify `ImageShortcodeProvider.php` to render the new lazy-loading attribute.

Background: https://forum.silverstripe.org/t/browser-level-lazy-loading-from-silverstripe-cms/4218

web.dev article - https://web.dev/browser-level-lazy-loading-for-cmss/

### Acceptance Criteria

- [x] The image shortcode provider is updated so that the loading=lazy attribute is added to the tag by default
- [x]  The loading="lazy" attribute will only be added to images where dimensions are defined
- [ ] ~~Where an image opts out of lazy-loading, the loading attribute should be skipped entirely~~
- [x] Where a site has globally opted out of lazy-loading, the loading attribute is skipped
- [x] Ensure that when the image is loading **in the CMS**  the image does not have the `loading` attribute
- [ ]  ~~A 'dummy' checkbox is created to enable/ disable lazy-loading on a per image basis in the placement tab, to show this working~~
  - ~~we'll follow up on this in a dedicated UX card~~
- [ ] ~~The checkbox does not show if a site has globally opted out of lazy-loading~~

### Notes
- It was discussed that it would be interesting to test having the WYSIWYG **in the CMS** lazy-load the images by default too. 

### PRs
- [x] https://github.com/silverstripe/silverstripe-assets/pull/455

bergice:
You may need to whitelist the `loading` attribute on `img` elements using `extended_valid_elements`, otherwise it will be stripped from the source when the content is saved.

PR #455, opened 2021-06-23
WIP NEW Image lazy-loading for wysiwyg images

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-assets/issues/451

### DIFF
--- a/src/Shortcodes/ImageShortcodeProvider.php
+++ b/src/Shortcodes/ImageShortcodeProvider.php
@@ -71,6 +71,8 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
         }
 
         // Check if a resize is required
+        $width = null;
+        $height = null;
         $src = $record->getURL($allowSessionGrant);
         if ($record instanceof Image) {
             $width = isset($args['width']) ? (int) $args['width'] : null;
@@ -84,6 +86,9 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
                 }
             }
         }
+
+        // Determine whether loading="lazy" is set
+        $args = self::updateLoadingValue($args, $width, $height);
 
         // Build the HTML tag
         $attrs = array_merge(
@@ -178,5 +183,30 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
         return Image::create([
             'Title' => _t(__CLASS__ . '.IMAGENOTFOUND', 'Image not found'),
         ]);
+    }
+
+    /**
+     * Updated the loading attribute which is used to either lazy-load or eager-load images
+     * Eager-load is the default browser behaviour so when eager loading is specified, the
+     * loading attribute is omitted
+     *
+     * @param array $args
+     * @param int|null $width
+     * @param int|null $height
+     * @return array
+     */
+    private static function updateLoadingValue(array $args, ?int $width, ?int $height): array
+    {
+        if (!Image::getLazyLoadingEnabled()) {
+            return $args;
+        }
+        if (isset($args['loading']) && $args['loading'] == 'eager') {
+            // per image override - unset the loading attribute unset to eager load (default browser behaviour)
+            unset($args['loading']);
+        } elseif ($width && $height) {
+            // width and height must be present to prevent content shifting
+            $args['loading'] = 'lazy';
+        }
+        return $args;
     }
 }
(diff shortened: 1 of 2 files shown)


src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -85,6 +88,17 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
             }
         }
 
+        // Image lazy loading - must be globally enabled (is enabled by default)
```
maxime-rainville, line 91: This method is already too big as is. New logic should be in a new `private` method. If time allows split the existing method into smaller private method as well.

emteknetnz: Updated

bergice (changes requested):
It doesn't look like this will lazy load images automatically until you've saved the page again. I was expecting that this is something that will automatically work out of the box for all previous data ideally without a data migration.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -179,4 +185,25 @@ protected static function find_error_record($errorCode)
             'Title' => _t(__CLASS__ . '.IMAGENOTFOUND', 'Image not found'),
         ]);
     }
+
+    /**
+     * @param array $args
+     * @param int|null $width
+     * @param int|null $height
+     * @return array
+     */
```
bergice, line 194: Could use a description in here.

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -149,4 +150,37 @@ public function testMissingImageDoesNotCache()
...
+        // missing width
+        $shortcode = '[image id="' . $id . '" height="200"]';
+        $this->assertNotContains('loading="lazy"', $parser->parse($shortcode));
+
+        // missing height
+        $shortcode = '[image id="' . $id . '" width="300"]';
+        $this->assertNotContains('loading="lazy"', $parser->parse($shortcode));
+
+        // loading="eager"
+        $shortcode = '[image id="' . $id . '" width="300" height="200" loading="eager"]';
```
maxime-rainville, line 174: Maybe add a case where some non-sense gets put in the loading attribute

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -71,6 +72,8 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
         }
 
         // Check if a resize is required
+        $width = null;
```
maxime-rainville, line 75: Thinking it could have been interesting to read the width/height from the record when it's not provided via the argument, but that's super edge casy ... and there's bound to be a mad scientist out there that explicictely doesn't want a width or height on its image tag.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -6,6 +6,7 @@
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Image;
 use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Core\Config\Config;
```
maxime-rainville, line 9: Don't think that's actually needed.

maxime-rainville (changes requested):
Basically, apply @bergice on line 200 and it's all good.

src/Shortcodes/ImageShortcodeProvider.php
emteknetnz: Updated

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -179,4 +185,25 @@ protected static function find_error_record($errorCode)
...
+     * @param int|null $width
+     * @param int|null $height
+     * @return array
+     */
+    private static function updateLoadingValue(array $args, ?int $width, ?int $height): array
+    {
+        if (!Image::getLazyLoadingEnabled()) {
+            return $args;
+        }
+        if (isset($args['loading']) && $args['loading'] = 'eager') {
```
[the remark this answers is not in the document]
emteknetnz: Updated

tests/php/Shortcodes/ImageShortcodeProviderTest.php
emteknetnz: I have added nonsense

src/Shortcodes/ImageShortcodeProvider.php
emteknetnz: Removed

maxime-rainville (approved):
Confirmed the shortcode defaults to adding the lazy loading attribute if nothing is provided ... and there's a unit test there for that.

Issue #450, opened 2021-06-22
SSViewer image renders `loading` attribute


brynwhyman:
### Overview

Modify `DBFile.php` and `DBFile_image.ss` to render the new lazy-loading attribute.

Background: https://forum.silverstripe.org/t/browser-level-lazy-loading-from-silverstripe-cms/4218

### Acceptance Criteria
- [ ] It is possible for the `loading="lazy"` attribute to be added when you render an image in the template
- [ ] The `loading="lazy"` attribute will only be added to images where [dimensions are defined](https://web.dev/browser-level-lazy-loading-for-cmss/#require-dimension-attributes-on-elements) 
- [ ] There is a method to opt-out of lazy-loading on a per image basis
- [ ] Where an image opts out of lazy-loading, the loading attribute should be skipped entirely
- [ ] Where a site globally opts out of lazy-loading, the loading attribute is skipped

### Notes
- When it comes to docs, make sure it's clear that as a developer, if you manually write your image tag, this won't apply.

### PR's

- [ ] https://github.com/silverstripe/silverstripe-assets/pull/458
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/9995

bergice:
### Proposed implementation

1. Add `DBFile.getLoadingAttribute` API for retrieving the lazy load value (or `null`)
   Returns `null` if `Image.enable_lazy_load` is `false`
   Otherwise, it returns the value of `Image.default_lazy_load`
    a. To opt out on a per-image basis, extend this function
2. Modify `DBFile_image.ss` to render the new lazy-loading attribute if not `null`
    If the loading attribute is supplied we also supply the `width`, `height` attributes

> Re AC:  The loading="lazy" attribute will only be added to images where dimensions are defined

We should just supply the width and height attributes on our end if lazy loading is enabled.
We've already got this data in the `Image` class anyways, and users would just expect this feature to work out of the box.

emteknetnz:
>> Add DBFile.getLoadingAttribute API

In the intention here to return the whole attribute e.g. `loading="lazy"` or just `lazy`?  The originally cited [web.dev article](https://web.dev/browser-level-lazy-loading-for-cmss/#allow-per-element-modifications) recommends only either showing the `loading="lazy"` attribute or completely omitted it (see the warning section below the linked section above)

Because of this I think it would make sense to return the entire `loading="lazy"` string, or an empty string (i.e. not return null)

> To opt out on a per-image basis, extend this function

[...]

PR #457, opened 2021-06-25
Add lazy loading to images

bergice:
Resolves https://github.com/silverstripe/silverstripe-assets/issues/450

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -63,7 +63,7 @@ trait ImageManipulation
      *
      * @var bool
      */
-    protected $lazyLoad = true;
+    private $lazyLoad = true;
 
     /**
      * Set whether image resizes are allowed
@@ -794,24 +794,34 @@ trait ImageManipulation
      *
      * @return bool
      */
-    public function IsLazyLoaded()
+    public function getIsLazyLoaded() : bool
     {
-        if (!Config::inst()->get('SilverStripe\\Assets\\Image', 'lazy_loading_enabled')) {
-            return false;
+        if (Image::getLazyLoadingEnabled() && $this->getWidth() && $this->getHeight()) {
+            return $this->lazyLoad;
         }
-        return $this->lazyLoad;
+        return false;
     }
 
     /**
      * Set whether image will be lazy loaded
      *
      * @param bool $lazyLoad
-     * @return $this
+     * @return self $this
      */
-    public function LazyLoad($lazyLoad)
+    public function LazyLoad(bool $lazyLoad): self
     {
         $this->lazyLoad = $lazyLoad;
         return $this;
+    }
+
+    /**
+     * Get whether image will be lazy loaded
+     *
+     * @return bool
+     */
+    public function getLazyLoad(): bool
+    {
+        return $this->lazyLoad;
     }
 
     /**
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -59,6 +59,13 @@ trait ImageManipulation
     protected $allowGeneration = true;
 
     /**
+     * If image is lazy loaded
+     *
+     * @var bool
+     */
+    protected $lazyLoad = true;
+
+    /**
      * Set whether image resizes are allowed
      *
      * @param bool $allow
@@ -783,6 +790,31 @@ trait ImageManipulation
     }
 
     /**
+     * Return whether image will be lazy loaded
+     *
+     * @return bool
+     */
+    public function IsLazyLoaded()
+    {
+        if (!Config::inst()->get('SilverStripe\\Assets\\Image', 'lazy_loading_enabled')) {
+            return false;
+        }
+        return $this->lazyLoad;
+    }
+
+    /**
+     * Set whether image will be lazy loaded
+     *
+     * @param bool $lazyLoad
+     * @return $this
+     */
+    public function LazyLoad($lazyLoad)
+    {
+        $this->lazyLoad = $lazyLoad;
+        return $this;
+    }
+
+    /**
      * Get the orientation of this image.
      *
      * @return int ORIENTATION_SQUARE | ORIENTATION_PORTRAIT | ORIENTATION_LANDSCAPE
@@ -951,7 +983,7 @@ trait ImageManipulation
                 $tuple = $result;
                 Deprecation::notice(
                     '5.0',
-                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array 
+                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array
                         containing a tuple and an image backend, i.e. [$tuple, $result]',
                     Deprecation::SCOPE_GLOBAL
                 );
--- a/templates/DBFile_image.ss
+++ b/templates/DBFile_image.ss
@@ -1,9 +1,1 @@
-<img
-        src="$URL.ATT"
-        alt="$Title.ATT"
-        <% if $IsLazyLoaded %>
-            loading="lazy"
-            width="$Width.ATT"
-            height="$Height.ATT"
-        <% end_if %>
-/>
+<img src="$URL.ATT" alt="$Title.ATT"<% if $IsLazyLoaded %> loading="lazy" width="$Width.ATT" height="$Height.ATT"<% end_if %> />
--- a/templates/DBFile_image.ss
+++ b/templates/DBFile_image.ss
@@ -1,1 +1,1 @@
-<img src="$URL.ATT" alt="$Title.ATT"<% if $IsLazyLoaded %> loading="lazy" width="$Width.ATT" height="$Height.ATT"<% end_if %> />
+<img src="$URL.ATT" alt="$Title.ATT"<% if $IsLazyLoaded %> loading="lazy"<% end_if %> width="$Width.ATT" height="$Height.ATT" />
--- a/templates/DBFile_image.ss
+++ b/templates/DBFile_image.ss
@@ -1,1 +1,9 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
+<img
+        src="$URL.ATT"
+        alt="$Title.ATT"
+        <% if $IsLazyLoaded %>
+            loading="lazy"
+            width="$Width.ATT"
+            height="$Height.ATT"
+        <% end_if %>
+/>
--- a/tests/php/ImageTest.php
+++ b/tests/php/ImageTest.php
@@ -532,6 +532,22 @@ abstract class ImageTest extends SapphireTest
         $this->assertEquals(99999, $image->getWidth());
     }
 
+    public function testLazyLoad()
+    {
+        Config::modify()->set(Image::class, 'lazy_loading_enabled', true);
+
+        /** @var Image $image */
+        $image = $this->objFromFixture(Image::class, 'imageWithTitle');
+        $this->assertTrue($image->IsLazyLoaded(), 'Images lazy load by default');
+
+        $image->LazyLoad(false);
+        $this->assertFalse($image->IsLazyLoaded(), 'Images can be eager loaded on a per-image basis');
+
+        $image->LazyLoad(true);
+        Config::modify()->set(Image::class, 'lazy_loading_enabled', false);
+        $this->assertFalse($image->IsLazyLoaded(), 'Lazy loading can be disabled globally');
+    }
+
     /**
      * @param $filename
      * @param $hash
--- a/tests/php/ImageTest.php
+++ b/tests/php/ImageTest.php
@@ -540,12 +540,18 @@ abstract class ImageTest extends SapphireTest
         $image = $this->objFromFixture(Image::class, 'imageWithTitle');
         $this->assertTrue($image->IsLazyLoaded(), 'Images lazy load by default');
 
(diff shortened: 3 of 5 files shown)


src/ImageManipulation.php
```diff
@@ -58,6 +58,13 @@ trait ImageManipulation
      */
     protected $allowGeneration = true;
 
+    /**
+     * If image is lazy loaded
+     *
+     * @var bool
+     */
+    protected $lazyLoad = true;
```
emteknetnz, line 66:
```suggestion
    private $lazyLoad = true;
```

src/ImageManipulation.php
```diff
@@ -782,6 +789,31 @@ public function getHeight()
         return 0;
     }
 
+    /**
+     * Return whether image will be lazy loaded
+     *
+     * @return bool
+     */
+    public function IsLazyLoaded()
```
emteknetnz, line 797:
```suggestion
    public function getIsLazyLoaded(): bool
```

src/ImageManipulation.php
```diff
@@ -782,6 +789,31 @@ public function getHeight()
...
+    {
+        if (!Config::inst()->get('SilverStripe\\Assets\\Image', 'lazy_loading_enabled')) {
```
emteknetnz, line 799:
```suggestion
        if (!Image::getLazyLoadingEnabled()) {
```

src/ImageManipulation.php
```diff
@@ -782,6 +789,31 @@ public function getHeight()
...
+        return $this->lazyLoad;
+    }
+
+    /**
+     * Set whether image will be lazy loaded
+     *
+     * @param bool $lazyLoad
+     * @return $this
+     */
+    public function LazyLoad($lazyLoad)
```
emteknetnz, line 811:
```suggestion
    public function setLazyLoad(bool $lazyLoad)
```

you'll also need a corredsponding `function getLazyLoad(): bool`

src/ImageManipulation.php
```diff
@@ -782,6 +789,31 @@ public function getHeight()
...
+    {
+        $this->lazyLoad = $lazyLoad;
```
emteknetnz, line 813: Need to check that width and height exist on the image

If not, return false

Will need to unit test this

templates/DBFile_image.ss
```diff
@@ -1 +1 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
+<img src="$URL.ATT" alt="$Title.ATT"<% if $IsLazyLoaded %> loading="lazy" width="$Width.ATT" height="$Height.ATT"<% end_if %> />
```
emteknetnz, line 1: We should probably add width + height regardless whether `$IsLazyLoaded` is true or not

emteknetnz (changes requested):
Approach looks good

bergice:
@emteknetnz I've addressed your feedback.
No idea what's going on with failing build.

src/ImageManipulation.php
```diff
@@ -782,6 +789,41 @@ public function getHeight()
...
+        return false;
+    }
+
+    /**
+     * Set whether image will be lazy loaded
+     *
+     * @param bool $lazyLoad
+     * @return $this
+     */
+    public function setLazyLoad($lazyLoad)
```
maxime-rainville, line 811: I think this works.

```suggestion
    public function setLazyLoad($lazyLoad): self
```

templates/DBFile_image.ss
```diff
@@ -1 +1 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
+<img src="$URL.ATT" alt="$Title.ATT"<% if $IsLazyLoaded %> loading="lazy"<% end_if %> width="$Width.ATT" height="$Height.ATT" />
```
maxime-rainville, line 1: Still think we should use getAttributes/getAttributeHTML here te be consistent with other parts of the CMS.

tests/php/ImageTest.php
```diff
@@ -532,6 +532,36 @@ public function testGetSetImageBackend()
...
+
+        $expected = '<img src="/assets/ImageTest/folder/test-image.png" alt="This is a image Title" loading="lazy" width="300" height="300" />';
+        $actual = trim($image->getTag());
+        $this->assertEquals($expected, $actual, 'Lazy load img tag renders correctly');
+
+        $image->setLazyLoad(false);
+        $this->assertFalse($image->getIsLazyLoaded(), 'Images can be eager loaded on a per-image basis');
+
+        $image->setLazyLoad(true);
+        Config::modify()->set(Image::class, 'lazy_loading_enabled', false);
```
maxime-rainville, line 551: You could [use the new withConfig API](https://github.com/silverstripe/silverstripe-framework/commit/e2bea6b41f05f2ad9d492d2e96a535dfff9f16ef) here

maxime-rainville (changes requested):
`ImageManipulation` is applied to `File` and `DBFile` as well as `Image`. I would expect the implementation and tests accounts for this (e.g. What's the expected results if you call `getIsLazyLoaded` on a PDF File? Where's the test for this scenario.)

templates/DBFile_image.ss
bergice: Sure, but that is out of scope of this issue.

src/ImageManipulation.php
emteknetnz: `bool $lazyLoad`

templates/DBFile_image.ss
maxime-rainville: I don't think taking the time to analyse a problem and finding the most sensible way of solving it is a question of "scope".

If you don't like the "getAttributes/getAttributeHTML" approach or if you think it's worth the extra work, explain why ... just don't call it out-of-scope.

bergice: The scope of this project is to make images lazy load, not create fancy new API's that will increase the total time we spend on this epic.

I like the idea of having an API for rendering attributes that is not exclusive to form fields, but it would require some refactoring and is considered a generic improvement / minor feature.
If you really need this done then either do it yourself or create another issue.

TLDR; it's not worth the extra work.

src/ImageManipulation.php
```diff
@@ -782,6 +789,41 @@ public function getHeight()
...
+        return false;
+    }
+
+    /**
+     * Set whether image will be lazy loaded
+     *
+     * @param bool $lazyLoad
+     * @return self $this
+     */
+    public function setLazyLoad(bool $lazyLoad): self
```
maxime-rainville, line 811: If you have this in your template
```
<h2>Not Lazy Loaded</h2>
$ScaleWidth(100).setLazyLoad(0)
$setLazyLoad(0)
$setLazyLoad(1).$setLazyLoad(0)

<h2>Lazy Loaded</h2>
$ScaleWidth(100)
$Me
$setLazyLoad(1)
```

All the rendered images will not have a `loading` attribute because only the first rendered image is honoured. If you reverse the two blocks all the Image will have the loading attribute.

`setLazyLoad` needs to return a copy of itself like the other `ImageManipulation` function.

maxime-rainville: This method probably should be start with an upper case letter to match the other image manipulation methods.

maxime-rainville: We need to remove the explicit `bool` type so people can provide the value as `'false'` or `'0'` because of a limitation in our SS rendering engine.

templates/DBFile_image.ss
maxime-rainville: You just need to copy this bit of logic and adjust it for your purposes https://github.com/silverstripe/silverstripe-framework/blob/dcdc25500be729f4340d42f8a4fc797461f862f3/src/Forms/FormField.php#L680-L757

It shouldn't be that long. Plus you'll need refactor the entire thing anyway to address the problem with sequentially rendered images.

maxime-rainville (changes requested):

src/ImageManipulation.php
bergice: > This method probably should be start with an upper case letter to match the other image manipulation methods.

Agree, I figured we should do that as well.

> We need to remove the explicit bool type so people can provide the value as 'false' or '0' because of a limitation in our SS rendering engine.

That's not gonna work because it will just set the variable to `false` as a string.
We should really just expect a 0/1 value like every other template API already does.

bergice: > All the rendered images will not have a loading attribute because only the first rendered image is honoured. If you reverse the two blocks all the Image will have the loading attribute.
> 
> setLazyLoad needs to return a copy of itself like the other ImageManipulation function.

I'm struggling to see how this would matter in a real world scenario. What would be the point of lazy loading and not lazy load the same image? After all, it's pointing to the same URL.

The only use case I see is if you manipulate the image, and if you do that it will create a copy and toggling the lazy load for the same image with a different variant will work as expected.

templates/DBFile_image.ss
bergice: > You just need to copy this bit of logic and adjust it for your purposes https://github.com/silverstripe/silverstripe-framework/blob/dcdc25500be729f4340d42f8a4fc797461f862f3/src/Forms/FormField.php#L680-L757

"Just".

My previous arguments against doing this still stand.
In addition to that I'm against copying identical code. If anything, we should be refactoring the other methods as well so they use the same code, which is not a trivial effort.

> It shouldn't be that long. Plus you'll need refactor the entire thing anyway to address the problem with sequentially rendered images.

That is not related.

src/ImageManipulation.php
maxime-rainville: As a developer, I shouldn't have to worry about how `SetLazyLoad(false)` will be cast internally. It should just work. So in the method, we should just convert a 'false' string to an actual boolean.

> I'm struggling to see how this would matter in a real world scenario. What would be the point of lazy loading and not lazy load the same image? After all, it's pointing to the same URL.

SSViewer tries to cache things as much as possible as it goes along. You could end up with weird scenario where you have an include that gets call in different context. We'll just end up with weird bugs and have inconsistent result if we do it this way.

That's why `ImageManipulation` has all this convoluted logic and jumps through all these hoops to create duplicate object of Image.

bergice: > As a developer, I shouldn't have to worry about how SetLazyLoad(false) will be cast internally. It should just work. 

Yes, it should. However this is something that should be solved for templates in general, not just hardcoded into individual methods.

It is clear that this is something that will keep being an issue unless we fix it properly, and I don't think this "duct-taping" fix approach is the right way to solve it. https://github.com/silverstripe/silverstripe-cms/commit/ed0680a26464dbc37ef35a3abd4e63014e361fd3

> SSViewer tries to cache things as much as possible as it goes along. You could end up with weird scenario where you have an include that gets call in different context. We'll just end up with weird bugs and have inconsistent result if we do it this way.
> 
> That's why ImageManipulation has all this convoluted logic and jumps through all these hoops to create duplicate object of Image.

I'm not sure about that yet. Do you have an example? I'm under the impression it duplicates the image after altering it / when the variant changes.

kinglozzer: > > As a developer, I shouldn't have to worry about how SetLazyLoad(false) will be cast internally. It should just work.
> 
> Yes, it should. However this is something that should be solved for templates in general, not just hardcoded into individual methods.
> 
> It is clear that this is something that will keep being an issue unless we fix it properly, and I don't think this "duct-taping" fix approach is the right way to solve it.

FYI this has been resolved in 5.x/dev-master with https://github.com/silverstripe/silverstripe-framework/pull/8152 (and https://github.com/silverstripe/silverstripe-framework/pull/8456/ for null values, which was a marginally more invasive change). I treated it as an API change at the time, but I suppose we could look at whether it’d be possible to cherry-pick those changes into 4.x?

I wouldn’t be opposed to “duct-taping” something like this for now. It’s one line of zero-maintenance code to reduce a bunch of confusion/extra documentation for developers using this.

bergice: Yeah, I was suggest cherry-picking as well. Might be fine as long as we consider it a bug.

maxime-rainville: It duplicates the PHP object, but it doesn't dupilcate the actual File database entry or the physical file.

The idea is to stack up all the image manipulation calls, but wait until you actually render the image at the end to execute them. Otherwise, you would be running a bunch of expensive (and potentially unneeded) image manipulation operations immediately when calrling a method like `CropHeight`.

My guess is you probably just need to do something like this to duplicate you current object.
```php
$this->manipulate('', function($callback, $store, $filename, $hash, $variant) {
    return [
    'Filename' => $filename,
    'Hash' => $hash,
    'Variant' => $this->getVariant()
]
});
```

Then you set a lazy loading flag on you duplicate.

You need to make sure that your lazy loading flag gets copied across to any follow up calls to `ImageManipulation::manipulate()`.

emteknetnz:
Unfortunately this approach does not appear to do what it needs to work as intended.  While the unit tests pass, adding an image in the template as `$MyImage.LazyLoad(false)` doesn't work because:

a) the argument is passed to PHP as a string "false" - which is fixiable within the getLazyLoaded function, however
b) the value in the private field $lazyLoad is not retained, I believe because ImageManipulation is a bit of an abstraction over DBFile

I'm closing this in favour of the alternative approach here https://github.com/silverstripe/silverstripe-assets/pull/458

PR #458, opened 2021-07-05
Add lazy load to Image/DBFile rendering

maxime-rainville:
# Parent issue
- fixes https://github.com/silverstripe/silverstripe-assets/issues/450

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -1256,8 +1256,8 @@ trait ImageManipulation
         }
 
         $equivalence = [
-            'eager' => [false, 0, '0', 'false', 'eager'],
-            'lazy' => [true, 1, '1', 'true', 'lazy'],
+            'eager' => [false, 0, '0', 'false'],
+            'lazy' => [true, 1, '1', 'true'],
         ];
 
         foreach ($equivalence as $loading => $vals) {
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -42,6 +42,9 @@ use SilverStripe\View\HTML;
  * - Icon()
  * - CMSThumbnail()
  *
+ * Other manipulations that do not create variants:
+ * - LazyLoad()
+ *
  * @mixin AssetContainer
  */
 trait ImageManipulation
@@ -57,6 +60,12 @@ trait ImageManipulation
      * @var bool
      */
     protected $allowGeneration = true;
+
+    /**
+     * List of attributes to render on the frontend
+     * @var array
+     */
+    protected $attributes = [];
 
     /**
      * Set whether image resizes are allowed
@@ -951,7 +960,7 @@ trait ImageManipulation
                 $tuple = $result;
                 Deprecation::notice(
                     '5.0',
-                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array 
+                    'Closure passed to ImageManipulation::manipulate() should return null or a two-item array
                         containing a tuple and an image backend, i.e. [$tuple, $result]',
                     Deprecation::SCOPE_GLOBAL
                 );
@@ -980,6 +989,9 @@ trait ImageManipulation
         if ($manipulationResult instanceof Image_Backend) {
             $file->setImageBackend($manipulationResult);
         }
+
+        // Copy our existing attributes to the new object
+        $file->initAttributes($this->attributes);
 
         return $file->setOriginal($this);
     }
@@ -1056,5 +1068,205 @@ trait ImageManipulation
             throw new InvalidArgumentException("{$dimension} must be positive");
         }
         return $value;
+    }
+
+    /**
+     * Create an equivalent DBFile object
+     * @return AssetContainer
+     */
+    private function copyImageBackend(): AssetContainer
+    {
+        // Store result in new DBFile instance
+        /** @var DBFile $file */
+        $file = DBField::create_field(
+            'DBFile',
+            [
+                'Filename' => $this->getFilename(),
+                'Hash' => $this->getHash(),
+                'Variant' => $this->getVariant()
+            ]
+        );
+
+        $backend = $this->getImageBackend();
+        if ($backend) {
+            $file->setImageBackend($backend);
+        }
+
+        return $file->setOriginal($this);
+    }
+
+    /**
+     * Add a HTML attribute when rendering this object. This method is immutable, it will return you a copy of the
+     * original object.
+     * @param string $name
+     * @param mixed $value
+     * @return AssetContainer
+     */
+    public function setAttribute($name, $value)
+    {
+        $file = $this->copyImageBackend();
+        $file->initAttributes(array_merge(
+            $this->attributes,
+            [$name => $value]
+        ));
+
+        return $file;
+    }
+
+    /**
+     * Initialise the attirbutes on this object. Should only be called imiditely after intialising the object.
+     * @param array $attributes
+     * @internal
+     */
+    public function initAttributes(array $attributes)
+    {
+        // This would normally be in a constructor, but there's no way for a trait to interact with that.
+        $this->attributes = $attributes;
+    }
+
+    /**
+     * Retrieve the value of an HTML attribute
+     * @param $name
+     * @return mixed|null
+     */
+    public function getAttribute($name)
+    {
+        $attributes = $this->getAttributes();
+
+        if (isset($attributes[$name])) {
+            return $attributes[$name];
+        }
+
+        return null;
+    }
+
+    /**
+     * Allows customization through an 'updateAttributes' hook on the base class.
+     * Existing attributes are passed in as the first argument and can be manipulated,
+     * but any attributes added through a subclass implementation won't be included.
+     *
+     * @return array
+     */
+    public function getAttributes()
+    {
+        $attributes = [];
+        if ($this->getIsImage()) {
+            $attributes = [
+                'width' => $this->getWidth(),
+                'height' => $this->getHeight(),
+                'alt' => $this->getTitle(),
+                'src' => $this->getURL()
+            ];
+
+            if ($this->IsLazyLoaded()) {
+                $attributes['loading'] = 'lazy';
+            }
+        }
+
+        $attributes = array_merge($attributes, $this->attributes);
+
+        if (isset($attributes['loading']) && $attributes['loading'] === 'eager') {
+            unset($attributes['loading']);
+        }
+
+        $this->extend('updateAttributes', $attributes);
+
+        return $attributes;
+    }
+
+    /**
+     * Custom attributes to process. Falls back to {@link getAttributes()}.
+     *
+     * If at least one argument is passed as a string, all arguments act as excludes, by name.
+     *
+     * @param array $attributes
+     * @return string
+     */
+    public function getAttributesHTML($attributes = null)
+    {
+        $exclude = null;
+
+        if (is_string($attributes)) {
+            $exclude = func_get_args();
+        }
+
+        if (!$attributes || is_string($attributes)) {
+            $attributes = $this->getAttributes();
+        }
+
+        $attributes = (array) $attributes;
+
+        $attributes = array_filter($attributes, function ($v) {
+            return ($v || $v === 0 || $v === '0');
+        });
+
+        if ($exclude) {
+            $attributes = array_diff_key(
+                $attributes,
+                array_flip($exclude)
+            );
+        }
+
+        // Create markup
+        $parts = [];
+
+        foreach ($attributes as $name => $value) {
+            if ($value === true) {
+                $value = $name;
+            } else {
+                if (is_scalar($value)) {
+                    $value = (string) $value;
+                } else {
+                    $value = json_encode($value);
+                }
+            }
+
+            $parts[] = sprintf('%s="%s"', Convert::raw2att($name), Convert::raw2att($value));
+        }
+
+        return implode(' ', $parts);
+    }
+
+    /**
+     * Determine whether the image should be lazy loaded
+     *
+     * @return bool
+     */
+    public function IsLazyLoaded(): bool
+    {
+        if (Image::getLazyLoadingEnabled() && $this->getIsImage()) {
+            return empty($this->attributes['loading']) || $this->attributes['loading'] === 'lazy';
+        }
+        return false;
+    }
+
+    /**
+     * Set the lazy loading state for this Image
+     * @param mixed $lazyLoad
+     * @return AssetContainer
+     */
+    public function LazyLoad($lazyLoad): AssetContainer
+    {
+        if (!Image::getLazyLoadingEnabled() || !$this->getIsImage()) {
+            return $this;
+        }
+
+        if (is_string($lazyLoad)) {
+            $lazyLoad = strtolower($lazyLoad);
+        }
+
+        $equivalence = [
+            'eager' => [false, 0, '0', 'false', 'eager'],
+            'lazy' => [true, 1, '1', 'true', 'lazy'],
+        ];
+
+        foreach ($equivalence as $loading => $vals) {
+            foreach ($vals as $val) {
+                if ($val === $lazyLoad) {
+                    return $this->setAttribute('loading', $loading);
+                }
+            }
+        }
+        return $this;
     }
 }
--- a/src/Storage/DBFile.php
+++ b/src/Storage/DBFile.php
@@ -113,6 +113,7 @@ class DBFile extends DBComposite implements AssetContainer, Thumbnail
         'MimeType' => 'Varchar',
         'String' => 'Text',
         'Tag' => 'HTMLFragment',
+        'getTag' => 'HTMLFragment',
         'Size' => 'Varchar',
         'AttributesHTML' => 'HTMLFragment',
         'getAttributesHTML' => 'HTMLFragment',
--- a/src/Storage/DBFile.php
+++ b/src/Storage/DBFile.php
@@ -113,7 +113,9 @@ class DBFile extends DBComposite implements AssetContainer, Thumbnail
         'MimeType' => 'Varchar',
         'String' => 'Text',
         'Tag' => 'HTMLFragment',
-        'Size' => 'Varchar'
+        'Size' => 'Varchar',
+        'AttributesHTML' => 'HTMLFragment',
+        'getAttributesHTML' => 'HTMLFragment',
     ];
 
     public function scaffoldFormField($title = null, $params = null)
--- a/templates/DBFile_image.ss
+++ b/templates/DBFile_image.ss
@@ -1,1 +1,1 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
+<img $AttributesHTML />
--- a/src/File.php
+++ b/src/File.php
@@ -164,6 +164,8 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         'getTreeTitle' => 'HTMLFragment',
         'Tag' => 'HTMLFragment',
         'getTag' => 'HTMLFragment',
+        'AttributesHTML' => 'HTMLFragment',
+        'getAttributesHTML' => 'HTMLFragment',
     ];
 
     private static $table_name = 'File';
--- a/src/Image.php
+++ b/src/Image.php
@@ -25,11 +25,6 @@ class Image extends File
      */
     private static $plural_name = "Images";
 
-    private static $casting = [
-        'AttributesHTML' => 'HTMLFragment',
-        'getAttributesHTML' => 'HTMLFragment',
-    ];
-
     /**
      * Globally control whether Images added via the WYSIWYG editor or inserted as Image objects in Silverstripe
      * templates have the loading="lazy" HTML added by default
--- a/src/Image.php
+++ b/src/Image.php
@@ -25,6 +25,11 @@ class Image extends File
      */
     private static $plural_name = "Images";
 
+    private static $casting = [
+        'AttributesHTML' => 'HTMLFragment',
+        'getAttributesHTML' => 'HTMLFragment',
+    ];
+
     /**
      * Globally control whether Images added via the WYSIWYG editor or inserted as Image objects in Silverstripe
      * templates have the loading="lazy" HTML added by default
--- a/tests/php/ImageManipulationTest.php
+++ b/tests/php/ImageManipulationTest.php
@@ -148,7 +148,7 @@ class ImageManipulationTest extends SapphireTest
                 'src' => $image->getURL(),
                 'loading' => 'lazy'
             ],
-            $image->getAttributes(),
+            $image->getAttributes()
         );
 
         $this->assertEquals(
--- a/tests/php/ImageManipulationTest.php
+++ b/tests/php/ImageManipulationTest.php
@@ -1,0 +1,434 @@
+<?php
+
+namespace SilverStripe\Assets\Tests;
+
+use InvalidArgumentException;
+use PHPUnit_Framework_MockObject_MockBuilder;
+use PHPUnit_Framework_MockObject_MockObject;
+use Prophecy\Prophecy\ObjectProphecy;
+use Silverstripe\Assets\Dev\TestAssetStore;
+use SilverStripe\Assets\File;
+use SilverStripe\Assets\Folder;
+use SilverStripe\Assets\Image;
+use SilverStripe\Assets\Image_Backend;
+use SilverStripe\Assets\InterventionBackend;
+use SilverStripe\Assets\Storage\AssetContainer;
+use SilverStripe\Assets\Storage\AssetStore;
+use SilverStripe\Assets\Storage\DBFile;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\View\SSViewer;
+
+/**
+ * ImageTest is abstract and should be overridden with manipulator-specific subtests
+ * @skipUpgrade
+ */
+class ImageManipulationTest extends SapphireTest
+{
+    protected static $fixture_file = 'ImageTest.yml';
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        // Set backend root to /ImageTest
+        TestAssetStore::activate('ImageTest');
+
+        // Copy test images for each of the fixture references
+        /** @var File $image */
+        $files = File::get()->exclude('ClassName', Folder::class);
+        foreach ($files as $image) {
+            $sourcePath = __DIR__ . '/ImageTest/' . $image->Name;
+            $image->setFromLocalFile($sourcePath, $image->Filename);
+            $image->publishSingle();
+        }
+
+        // Set default config
+        InterventionBackend::config()->set('error_cache_ttl', [
+            InterventionBackend::FAILED_INVALID => 0,
+            InterventionBackend::FAILED_MISSING => '5,10',
+            InterventionBackend::FAILED_UNKNOWN => 300,
+        ]);
+    }
+
+    public function tearDown()
+    {
+        TestAssetStore::reset();
+        parent::tearDown();
+    }
+
+    public function testAttribute()
+    {
+        /** @var Image $origin */
+        $origin = $this->objFromFixture(Image::class, 'imageWithTitle');
+
+        $this->assertEmpty(
+            $origin->getAttribute('foo'),
+            'Undefined attributes are falsy'
+        );
+
+        /** @var DBFile $image */
+        $image = $origin->setAttribute('foo', 'bar');
+        $this->assertEquals(
+            'bar',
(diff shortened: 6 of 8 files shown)


src/Attributable.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+namespace SilverStripe\Assets;
+
+use SilverStripe\Core\Convert;
+
+trait Attributable
```
maxime-rainville, line 7: This trait is made up of code copied from FormField. I've got here for now, but if we go ahead with this, I would put it in Framework

The idea is that a bunch of classes use this attribute handling logic. By putting in one trait, we'll be bale to have a standardised pattern we can reuse across a bunch of classes.

This is a bit of retroactive class composition. Not overly fuss about this, we could just put all that logic directly on `ImageManipulation'

src/Attributable.php
```diff
@@ -0,0 +1,112 @@
...
+{
+    protected $attributes = [];
+
+//    private static $casting = [
```
maxime-rainville, line 11: This bit of code needs to be copied into any class that implements this trait.

src/Attributable.php
```diff
@@ -0,0 +1,112 @@
...
+        $attributes = $this->getAttributes();
+
+        if (isset($attributes[$name])) {
+            return $attributes[$name];
+        }
+
+        return null;
+    }
+
+    protected abstract function getDefaultAttributes(): array;
```
maxime-rainville, line 36: This would require any any class that implements this trait to provide a default list of attributes to get us started.

src/Attributable.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @return array
+     */
+    public function getAttributes()
+    {
+        $defaultAttributes = $this->getDefaultAttributes();
+
+        $attributes = array_merge($defaultAttributes, $this->attributes);
+
+        if (method_exists($this, 'extend')) {
```
maxime-rainville, line 51: It would be fair to assume that which ever class implements this must also implements Extensible, but by have a function check here, we don't force it.

src/ImageManipulation.php
```diff
@@ -981,6 +984,8 @@ public function manipulate($variant, $callback)
             $file->setImageBackend($manipulationResult);
         }
 
+        $file->setMutableAttributes($this->attributes);
```
maxime-rainville, line 987: This bits just make sure that when we create variant of our file we pass along any attribute attach to the file.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
         }
         return $value;
     }
+
+    private function copyImageBackend()
```
maxime-rainville, line 1066: This bit was mostly extracted out of `manipulate`. I tried calling `manipulate` directly but it just does too many thing to work well.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+                'Hash' => $this->getHash(),
+                'Variant' => $this->getVariant()
+            ]
+        );
+
+        $file->setImageBackend($this->getImageBackend());
+        return $file->setOriginal($this);
+    }
+
+    public function setAttribute($name, $value)
```
maxime-rainville, line 1083: This overrides the `setAttribute` method for our new trait. The reason I do this, is that we need our Image to be "Immutable" when we manipulate it. This allows us to return a new copy each time.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+        $file = $this->copyImageBackend();
+        $file->setMutableAttributes(array_merge(
+            $this->attributes,
+            [$name => $value]
+        ));
+
+        return $file;
+    }
+
+    public function setMutableAttributes($attributes)
```
maxime-rainville, line 1094: This allows us to bypass our attributes Immutability. It's admittedly a bit hacky. Maybe it should be called something like `initAttributes` to make it's purpose clearer and to make sure people understand it should only be called once.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+    {
+        $this->attributes = $attributes;
+    }
+
+    protected function getDefaultAttributes(): array
+    {
+        if (!$this->getIsImage()) {
```
maxime-rainville, line 1101: I'm assuming that Images will always have a Height and a Width here.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+            return [];
+        }
+
+        $attributes = [
+            'width' => $this->getWidth(),
```
maxime-rainville, line 1106: Maybe those should only be added to the HTML tag if lazy loading is enabled. Not sure.

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+            $lazyLoad = strtolower($lazyLoad);
+        }
+
+        switch ($lazyLoad) {
+            case false:
+            case 0:
+            case '0':
+            case 'false':
+            case 'eager':
+                return $this->setAttribute('loading', 'eager');
```
maxime-rainville, line 1146: A side affect of this is that we will always add `loading="eager"` when we disable lazy loading on a specific image.

If we thing this is a problem, I can probably find a way to not output the attribute at all.

templates/DBFile_image.ss
```diff
@@ -1 +1 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
+<img src="$URL.ATT" alt="$Title.ATT" $AttributesHTML />
```
maxime-rainville, line 1: Maybe we want all the attributes to be rendered with `$AttributesHTML`. Not sure yet.

maxime-rainville (commented):
This is a work in progress. It's not ready for review, but I could use some high level thought on the approach.

src/Attributable.php
```diff
@@ -0,0 +1,112 @@
...
+    protected abstract function getDefaultAttributes(): array;
+
+    /**
+     * Allows customization through an 'updateAttributes' hook on the base class.
+     * Existing attributes are passed in as the first argument and can be manipulated,
+     * but any attributes added through a subclass implementation won't be included.
+     *
+     * @return array
+     */
+    public function getAttributes()
```
emteknetnz, line 45: Is, the intention here to replace things like this? https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/CheckboxField.php#L29

maxime-rainville: My initial intention would be to replace only the `FormField` one. All the other other ones are just overriding the FormField one.

I think there's a couple other ViewableData that reimplements this Attribute pattern. 

Possible next step would be:
1. Delete this trait and copy all the logic straight into ImageManipulation.
2. Move this trait and move it to framework. Initally only the ImageManipulation trait would reference it.
3. Step 2 + update FormFiled to use it, but nothing else.

Don't have any preference at this stage. Just want to get this over the line in time.

emteknetnz: At this stage I'd say option 1. since that fixes the immediate need

src/Attributable.php
emteknetnz: When we go with option 1 there's probably extra code like this that's currently in the trait that can be removed

src/ImageManipulation.php
```diff
@@ -46,6 +46,9 @@
  */
 trait ImageManipulation
 {
+    use Attributable;
```
emteknetnz, line 49: trait in a trait :-p

src/ImageManipulation.php
emteknetnz: does it still work calling `manipulate()` though? If so I'd be inclined to still use it.  Just looking at a comment within `mainpulate()`

```
        // Pass the manipulated image backend down to the resampled image - this allows chained manipulations
        // without having to re-load the image resource from the manipulated file written to disk
```

Possible but not calling manipulate we'll create weird edge case bugs?  e.g. setWidth(300).setLazyLoad(false) vs setLazyLoad(false).setWidth(300) will behave differently somehow?  I just don't know

src/ImageManipulation.php
emteknetnz: I'm kind of surprised we need to do this given File objects are already chaining within templates?

src/ImageManipulation.php
emteknetnz: This method is on ImageManipulation, it's probably fair to assume that we're dealing with an Image at this point?

src/Attributable.php
michalkleiner: If we stick with the trait, I'd rename it to something like HTMLAttributes or similar; Attributable sounds more like an interface suggesting the class using it can be attributed to something/someone, which probably isn't the case here.

src/ImageManipulation.php
emteknetnz: I'd say we should just add them regardless

This might cause some images on existing sites that have their dimensions specified via CSS (or maybe not, I'm not sure on whether attributes or CSS (or width and/or height specified in style attributes) has priority

However the vast majority of images are now going to now be lazy loaded by default, which means they're getting width + height attributes added

So excluding width + height on eager loaded only images doesn't really make much sense in this context

src/ImageManipulation.php
emteknetnz: Thinking so far has been to omit the loading attribute entirely when eager loading.  Seems like all the online articles talking about recommend that it's omitted

If we changes this to a `removeAttribute('loading)` call would everything (immutable/chainable) work as it should?

src/ImageManipulation.php
```diff
@@ -1057,4 +1062,98 @@ protected function castDimension($value, $dimension)
...
+
+        switch ($lazyLoad) {
+            case false:
+            case 0:
+            case '0':
+            case 'false':
+            case 'eager':
+                return $this->setAttribute('loading', 'eager');
+
+            case true:
```
emteknetnz, line 1148: +1 for adding the truthy conditions here

just double checking - is a switch statement a strict === check?

src/Storage/DBFile.php
```diff
@@ -113,7 +113,9 @@ protected function getStore()
         'MimeType' => 'Varchar',
         'String' => 'Text',
         'Tag' => 'HTMLFragment',
-        'Size' => 'Varchar'
+        'Size' => 'Varchar',
+        'AttributesHTML' => 'HTMLFragment',
+        'getAttributesHTML' => 'HTMLFragment',
```
emteknetnz, line 118: do we need to separately specify getAttributesHTML? we don't specify getTag, we only have Tag

templates/DBFile_image.ss
emteknetnz: I'd say all attributes in this case.  Should either do all within PHP or all within the template, not both IMO

emteknetnz (changes requested):
I think the trait is a little overly ambitions :-)  I'm not entirely opposed to the idea of it, it does seem like it may be a good opportunity to reduce code duplication on FormField subclasses, however I don't think it makes sense given the context of this image lazy load work

Assuming we move the code from the trait to ImageManipulation, they what you have looks seems necessary to support scenarios such as putting the same $Img twice on the same temlate, with different prior manipulations such as different SetWidth(int) values, with different values for LazyLoad()

Needs unit tests, presumably this should be doable with something like renderWith()?

src/Attributable.php
maxime-rainville: Yeah ... not a big fan of Attributable name. I pick it up quickly just to get me unblock. Looks like we'll do a separate PR for this anyway.

src/ImageManipulation.php
maxime-rainville: Inception! I don't think there's anything inherently wrong with it and we actually have a couple other places where we do this already:
- Extensible
- FileUploadReceiver

src/ImageManipulation.php
maxime-rainville: `manipulate` will only call the callback if the variant does not exist. Since we're not actually creating a variant. This means there's a good chance our callback will never be called.

I'm pretty sure `setWidth(300).setLazyLoad(false)` and `setLazyLoad(false).setWidth(300)` will end up in the same place, but I'll double check.

src/ImageManipulation.php
maxime-rainville: The problem is that when I create a new DBFile instance, I don't have a way to copy the attributes to the new instance:
- I can't call `setAttribute` on it because I've overriwritten that method already to be immutable (I would just end up in an infinite recursion)
- I can't called the protected `$attributes` class property because `Image` and `DBFile` are not the same class and won't have access to each others' protected members.

src/ImageManipulation.php
maxime-rainville: `ImageManipulation` is applied to file ... which is very confusing. All those methods are available on `File`, they just return null in most cases.

src/ImageManipulation.php
maxime-rainville: I'll get it removed. Just need to tweak the `getAttributes` method a bit.

src/ImageManipulation.php
maxime-rainville: Good question, I'll have a look.

src/Storage/DBFile.php
maxime-rainville: That bit got copied over from `FormField`. You can still technically call `$getAttributesHTML` in your SS template.

Arguably we should have a `getTag` entry as well. Don't think it makes much difference either way. Don't think it makes much difference either way. I expect the vast majority of devs will use the method without the "get".

src/ImageManipulation.php
maxime-rainville: I've added a bit of logic so we don't output `loading="eager"`

src/ImageManipulation.php
maxime-rainville: `switch` does a loose comparison so I had to scrap this approach.

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+            'DBFile',
+            [
+                'Filename' => $this->getFilename(),
+                'Hash' => $this->getHash(),
+                'Variant' => $this->getVariant()
+            ]
+        );
+
+        $backend = $this->getImageBackend();
+        if ($backend) {
```
maxime-rainville, line 1091: Regular files don't have an image backend

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+            ];
+
+            if ($this->IsLazyLoaded()) {
+                $attributes['loading'] = 'lazy';
+            }
+        }
+
+        $attributes = array_merge($attributes, $this->attributes);
+
+        if (isset($attributes['loading']) && $attributes['loading'] === 'eager') {
```
maxime-rainville, line 1168: This bit unset is here to make sure we don't have a `loading="eager"` snippet in our HTML

templates/DBFile_image.ss
```diff
@@ -1 +1 @@
-<img src="$URL.ATT" alt="$Title.ATT" />
```
maxime-rainville, line 1: There's a similar template for regular files that could be updated for this as well. Allowing regular files to have attributes as well could be interesting.

Could be a follow up issue.

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+     * but any attributes added through a subclass implementation won't be included.
+     *
+     * @return array
+     */
+    public function getAttributes()
+    {
+        $attributes = [];
+        if ($this->getIsImage()) {
+            $attributes = [
+                'width' => $this->getWidth(),
```
emteknetnz, line 1155: What about other attributes such as `class` ?

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+    }
+
+    /**
+     * Allows customization through an 'updateAttributes' hook on the base class.
+     * Existing attributes are passed in as the first argument and can be manipulated,
+     * but any attributes added through a subclass implementation won't be included.
+     *
+     * @return array
+     */
+    public function getAttributes()
```
emteknetnz, line 1150: Add php7 return types to all the new API e.g `public function getAttributes(): array`

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+     *
+     * @param array $attributes
+     * @return string
+     */
+    public function getAttributesHTML($attributes = null)
+    {
+        $exclude = null;
+
+        if (is_string($attributes)) {
+            $exclude = func_get_args();
```
emteknetnz, line 1190: Is it possible to not use `func_get_args()` if at all possible and use a proper function signature?  Or is it done this way to be consistent with how it's done elsewhere is Silverstripe?

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+     * If at least one argument is passed as a string, all arguments act as excludes, by name.
+     *
+     * @param array $attributes
+     * @return string
+     */
+    public function getAttributesHTML($attributes = null)
+    {
+        $exclude = null;
+
+        if (is_string($attributes)) {
```
emteknetnz, line 1189: dockblock function signature says $attributes is an array, yet we're testing if it's a string?

I'm really not a fan of mixed or variable length arguments

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+        }
+
+        if (!$attributes || is_string($attributes)) {
+            $attributes = $this->getAttributes();
+        }
+
+        $attributes = (array) $attributes;
+
+        $attributes = array_filter($attributes, function ($v) {
+            return ($v || $v === 0 || $v === '0');
```
emteknetnz, line 1200: This reads pretty weird.

`return !in_array($v, ['', null], true) ` might be clearer

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+        }
+
+        // Create markup
+        $parts = [];
+
+        foreach ($attributes as $name => $value) {
+            if ($value === true) {
+                $value = $name;
+            } else {
+                if (is_scalar($value)) {
```
emteknetnz, line 1217: I'd be included to just have `$value = (string) $value;` regardless of what $type value is

If you're passing in something like an object for whatever reason, it's probably a mistake and I see no reason to turn it into a json blob so that it displays "nicely".  If you really wanted to pass in json, you can just encode it before adding it as an attribute

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+                $attributes,
+                array_flip($exclude)
+            );
+        }
+
+        // Create markup
+        $parts = [];
+
+        foreach ($attributes as $name => $value) {
+            if ($value === true) {
```
emteknetnz, line 1214: is this for the `checked="checked"` type of thing?  These days you can just omit the value? e.g. `<input checked>` https://www.w3schools.com/tags/att_input_checked.asp

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+
+        return implode(' ', $parts);
+    }
+
+    /**
+     * Determine whether the image should be lazy loaded
+     *
+     * @return bool
+     */
+    public function IsLazyLoaded(): bool
```
emteknetnz, line 1235: should be isLazyLoaded or getIsLazyLoaded

I don't think this will ever be called from a template

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+        if (!Image::getLazyLoadingEnabled() || !$this->getIsImage()) {
+            return $this;
+        }
+
+        if (is_string($lazyLoad)) {
+            $lazyLoad = strtolower($lazyLoad);
+        }
+
+        $equivalence = [
+            'eager' => [false, 0, '0', 'false', 'eager'],
```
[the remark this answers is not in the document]
emteknetnz: I'd prefer not to have 'eager' included here

.LazyLoad(false) makes sense, while .LazyLoad(eager) isn't quite right?  i.e. The function toggles loading="lazy", it is not to specify the value of the loading attribute

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+            'lazy' => [true, 1, '1', 'true', 'lazy'],
```
emteknetnz, line 1260: So, I don't think .LazyLoad(true) will ever actually do anything due to design design to lazy load by default and not make the default value configurable

The only configurable change a dev can make it to globaly disable this functionality

If they do this, then it will fail the `Image::getLazyLoadingEnabled()` conditional and never reach this point

I manually validated this locally testing locally with lazy loading disabled and have a template with $MyImage.LazyLoad(true) - no loading attribute was added

We should just remove this code, and any related code to allows a dev to "add" loading="lazy"

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+
+        return $file;
+    }
+
+    /**
+     * Initialise the attirbutes on this object. Should only be called imiditely after intialising the object.
+     * @param array $attributes
+     * @internal
+     */
+    public function initAttributes(array $attributes)
```
emteknetnz, line 1121: Does this need to be a public function?  Seems like this should be a private?

src/ImageManipulation.php
emteknetnz: I'm a bit concerned this method name may be a bit too likely to collide with something else, like some sort of future HTML wrapper

src/ImageManipulation.php
```diff
@@ -1057,4 +1069,204 @@ protected function castDimension($value, $dimension)
...
+        }
+        return false;
+    }
+
+    /**
+     * Set the lazy loading state for this Image
+     * @param mixed $lazyLoad
+     * @return AssetContainer
+     */
+    public function LazyLoad($lazyLoad): AssetContainer
```
emteknetnz, line 1248: The particular method name is consistent with the other methods called from templates in this file such as `ScaleWidth()`

emteknetnz (changes requested):
Look good, I've tested locally and things worked as expected, I'll put testing details in a comment below.   This is in a mergable state, though there's some time left for some improvements 

Is it possible to add a unit/functional test where we have essentially a TestOnly template?  Something that includes $MyImage.LazyLoad(false), $MyImage.ScaleWidth(300).LazyLoad(false) etc?  Reason I asked is that original PR passed these unit tests, but failed when manually tested against a template file.

emteknetnz:
Testing details for first peer-review:

**global lazy-load on (default)**

PageController (FileID 1 is an uploaded image)

```php
public function MyImage()
{
    return SilverStripe\Assets\File::get()->byID(1);
}
```

Template:

```html
<div class="global-lazy-load-on">
    $MyImage
    <hr>
    $MyImage.LazyLoad(false)
    <hr>
    $MyImage.ScaleWidth(300)
    <hr>
    $MyImage.ScaleWidth(300).LazyLoad(false)
    <hr>
    $MyImage.LazyLoad(false).ScaleWidth(300)
</div>
```

Rendered HTML:

```html
<div class="global-lazy-load-on">
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02.jpg" loading="lazy" width="2048" height="1152">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02.jpg" width="2048" height="1152">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" loading="lazy" width="300" height="169">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" width="300" height="169">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" width="300" height="169">
</div>
```


**global lazy-load off -- missing loading="lazy" attributes**

```html
<div class="global-lazy-load-off">
    $MyImage
    <hr>
    $MyImage.LazyLoad(true)
    <hr>
    $MyImage.ScaleWidth(300)
    <hr>
    $MyImage.ScaleWidth(300).LazyLoad(true)
    <hr>
    $MyImage.LazyLoad(true).ScaleWidth(300)
</div>
```

```html
<div class="global-lazy-load-off">
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02.jpg" width="2048" height="1152">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02.jpg" width="2048" height="1152">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" width="300" height="169">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" width="300" height="169">
    <hr>
    <img alt="sample robotic arms 02" src="/assets/sample-robotic-arms-02__ScaleWidthWzMwMF0.jpg" width="300" height="169">
</div>
```

src/ImageManipulation.php
maxime-rainville: There's no getters/setters for those. A dev could go in and directly call those if they want to. Those extra attributes will be merged-in into `$attributes`.

Could be a nice follow up task to add some more methodsfor common things like classes or Alt text, etc.

src/ImageManipulation.php
maxime-rainville: This got copied from FormField. I still hope to do a follow up task where we move all of that logic to a dedicated trait. So there's value in keeping the same signature.

src/ImageManipulation.php
maxime-rainville: That entire method was copied as-is from `FormField`. That would be part of that eventual dedicated "attributable" trait.

src/ImageManipulation.php
maxime-rainville: Same deal as thing as https://github.com/silverstripe/silverstripe-assets/pull/458#discussion_r665731990

src/ImageManipulation.php
maxime-rainville: Same deal as https://github.com/silverstripe/silverstripe-assets/pull/458#discussion_r665731990

src/ImageManipulation.php
maxime-rainville: I'll double check to be sure, but I think a trait cannot access its own private members on an other class.

src/ImageManipulation.php
maxime-rainville: We could look at tweaking this bit if/when we do global trait

src/ImageManipulation.php
maxime-rainville: Many DEV prefers to write their own `` tag in their SS files. We need to give them an option to retrieve the value in those scenarios.

src/ImageManipulation.php
maxime-rainville: Same deal as thing as https://github.com/silverstripe/silverstripe-assets/pull/458#discussion_r665731990

src/ImageManipulation.php
maxime-rainville: Did you mean to write "inconsistent" or are you formally noting your agreement with the method name.

Just double-checking :sweat_smile:

src/ImageManipulation.php
maxime-rainville: If you can disable something, then you need a way to re-enable it later on. e.g.: A developer could have a bit of logic that start by disabling lazy loading on an image and selectively turns it back on later on.

src/ImageManipulation.php
maxime-rainville: Just tried running the unit test locally with private/protected and it fails both time. Basically, `Image` can't called a private/protected method on 'DBFile` even if they share a trait.

I marked the method `@internal` to communicate to third party devs they shouldn't be calling this method directly.

Technically, this will only ever be called on DBFile and `Image` doesn't need it. We could move the method into `DBFile` if you think that makes more sense.

src/ImageManipulation.php
emteknetnz: I'm noting agreement.  No change required.

emteknetnz (approved):

src/Attributable.php
maxime-rainville: Got a follow up PR and issue https://github.com/silverstripe/silverstripe-framework/pull/10033

PR #460, opened 2021-07-20
API allow chaining for Upload_Validator

wilr:


emteknetnz (changes requested):
Seems fine, matches what happens often happens elsewhere with `set` methods e.g.
- https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/Validator.php#L150
- https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/GridField/GridFieldExportButton.php#L283

Could you add a simple unit test that chains everything together and asserts that $this is returned.

wilr:
> Could you add a simple unit test that chains everything together and asserts that $this is returned.

Test added

emteknetnz:
@maxime-rainville you OK with changing the pseudo API signature of what's functionally changing the return type of `:void` to `:Upload_Validator`?

I think this is fine for a minor since nothing will be relying on a return value right now, and I don't think think there's really anything else we'd return.  Possible exception may be `clearErrors()` since it's not a `set` function, though I can't really imagine why we'd every return a `bool` or `array` for this method

dhensby (approved):

GuySartorelli:
Retargetted to `2` and rebased. If CI passes fine I'll be happy to merge this - it's not BC breaking by our current standards.

PR #465, opened 2021-08-18
API Create extension hook for updating session grant on files

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-sharedraftcontent/issues/132


michalkleiner:
Looking reasonable from my perspective as well, just one minor ask if we could add an inline comment explaining why we explicitly pass `null` as the file record so that we don't scratch our heads in future.

PR #470, opened 2021-10-20
Remove appended new line after img tag

axllent:
Remove the new line appended to `$Image` in templates

eg:

```
<a href="/">$Image</a>
```
currently results in
```
<a href="/"><img .... />
</a>
```


dhensby:
Interesting, maybe we should trip template rendered strings because having new-lines at the end of files is pretty common/normal so this is likely a systemic issue.

axllent:
Agreed, that would be a better {and safer) solution all round.

michalkleiner:
+1 for trimming template-rendered strings

dhensby:
I presume if this code is changed to the that `file_get_contents()` call is `rtrim()` for new lines, we should be good?

https://github.com/silverstripe/silverstripe-framework/blob/4/src/View/SSTemplateParser.peg#L1338-L1341

michalkleiner:
Yep, that sounds like the safest option, though I wonder whether we want to do a general `trim` on all whitespace characters if we're touching this. Do we need possible leading whitespace chars preserved for any reason?

dhensby:
I think leading whitespace characters in templates are much less likely to be there by accident, whereas lots of IDEs and `.editorconfig` setups will mandate new lines at the end of files (which the SS repos do), which is why I'm cautious about this PR - anyone editing a template will add accidental new lines by mistake.

There is *some* risk associated with trimming whitespace from the end of files, for example if the content is in a `<pre>` block, where whitespace matters... but that feels less likely and less of an issue than *adding* unexpected whitespace to inline elements like we are seeing here.

kinglozzer:
I knew I’d seen some discussion about this before! https://github.com/silverstripe/silverstripe-framework/issues/4896

dhensby:
Oh the rabbit hole of templating. I was just thinking about what joy it would be to rewrite the whole thing 😅

GuySartorelli:
Closing as it sounds like there are preferred alternatives solutions.

axllent:
Unless I'm missing something, I'm not sure a 6-year-old (unimplemented) RFC can be considered a valid alternative solution.

GuySartorelli:
There was discussion above about trimming rendered templates globally, which commenters above (including yourself, stating "that would be a better {and safer) solution all round") agreed would be preferable to this PR's solution. I was referring to that more so than the RFC.

PR #471, opened 2021-11-21
API Switch to using AttributesHTML on ImageManipulation trait

maxime-rainville:
Follow up to https://github.com/silverstripe/silverstripe-framework/pull/10033

# Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10032

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -14,6 +14,7 @@ use SilverStripe\Core\Injector\InjectorNotFoundException;
 use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
+use SilverStripe\View\AttributesHTML;
 use SilverStripe\View\HTML;
 
 /**
@@ -49,6 +50,8 @@ use SilverStripe\View\HTML;
  */
 trait ImageManipulation
 {
+    use AttributesHTML;
+
     /**
      * @var Image_Backend
      */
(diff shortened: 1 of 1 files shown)


src/ImageManipulation.php
```diff
@@ -1104,6 +1101,7 @@ private function copyImageBackend(): AssetContainer
      */
     public function setAttribute($name, $value)
```
maxime-rainville, line 1102: The `ImageManipulation::setAttirbute()` has some custom logic not present in the `AttributesHTML` equivalent.

src/ImageManipulation.php
```diff
@@ -1162,69 +1144,15 @@ public function getAttributes()
...
-                if (is_scalar($value)) {
-                    $value = (string) $value;
-                } else {
-                    $value = json_encode($value);
-                }
-            }
-
-            $parts[] = sprintf('%s="%s"', Convert::raw2att($name), Convert::raw2att($value));
+        // We don't want an loading=eager attribute because that's the browser default
+        if (isset($attributes['loading']) && $attributes['loading'] === 'eager') {
```
maxime-rainville, line 1153: This needs to happen after the `$attributes = array_merge($attributes, $this->attributes);` call. This is why its in `updateAttributes`.

emteknetnz (approved):

kinglozzer:
@maxime-rainville should be update the version constraint in `composer.json`? We’re now using a new API only present in framework 4.10.x. Not sure what our policy is on this!

PR #476, opened 2022-02-07
NEW WebP support in CMS

maxime-rainville:
This PR adds support for WebP images. The CMS will essentially treat them like JPEG and PNG images.

Please note:
- I tested with GD ... didn't try Imagick
- This doesn't address the issue that some browsers (aka IE11) don't support WebP images. The CMS doesn't provide any feedback to content authors about the image potentially not working in older browsers.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -207,7 +207,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         'docx', 'dotx', 'flv', 'gif', 'gpx', 'gz', 'hqx', 'ico', 'jpeg', 'jpg', 'kml',
         'm4a', 'm4v', 'mid', 'midi', 'mkv', 'mov', 'mp3', 'mp4', 'mpa', 'mpeg', 'mpg', 'ogg', 'ogv', 'pages',
         'pcx', 'pdf', 'png', 'pps', 'ppt', 'pptx', 'potx', 'ra', 'ram', 'rm', 'rtf', 'sit', 'sitx',
-        'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'zip',
+        'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'webp', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'zip',
         'zipx',
     ];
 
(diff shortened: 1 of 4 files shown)


src/Storage/DBFile.php
```diff
@@ -64,6 +64,7 @@ class DBFile extends DBComposite implements AssetContainer, Thumbnail
         'image/x-icon',
         'image/vnd.microsoft.icon',
         'image/vnd.adobe.photoshop',
+        'image/webp'
```
michalkleiner, line 67:
```suggestion
        'image/webp',
```

src/File.php
```diff
@@ -197,7 +197,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         'm4a', 'm4v', 'mid', 'midi', 'mkv', 'mov', 'mp3', 'mp4', 'mpa', 'mpeg', 'mpg', 'ogg', 'ogv', 'pages',
         'pcx', 'pdf', 'png', 'pps', 'ppt', 'pptx', 'potx', 'ra', 'ram', 'rm', 'rtf', 'sit', 'sitx',
         'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'zip',
```
michalkleiner, line 199:
```suggestion
        'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'webp', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'zip',
```

src/File.php
```diff
@@ -197,7 +197,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
...
-        'zipx',
+        'zipx', 'webp'
```
michalkleiner, line 200:
```suggestion
        'zipx',
```

michalkleiner: the list seems to be in alphabetical order

michalkleiner:
Can we extend or adjust an existing test adding the webp support there, too?

maxime-rainville:
If we're concerned about content authors accidentally adding WepP image to site targetting IE11 users, we could just put this in a separate module which would allow project to opt-in to webp support.

Looking at the statcounter, [IE11 has a global market share right now of 0.45%](https://gs.statcounter.com/browser-market-share). And it's dropping rapidly ... it was 0.81% a year ago.

I'm thinking, most sites that still want to cater to IE11 users will probably be well aware not to use WebP images any way. We can probably address this with some user docs.

So it's basically making life somewhat better for the vast majority of project with a tiny risk that someone will accidentally publish a WebP image to a site that still support IE11.

@silverstripe/core-team Yeah :+1:  Nay :-1:  on merging this PR?

sminnee:
IMO add support by default, and provide a config flag so that developers can turn it off on sites that require IE11 support

maxime-rainville:
I tested to make sure this works with Imagick, which it does. I also validated that you can block webp file upload by adding snippet to your YML config:
```yml
SilverStripe\Assets\File:
  allowed_extensions:
    webp: false
```

Found a related bug here: https://github.com/silverstripe/silverstripe-asset-admin/pull/1256

michalkleiner (approved):

PR #477, opened 2022-02-08
BUG Don't auto grant session access when resampling images

maxime-rainville:
We considered treating this as a security issue but decided that the amount of sensitive information in an image does not warrant it.

In some context, the CMS will grant your session permission to view a file irrespective of if you have access to view it. The specific thing we are trying to address here is being able to view a restricted image if it's added to a campaign.

Previous places where we addressed this included an option to [allow automatic  session grant](https://github.com/silverstripe/silverstripe-assets/blob/1/src/Shortcodes/FileShortcodeProvider.php#L52) via a config. I don't think we need to do this anymore since tho AssetStore now automatically grant you access to view files.

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -303,10 +303,9 @@ trait ImageManipulation
 
         // Only update if resampled file is a smaller file size
         if ($resampled->getAbsoluteSize() < $this->getAbsoluteSize()) {
-            $url = $resampled->getURL();
+            $url = $resampled->getURL(false);
         }
     }
-
 
     /**
      * Generate a resized copy of this image with the given width & height.
@@ -705,7 +704,7 @@ trait ImageManipulation
     {
         $thumbnail = $this->Thumbnail($width, $height);
         if ($thumbnail) {
-            return $thumbnail->getURL();
+            return $thumbnail->getURL(false);
         }
         return $this->getIcon();
     }
@@ -1130,7 +1129,7 @@ trait ImageManipulation
                 'width' => $this->getWidth(),
                 'height' => $this->getHeight(),
                 'alt' => $this->getTitle(),
-                'src' => $this->getURL()
+                'src' => $this->getURL(false)
             ];
 
             if ($this->IsLazyLoaded()) {
--- a/tests/php/ImageManipulationTest.php
+++ b/tests/php/ImageManipulationTest.php
@@ -427,4 +427,28 @@ class ImageManipulationTest extends SapphireTest
             trim($image->renderWith(SSViewer::fromString($template)))
         );
     }
+
+    public function testThumbnailURL()
+    {
+        $img = $this->objFromFixture(Image::class, 'imageWithTitle');
+
+        // File needs to be in draft and users need to be anonymous to test the access
+        $this->logOut();
+        $img->doUnpublish();
+
+        $fileUrl = 'folder/444065542b/test-image__FillWzEwLDEwXQ.png';
+
+        $this->assertEquals(
+            '/assets/' . $fileUrl,
+            $img->ThumbnailURL(10, 10),
+            'Thumbnail URL is correct'
+        );
+
+        /** @var AssetStore assetStore */
+        $assetStore = Injector::inst()->get(AssetStore::class);
+        $this->assertFalse(
+            $assetStore->isGranted($fileUrl),
+            'Current user should not automatically be granted access to view thumbnail'
+        );
+    }
 }
--- a/tests/php/ImageTest.php
+++ b/tests/php/ImageTest.php
@@ -231,6 +231,32 @@ abstract class ImageTest extends SapphireTest
         $this->assertNotEquals($imageHQ->getURL(), $imageHQR->getSourceURL(), 'Path to the original image file was returned by getURL()');
     }
 
+    /**
+     * Tests that a URL to a resampled image is provided when force_resample is
+     * set to true, if the resampled file is smaller than the original.
+     */
+    public function testUpdateURL()
+    {
+        // Test resampled file is served when force_resample = true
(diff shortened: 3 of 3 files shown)


maxime-rainville (commented):
Note tht you can still an entry for the image in campaign admin, but you won't be able to see the actual thumbnail.

image

maxime-rainville:
Not 100% sure if this is worth covering with Unit test ... I can add some if we think it is.

src/ImageManipulation.php
```diff
@@ -303,7 +303,7 @@ public function updateURL(&$url)
 
         // Only update if resampled file is a smaller file size
         if ($resampled->getAbsoluteSize() < $this->getAbsoluteSize()) {
-            $url = $resampled->getURL();
+            $url = $resampled->getURL(false);
```
emteknetnz, line 306: This will probably cover a lot of regular images on pages won't it?  Won't this break the existing functionality of automatically granting view permissions on draft images on draft pages that a content editor has access to?

emteknetnz (changes requested):

src/ImageManipulation.php
maxime-rainville: We're not meant to be automatically granting access to draft images on draft page ... we had a security issue: https://github.com/silverstripe/silverstripe-assets/commit/fddf889917c4e58d32a3e6f476bddaf3fa595e41

If this line creates this ability, than it's bug. In any case, this method as zero context on who is using the file, so it's the wrong place to make the call about whether we want to grant access to the file.

emteknetnz: We've got it as a configurable opt-in to grant on FileShortCodeProvider, which is inherited by ImageShortCodeProvider - https://github.com/silverstripe/silverstripe-assets/blob/1/src/Shortcodes/FileShortcodeProvider.php#L52

Wondering if there's any value in doing something similar here

maxime-rainville: My instinct would be no ... the methods here are generic. They are unaware of the wider context they are being called in. 

Also, when we implemented the "configurable session grant" option, session grant was the only way of accessing protected files. Since then, we updated the AssetStore to dynamically check if the current user has access to the file on the request. So the need to make this configurable has been reduce.

emteknetnz (changes requested):
Unit tests are probably worth it to ensure someone doesn't revert this in the future in order to fix something else

maxime-rainville:
Just added some unit test ... I also specifically tested that they would have failed with the old logic

tests/php/ImageTest.php
```diff
@@ -231,6 +231,33 @@ public function testForceResample()
         $this->assertNotEquals($imageHQ->getURL(), $imageHQR->getSourceURL(), 'Path to the original image file was returned by getURL()');
     }
 
+        /**
```
emteknetnz, line 234: Indentation

emteknetnz, line 234: Indentation
```suggestion
        /**
```

tests/php/ImageTest.php
```diff
@@ -231,6 +231,33 @@ public function testForceResample()
...
+     * Tests that a URL to a resampled image is provided when force_resample is
+     * set to true, if the resampled file is smaller than the original.
+     */
+    public function testUpdateURL()
+    {
+
```
emteknetnz, line 240:
```suggestion
```

tests/php/ImageTest.php
```diff
@@ -231,6 +231,33 @@ public function testForceResample()
...
+        $imageHQ->doUnpublish();
+
+        $url = '';
+        $imageHQ->updateURL($url);
+
+        /** @var AssetStore assetStore */
+        $assetStore = Injector::inst()->get(AssetStore::class);
+        $this->assertFalse(
+            $assetStore->isGranted('folder/a870de278b/test-image-high-quality__Resampled.jpg'),
+            'Current user is not automatically granted access to resampled image'
```
emteknetnz, line 257:
```suggestion
            'Logged out user was granted access to draft resampled image'
```

tests/php/ImageManipulationTest.php
```diff
@@ -427,4 +427,28 @@ public function testRender(string $template, string $expected)
...
+            '/assets/' . $fileUrl,
+            $img->ThumbnailURL(10, 10),
+            'Thumbnail URL is correct'
+        );
+
+        /** @var AssetStore assetStore */
+        $assetStore = Injector::inst()->get(AssetStore::class);
+        $this->assertFalse(
+            $assetStore->isGranted($fileUrl),
+            'Current user is not automatically granted access'
```
emteknetnz, line 451:
```suggestion
            'Logged out user was granted access to draft resampled image'
```

emteknetnz (changes requested):
Looks good, some trivial changes to make to failed test messaging and whitespace changes

tests/php/ImageManipulationTest.php
maxime-rainville: I prefer assertion message to describe what is being asserted, rather than what has failed. 

But I can see the message here could be confusing if you think `Current user is not automatically granted access` describes what has failed.

What about this instead?
 ```suggestion
            'Current user should not automatically be granted access to view thumbnail'
```

File: tests/php/ImageTest.php (hunk 4 above)
Comment (maxime-rainville) [reply tests/php/ImageTest.php:256]: What about this?
```suggestion
            'Current user should not automatically be granted access to resampled image'
```

emteknetnz (approved):

emteknetnz:
Merge on green

emteknetnz:
PHPCS issue https://app.travis-ci.com/github/silverstripe/silverstripe-assets/jobs/559881306#L1336

maxime-rainville:
Just fixed the linting issue.

emteknetnz (approved):

PR #479, opened 2022-02-19
ENH Add config to optionally disable image resampling

leochenftw:
This is used in https://github.com/silverstripe/silverstripe-asset-admin/pull/1252

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -118,14 +118,15 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     private static $plural_name = "Files";
     
     /**
-     * define a default value for no_resample config
-     * this will be used to decide if a File subclass needs
-     * to be resampled. This is useful when using AWS S3 module
+     * Control whether images in the admin will be resampled
+     *
+     * Can be turned off e.g. when using an external resampling service
+     * or a backend that may not support resampling.
      *
      * @config
      * @var bool
      */
-    private static $no_resample = false;
+    private static $resample_images = true;
 
     /**
      * Anyone with CMS access can view draft files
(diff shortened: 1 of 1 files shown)


src/File.php
```diff
@@ -116,6 +116,16 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
...
+    
+    /**
+     * define a default value for no_resample config
+     * this will be used to decide if a File subclass needs
+     * to be resampled. This is useful when using AWS S3 module
+     *
+     * @config
+     * @var bool
+     */
+    private static $no_resample = false;
```
michalkleiner, line 128:
```suggestion
    
    /**
     * Control whether images in the admin will be resampled
     *
     * Can be turned off e.g. when using an external resampling service
     * or a backend that may not support resampling.
     *
     * @config
     * @var bool
     */
    private static $resample_images = true;
```

michalkleiner:
Thank you for raising the PR, @leochenftw. I saw the linked PR as well and I think it makes sense to allow for this configuration to be added, even on the `File` class despite there being `Image` class as well (which doesn't do much on its own).
I've suggested to reverse the name into a positive form (to reflect what the default is) and that the intention is to allow to optionally turn the resampling off.
If you could force-push an update and also change the commit message to read "ENH Add config to optionally disable image resampling" I'd be happy to merge that.

Thanks!

michalkleiner (changes requested):

michalkleiner:
Thanks for the tweaks @leochenftw. Since this is an enhancement, the PR needs to be based off and target the `1` branch. 
Can you please also squash the commits so that it's just a single commit with the message I suggested? I can squash the commits at our end, but I can't rebase the PR.

leochenftw:
thanks, @michalkleiner , I've rebased it to `1`

michalkleiner (approved):

PR #481, opened 2022-03-08
DEP Update minimum version of intervention/image

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10253


maxime-rainville (approved):
Not against this, but it's worth pointing out this is not strictly needed since `^2.3` would have allowed the installation of 2.6 and we're not using any APIs that got introduced in 2.6.

PR #483, opened 2022-04-04
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/src/Dev/Tasks/SecureAssetsMigrationHelper.php
+++ b/src/Dev/Tasks/SecureAssetsMigrationHelper.php
@@ -138,14 +138,14 @@ class SecureAssetsMigrationHelper
     public function htaccessMatch($content)
     {
         $regexes = $this->htaccessRegexes;
-        $lines = explode("\n", $content);
+        $lines = explode("\n", $content ?? '');
 
-        if (count($lines) != count($regexes)) {
+        if (count($lines ?? []) != count($regexes ?? [])) {
             return false;
         }
 
         foreach ($lines as $i => $line) {
-            if (!preg_match($regexes[$i], $line)) {
+            if (!preg_match($regexes[$i] ?? '', $line ?? '')) {
                 return false;
             }
         }
--- a/src/Dev/Tasks/TagsToShortcodeHelper.php
+++ b/src/Dev/Tasks/TagsToShortcodeHelper.php
@@ -93,7 +93,7 @@ class TagsToShortcodeHelper
         $this->baseClass = $baseClass ?: DataObject::class;
         $this->includeBaseClass = $includeBaseClass;
 
-        $this->validTagsPattern = implode('|', array_keys(static::VALID_TAGS));
+        $this->validTagsPattern = implode('|', array_keys(static::VALID_TAGS ?? []));
         $this->validAttributesPattern = implode('|', static::VALID_ATTRIBUTES);
     }
 
@@ -152,7 +152,7 @@ class TagsToShortcodeHelper
     {
         $sqlSelect = SQLSelect::create(['"ID"', "\"$field\""], "\"$table\"");
         $whereAnys = [];
-        foreach (array_keys(static::VALID_TAGS) as $tag) {
+        foreach (array_keys(static::VALID_TAGS ?? []) as $tag) {
             $whereAnys[]= "\"$table\".\"$field\" LIKE '%<$tag%'";
             $whereAnys[]= "\"$table\".\"$field\" LIKE '%[$tag%'";
         }
@@ -194,7 +194,7 @@ class TagsToShortcodeHelper
         $tags = $this->getTagsInContent($content);
         foreach ($tags as $tag) {
             if ($newTag = $this->getNewTag($tag)) {
-                $content = str_replace($tag, $newTag, $content);
+                $content = str_replace($tag ?? '', $newTag ?? '', $content ?? '');
             }
         }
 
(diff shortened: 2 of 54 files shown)


src/Dev/Tasks/SecureAssetsMigrationHelper.php
```diff
@@ -138,14 +138,14 @@ public function setLogger(LoggerInterface $logger)
...
+        $lines = explode("\n", $content ?? '');
 
-        if (count($lines) != count($regexes)) {
+        if (count($lines ?? []) != count($regexes ?? [])) {
             return false;
         }
 
         foreach ($lines as $i => $line) {
-            if (!preg_match($regexes[$i], $line)) {
+            if (!preg_match($regexes[$i] ?? '', $line ?? '')) {
```
GuySartorelli, line 148:
```suggestion
            if (!preg_match($regexes[$i] ?? '//', $line ?? '')) {
```
`preg_match`, `preg_replace`, and probably other regex functions can't take an empty string for the regex argument. You end up getting
> [Warning]: preg_match(): Empty regular expression

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -211,7 +211,7 @@ private function getTagsInContent($content)
         $resultTags = [];
 
         $regex = '/<('.$this->validTagsPattern.')\s[^>]*?('.$this->validAttributesPattern.')\s*=.*?>/i';
-        preg_match_all($regex, $content, $matches, PREG_SET_ORDER);
+        preg_match_all($regex ?? '', $content ?? '', $matches, PREG_SET_ORDER);
```
GuySartorelli, line 214:
```suggestion
        preg_match_all($regex ?? '//', $content ?? '', $matches, PREG_SET_ORDER);
```

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -232,7 +232,7 @@ private function getTagTuple($tag)
             $this->validTagsPattern,
             $this->validAttributesPattern
         );
-        preg_match($pattern, $tag, $matches);
+        preg_match($pattern ?? '', $tag ?? '', $matches);
```
GuySartorelli, line 235:
```suggestion
        preg_match($pattern ?? '//', $tag ?? '', $matches);
```

src/Dev/Tasks/TagsToShortcodeHelper.php
```diff
@@ -253,7 +253,7 @@ private function getParsedFileIDFromSrc($src)
 
         // set fileID to the filepath relative to assets dir
         $pattern = sprintf('#^/?(%s/?)?#', ASSETS_DIR);
-        $fileID = preg_replace($pattern, '', $src);
+        $fileID = preg_replace($pattern ?? '', '', $src ?? '');
```
GuySartorelli, line 256:
```suggestion
        $fileID = preg_replace($pattern ?? '//', '', $src ?? '');
```

GuySartorelli: Though in this case we can just trust that it's a valid regex based on how it's assigned above.

src/Dev/Tasks/TagsToShortcodeHelper.php
GuySartorelli: Though in this case we can just trust that it's a valid regex based on how it's assigned above.

src/Dev/Tasks/TagsToShortcodeHelper.php
GuySartorelli: May be worth looking at where `$pattern` is assigned since this needs to be touched anyway and seeing if that null coalescing operator is needed

GuySartorelli (changes requested):
This is a really long PR so I stopped marking them pretty quickly but there are a lot of examples of the below regex functions getting an empty string where an empty regex is needed. Probably worth amending the automated thing and re-running it rather than trying to go through them all by hand.

src/Dev/Tasks/SecureAssetsMigrationHelper.php
emteknetnz: > [Warning]: preg_match(): Empty regular expression

IMO this is a better outcome than passing in a legit made up regex such as `//` which is almost a logic change.

We'd almost be better off not using the null collascing operator at all and which would risk null values here and potentially deprecation warnings.  I think I'd rather have a regular `Warning` over a `deprecation warning` though, as deprecated warnings in CI are very easier to ignore.

GuySartorelli: Yup that's a fair point, I can get behind that logic.

GuySartorelli (approved):
For posterity: See discussion in silverstripe/silverstripe-admin#1294 about the general approach. tl;dr:
> the purposes of this PR is to avoid passing null (specifically null) from built-in functions that as of php 8.1 will throw deprecation warnings if null is passed in. The approach is necessarily heavy-handed, and while there are many situations where it isn't needed, it would be prohibitively laborious to find all of those situations by hand.

Looks good to me, but I'll hold off merging until the outstanding questions about `file_get_contents()`, and `file_put_contents()`, and `unlink()` have been answered on the other PRs.
Edit: Question was answered. There is no problem - the approach here is correct.

PR #504, opened 2022-08-06
API File is not CMSPreviewable

GuySartorelli:
The assets CMS section no longer uses the traditional preview panel. This class being previewable via the preview panel is no longer useful.

~~Now that the `CMSPreviewable` interface has been moved to the `silverstripe/admin` module (see https://github.com/silverstripe/silverstripe-admin/pull/1340 and https://github.com/silverstripe/silverstripe-framework/pull/10440), which is not a dependency of this module, it's time to stop implementing that interface here.~~
The `CMSPreviewable` interface isn't moving anymore (see No longer needed - see https://github.com/silverstripe/silverstripe-framework/pull/10440#issuecomment-1246137962) but I still think this PR is worthwhile.

Note that I have kept the `getMimeType()` method because that's useful for `File` generally.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1339

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -20,7 +20,6 @@ use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\HTMLReadonlyField;
 use SilverStripe\Forms\TextField;
 use SilverStripe\ORM\ArrayList;
-use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\HasManyList;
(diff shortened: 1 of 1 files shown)


GuySartorelli:
CI failures are unrelated - they're happening on the `2` branch as well. See https://github.com/silverstripe/silverstripe-assets/actions/runs/3139439033

GuySartorelli:
Closing this - I forgot that assets can be used without asset-admin

PR #507, opened 2022-08-26
Rescue Master Branch PR: ENH Stop using old deprecated `allVersions` call

GuySartorelli:
Rescues https://github.com/silverstripe/silverstripe-assets/commit/f4259a148fa3f6139f85853d7af6d5f06e766e23

Targetting 1 because `allVersions()` is already deprecated and we don't want to be throwing deprecation warnings for core usage.

I only kept the actual change of method used - the other changes from the original commit caused failures in tests.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10350


emteknetnz (changes requested):
This looks completely different from commit that was being rescued?

GuySartorelli:
Please read the PR description which explains the difference between this PR and the original which is rescued.

emteknetnz (approved):

PR #514, opened 2022-09-06
FIX: Ensure attributes are stored against the Image record (fixes #513)

### LINKED ISSUE #513: Unable to access LazyLoad setting from extensions applied to Image
Example:

```php
<?php

namespace App\Extensions\Assets;

use SilverStripe\Core\Extension;

class ImageExtension extends Extension
{
    public function Test()
    {
        var_dump($this->getOwner()->IsLazyLoaded());
    }
}
```

Calling `$Image.LazyLoad(false).Test` in a template will always output `true`.

This seems to be because the `ImageManipulation` trait is applied to both `File` (`Image` extends `File`) and `DBFile`. Calling `.LazyLoad(false)` calls the following code:

https://github.com/silverstripe/silverstripe-assets/blob/ce9464b6282c34e37324987ea2dd74cae89c50ca/src/ImageManipulation.php#L1101-L1111

This then returns the a `DBFile` instance with the attribute set, but the attribute is never set on the `Image` object itself. Setting the attributes on the `Image` object via `$file->getFailover()->initAttributes()` fixes this, but goes against the documentation for `ImageManipulation::setAttribute()` which states the method is supposed to be immutable

### PRs
- https://github.com/silverstripe/silverstripe-assets/pull/514

lozcalver:
Proposed solution to #513 - pushing to see if CI builds pass.

Currently calling `Image::setAttributes()` will return a `DBFile` instance instead of a modified copy of the original image, this changes that behaviour. Both `Image` and `DBFile` implement the `AssetContainer` interface so this isn’t a BC break

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -1072,23 +1072,15 @@ trait ImageManipulation
      */
     private function copyImageBackend(): AssetContainer
     {
-        // Store result in new DBFile instance
-        /** @var DBFile $file */
-        $file = DBField::create_field(
-            'DBFile',
-            [
-                'Filename' => $this->getFilename(),
-                'Hash' => $this->getHash(),
-                'Variant' => $this->getVariant()
-            ]
-        );
+        $file = clone $this;
 
         $backend = $this->getImageBackend();
         if ($backend) {
             $file->setImageBackend($backend);
         }
 
-        return $file->setOriginal($this);
+        $file->File->setOriginal($this);
+        return $file;
     }
 
     /**
@@ -1106,6 +1098,9 @@ trait ImageManipulation
             $this->attributes,
             [$name => $value]
         ));
+
+        // If this file has already been rendered then AttributesHTML will be cached, so we have to clear the cache
+        $file->objCacheClear();
 
         return $file;
     }
--- a/tests/php/ImageManipulationTest.php
+++ b/tests/php/ImageManipulationTest.php
@@ -2,19 +2,16 @@
 
 namespace SilverStripe\Assets\Tests;
 
-use InvalidArgumentException;
 use Prophecy\Prophecy\ObjectProphecy;
 use Silverstripe\Assets\Dev\TestAssetStore;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Folder;
 use SilverStripe\Assets\Image;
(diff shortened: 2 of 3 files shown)


lozcalver:
Anyone have any feedback on this? It’s still causing me issues 😞

michalkleiner:
I'm not against it, but as Guy said on the issue, it's hard to foresee the potential implications and what this could break, mainly about the backend.

Could you try a mixed solution where we keep the method and update the implementation to e.g.
```
/**
     * Create an equivalent DBFile object
     * @return AssetContainer
     */
    private function copyImageBackend(): AssetContainer
    {
        $file = clone $this;

        $backend = $this->getImageBackend();
        if ($backend) {
            $file->setImageBackend($backend);
        }

        return $file->setOriginal($this);
    }
```

and then in the `setAttribute` keep it as it was and add the cache clear call. Can you test if that would still work for you? If so, I'd be probably ok with that.

lozcalver:
Thanks @michalkleiner, I’ve pushed up a new commit with what you’ve suggested

src/ImageManipulation.php
```diff
@@ -1072,23 +1072,15 @@ protected function castDimension($value, $dimension)
...
+        $file = clone $this;
 
         $backend = $this->getImageBackend();
         if ($backend) {
             $file->setImageBackend($backend);
         }
 
-        return $file->setOriginal($this);
+        $file->File->setOriginal($this);
+        return $file;
```
lozcalver, line 1083: This slight tweak is because `copyImageBackend()` previously returned a `DBFile` instance, it now returns a `File` containing a `DBFile`, and `setOriginal` doesn’t exist on `File`. As before, no BC break because the return type is unchanged

michalkleiner:
Thanks for the update @kinglozzer. Don't want to be too annoying, but do you think we can add a test for this that would fail with the original implementation?

lozcalver:
Have added a simple test that fails with the original implementation

michalkleiner:
Legend, will try to have a look later today and get this in.

GuySartorelli (approved):
Nice!

GuySartorelli:
Released as [1.12.1](https://github.com/silverstripe/silverstripe-assets/releases/tag/1.12.1)

Issue #501, opened 2022-07-14
Change the default on File Grant


maxime-rainville:
`File::getURL()` and `getSourceURL()` implicitly session grant you access to view the file. This has cause us no ends of grievance over the years. We should change that behaviour in CMS5 either to:
- change to default to require `false` so you don't get an implicit session grant or
- only give you a session grant if you have read access to the file.

https://github.com/silverstripe/silverstripe-assets/blob/db24a78282d4cd7e98c847da660d7a273b442b60/src/File.php#L894-L914

While the change itself is pretty simple it could have a lot of downstream affect that might not be immediately apparent.

## Acceptance criteria
- [x] Default methods doesn't auto grant you session access to view files
- [x] There's a sensible way to session grant you access to the file, but it's not used by default
- [ ] Not using session grant out of the box doesn't have a big performance impact.
- [x] Validate that no other methods on File auto session grant you access to file.

## Note
- Maybe the default URL methods make a CanView check and grant you session access for later.
- This may break a few things along the way. The obvious problems can be fixed on the spot and/or follow up cards can be created.

### PR
- https://github.com/silverstripe/silverstripe-assets/pull/517

maxime-rainville:
@silverstripe/core-team Any views on this. Auto-grant was very needed early on, but we updater the asset store to do a can view check in later minors.

Do we expect this to cause a lot of headaches for people upgrading?

Do we have a preferences between having dedicated method for sessions granting vs changing the default parameter to `false`?

sminnee:
The second option - checking permissions before providing a session grant - seems like a good default as it preserves the intent of the original implementation while clearing out bad cases.

Of course, it assumes that there *is* a clear notion of view permissions and so might be limited to use on the File object.

If you had no grant at all by default, what would it mean, specifically, for people upgrading projects? That seems to be an important question to answer when determining upgrade impact.

emteknetnz:
cms-editors would need to have a view permissions on files explicitly set so that they're able to view draft files.  There's a couple of ways I can think of doing this:
- Check the FILE_EDIT_ALL permission for a group the cms-editor belongs to, OR
- Add the cms-editor group as a group able to view an ancestor folder of the file (permissions cascade up as intermediate folder 'inherit')

This seems like a fairly minimal upgrade step that would need to be documented

sminnee:
Yeah I agree that for a major bump that seems fine.

sabina-talipova:
[...]

PR #515, opened 2022-09-16
ENH Check canViewFile permissions before automatically grant access

sabina-talipova:
### Description
Add condition to check if user has access to view the file and only then give him a session grant access to the file.

### Parent Issue
- https://github.com/silverstripe/silverstripe-assets/issues/501

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -891,8 +891,12 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      * @param bool $grant Ensures that the url for any protected assets is granted for the current user.
      * @return string
      */
-    public function getURL($grant = true)
+    public function getURL($grant = false)
     {
+        if ($this->File->canViewFile()) {
+            $grant = true;
+        }
+        
         if ($this->File->exists()) {
             return $this->File->getURL($grant);
         }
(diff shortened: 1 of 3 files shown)


src/File.php
```diff
@@ -893,6 +893,9 @@ public function getAbsoluteURL()
      */
     public function getURL($grant = true)
```
ChloeSartorelli: The acceptance criteria in the issue say this should not automatically grant by default - same as `getSourceURL()`.

GuySartorelli (changes requested):
Is this meant to still be in draft?

One of the acceptance criteria for the issue this is attached with is to check for anything else that grants permissions - have you done that?

src/File.php
sabina-talipova: There is some discussion (https://github.com/silverstripe/silverstripe-assets/issues/501#issuecomment-1207497520) in the parent issue regarding best approach . 
If I understood properly this :
"**_The second option - checking permissions before providing a session grant - seems like a good default as it preserves the intent of the original implementation while clearing out bad cases._**"
I left argument `$grant` as `true` and check does user have a permission to the file or not. I also mentioned this in my comment here https://github.com/silverstripe/silverstripe-assets/issues/501#issuecomment-1248874115

This is the reason why I left this task in Draft for now.

emteknetnz (changes requested):
This is for CMS5 so it needs to target the `2` branch.  You'll need to
a) update the target branch for this PR
b) rebase your work on top of `2`

src/File.php
emteknetnz: Must be `$grant = false` - the entire purpose of this original issue is to change the default of the optional param to `false`

sabina-talipova: DONE

sabina-talipova:
CLOSED after rebase.

Issue #497, opened 2022-06-25
CMS5 Upgrade to league/flysystem v3


maxime-rainville:
## Acceptance criteria
- [x] CMS5 is updated to use `league/flysystem v3`
- [ ] Where possible use the flysystem interface instead of actual classes

## PR
- https://github.com/silverstripe/silverstripe-assets/pull/524

## silverstripe/installer CI run
- https://github.com/emteknetnz/silverstripe-installer/actions/runs/3358146854

sabina-talipova:
@maxime-rainville , 
The module already requires `"symfony/filesystem": "^6.1"`.

emteknetnz:
Linked PR has been merged

PR #516, opened 2022-09-21
Updating league/flysystem minimum

dawb:
Updating to minimum version of ^1.1.0 to align with the minimum requirements with fullscreeninteractive/silverstripe-azure-blob-storage. 

Latest version of 1 is 1.1.9

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "silverstripe/vendor-plugin": "^1.0",
         "symfony/filesystem": "^3.4|^4.0|^5.0",
         "intervention/image": "^2.7",
-        "league/flysystem": "^1.0.70"
+        "league/flysystem": "^1.1.0"
     },
     "require-dev": {
         "silverstripe/recipe-testing": "^2",


michalkleiner:
@dawb the reason for an update shouldn't be aligning with another 3rd party module.
Also requiring `^1.1.0` might be too new for some other dependencies and cause issues elsewhere.

What if the dependency constraint got relaxed to `^1.0` which would work for both this module and the azure blob storage module?

It's also what [silverstripe/silverstripe-s3](https://github.com/silverstripe/silverstripe-s3/blob/main/composer.json#L22) uses.

lozcalver:
`^1.0.70` should already allow you to install `1.1.19`, IIRC it’s the “tilde” operator (`~1.0.70`) that would restrict you to 1.0.x releases only

dawb:
Thanks @kinglozzer , currently it doesn't allow for the install as it comes back that assets 1.11.0 only satisfies with 1.0.70 (see screenshot below).

If @michalkleiner 's solution opens up this requirement, could we use that as a solution instead? I understand that we shouldn't be updating based on other module requirements but currently it's also limiting requirements.

Screen Shot 2022-09-27 at 11 26 01 AM

lozcalver:
Are you running the latest version of composer? `^1.0.70` should definitely allow `1.1.9` to be installed: https://getcomposer.org/doc/articles/versions.md#caret-version-range-

dawb:
@kinglozzer we are using composer 2.4.2

lozcalver:
What happens if you add `"league/flysystem": "^1.1.0"` to your root composer.json?

dawb:
Thanks @kinglozzer , that resolves that issue. The league/filesystem is actually required by version 3.0 or higher. Is there a plan in the near future to upgrade this module to work with version 3 of league/flysystem?

lozcalver:
Yep, that’s being tracked here for CMS5: https://github.com/silverstripe/silverstripe-assets/issues/497

PR #517, opened 2022-09-28
ENH Check canViewFile permissions before automatically grant access

sabina-talipova:
### Description
Add condition to check if user has access to view the file and only then give him a session grant access to the file.

### Parent Issue
- https://github.com/silverstripe/silverstripe-assets/issues/501

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -891,8 +891,12 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      * @param bool $grant Ensures that the url for any protected assets is granted for the current user.
      * @return string
      */
-    public function getURL($grant = true)
+    public function getURL($grant = false)
     {
+        if (!$grant && $this->canView()) {
+            $grant = true;
+        }
+        
         if ($this->File->exists()) {
             return $this->File->getURL($grant);
         }
@@ -905,8 +909,12 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      * @param bool $grant Ensures that the url for any protected assets is granted for the current user.
      * @return string
      */
-    public function getSourceURL($grant = true)
+    public function getSourceURL($grant = false)
     {
+        if (!$grant && $this->canView()) {
+            $grant = true;
+        }
+
         if ($this->File->exists()) {
             return $this->File->getSourceURL($grant);
         }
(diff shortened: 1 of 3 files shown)


src/File.php
```diff
@@ -891,8 +891,12 @@ public function getAbsoluteURL()
      * @param bool $grant Ensures that the url for any protected assets is granted for the current user.
      * @return string
      */
-    public function getURL($grant = true)
+    public function getURL($grant = false)
     {
+        if ($this->File->canViewFile()) {
+            $grant = true;
+        }
```
emteknetnz:
```suggestion
        if (!$grant && $this->canView()) {
            $grant = true;
        }
```

src/File.php
```diff
@@ -905,8 +909,12 @@ public function getURL($grant = true)
      * @param bool $grant Ensures that the url for any protected assets is granted for the current user.
      * @return string
      */
-    public function getSourceURL($grant = true)
+    public function getSourceURL($grant = false)
     {
+        if ($this->File->canViewFile()) {
+            $grant = true;
+        }
```
emteknetnz:
```suggestion
        if (!$grant && $this->canView()) {
            $grant = true;
        }
```

src/File.php
emteknetnz: Is there any reason why we'd use `$this->File->canViewFile()` over regular `$this->canView()`?

We should also perform a `!$grant` check first to prevent unnecessary logic

emteknetnz (changes requested):
There's some broken unit tests in CI - however they relate to this https://github.com/silverstripe/silverstripe-assets/pull/518#issue-1389946419

Did you identify any content editing experiences that are likely to break because of this change?

src/File.php
sabina-talipova: DONE

src/File.php
sabina-talipova: DONE

emteknetnz:
@sabina-talipova Did you identify any content editing experiences that are likely to break because of this change?

sabina-talipova:
@emteknetnz , I did a few tests in Page section, Submission section on UserFormPage and Files section. And I also tested DRAFT content on Published and Unpublished pages. I haven't found any problem with Content or UI.
If user has "Access to Page section" he will have an access to the File.

emteknetnz (approved):

PR #523, opened 2022-10-18
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10531


GuySartorelli (changes requested):
Some version numbering is incorrect. Use `1` major version instead and apply to any other deprecations in this module as well.

GuySartorelli (approved):

PR #524, opened 2022-10-18
DEP Upgrade legue/flysystem to version 3.0

sabina-talipova:
## Issue
https://github.com/silverstripe/silverstripe-assets/issues/497

## Description

- Upgrade Flysystem to version 3.0
- Upgrade existing functionality in Asset classes by following provided documentation: 
   - [Upgrade from 1.x](https://flysystem.thephpleague.com/docs/upgrade-from-1.x/);
   - [What is new in Flysystem V2 & V3](https://flysystem.thephpleague.com/docs/what-is-new/) 
- New class `ExtendedFilesystem extends Filesystem` to provide `getAdapter()` method, that was  removed from main `legue/flysystem` `Filesystem` class ([GitHub discussion](https://github.com/thephpleague/flysystem/issues/1285))
- New class `ExtendedLocalFilesystemAdapter extends LocalFilesystemAdapter` to provide `prefixPath` method that replaces `applyPathPrefix` method from removed `abstract class AbstractAdapter` ([GitHub discussion](https://github.com/thephpleague/flysystem/issues/1285#issuecomment-808947630))
- New class `SilverStripe\Assets\Util` to replace removed `League\Flysystem\Util` and implement some of required methods.

## Notes:
- Replace invoke `Filesystem has()` method to more appropriated `directoryExists ` or `fileExists`, since `has` checks both cases `directoryExists ` or `fileExists` and returns `true`, if one of them returns `true`;
- New `listContents` method returns `DirectoryListing`, but previous version returned `array`. For this case, invoke `toArray` method on result to return `array` to support existing functionality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "silverstripe/vendor-plugin": "^2",
         "symfony/filesystem": "^6.1",
         "intervention/image": "^2.7.2",
-        "league/flysystem": "^1.1.9"
+        "league/flysystem": "^3.9.0"
     },
     "require-dev": {
         "silverstripe/recipe-testing": "^3",
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -2,14 +2,12 @@
 
 namespace SilverStripe\Assets\Flysystem;
 
+use Exception;
 use Generator;
 use InvalidArgumentException;
-use League\Flysystem\Directory;
-use League\Flysystem\Exception as FlysystemException;
-use League\Flysystem\Filesystem;
-use League\Flysystem\Util;
 use LogicException;
 use SilverStripe\Assets\File;
+use SilverStripe\Assets\Util;
 use SilverStripe\Assets\FilenameParsing\FileIDHelper;
 use SilverStripe\Assets\FilenameParsing\FileResolutionStrategy;
 use SilverStripe\Assets\FilenameParsing\HashFileIDHelper;
@@ -581,7 +579,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
                 }
             }
 
-            $result = $filesystem->putStream($fileID, $stream);
+            $result = $filesystem->writeStream($fileID, $stream);
 
             // If we have an hash for a main file, let's pre-warm our file hashing cache.
             if ($hash || !$variant) {
@@ -640,7 +638,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
                             // Invalidate hash of delete file
                             $hasher->invalidate($origin, $fs);
                         } else {
-                            $fs->rename($origin, $destination);
+                            $fs->move($origin, $destination);
                             // Move cached hash value to new location
                             $hasher->move($origin, $fs, $destination);
                         }
@@ -754,9 +752,9 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         if ($dirname
             && ltrim($dirname ?? '', '.')
             && !$this->config()->get('keep_empty_dirs')
-            && !$filesystem->listContents($dirname)
+            && !$filesystem->listContents($dirname)->toArray()
         ) {
-            $filesystem->deleteDir($dirname);
+            $filesystem->deleteDirectory($dirname);
             $this->truncateDirectory(dirname($dirname ?? ''), $filesystem);
         }
     }
@@ -772,7 +770,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
     protected function findVariants($fileID, Filesystem $filesystem)
     {
         $dirname = ltrim(dirname($fileID ?? ''), '.');
-        foreach ($filesystem->listContents($dirname) as $next) {
+        foreach ($filesystem->listContents($dirname)->toArray() as $next) {
             if ($next['type'] !== 'file') {
                 continue;
             }
@@ -840,7 +838,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
 
                 // Cache destination file into the origin store under a `.swap` directory
                 $stream = $to->readStream($toFileID);
-                $from->putStream('.swap/' . $fromFileID, $stream);
+                $from->writeStream('.swap/' . $fromFileID, $stream);
                 if (is_resource($stream)) {
                     fclose($stream);
                 }
@@ -862,7 +860,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
             $toFileID = $toStrategy->buildFileID($variantParsedFileID);
 
             $stream = $from->readStream($fromFileID);
-            $to->putStream($toFileID, $stream);
+            $to->writeStream($toFileID, $stream);
             if (is_resource($stream)) {
                 fclose($stream);
             }
@@ -875,10 +873,10 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
 
         foreach ($swapFiles as $variantParsedFileID) {
             $fileID = $variantParsedFileID->getFileID();
-            $from->rename('.swap/' . $fileID, $fileID);
+            $from->move('.swap/' . $fileID, $fileID);
             $hasher->move('.swap/' . $fileID, $from, $fileID);
         }
-        $from->deleteDir('.swap');
+        $from->deleteDirectory('.swap');
     }
 
     public function protect($filename, $hash)
@@ -916,7 +914,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         foreach ($this->findVariants($fileID, $from) as $nextID) {
             // Copy via stream
             $stream = $from->readStream($nextID);
-            $to->putStream($nextID, $stream);
+            $to->writeStream($nextID, $stream);
             if (is_resource($stream)) {
                 fclose($stream);
             }
@@ -954,7 +952,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
             $toFileID = $toStrategy->buildFileID($variantParsedFileID);
 
             $stream = $from->readStream($fromFileID);
-            $to->putStream($toFileID, $stream);
+            $to->writeStream($toFileID, $stream);
             if (is_resource($stream)) {
                 fclose($stream);
             }
@@ -1071,7 +1069,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
      *
      * @param resource $stream
      * @return string Filename of resulting stream content
-     * @throws FlysystemException
+     * @throws Exception
      */
     protected function getStreamAsFile($stream)
     {
@@ -1079,14 +1077,14 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         $file = tempnam(sys_get_temp_dir(), 'ssflysystem');
         $buffer = fopen($file ?? '', 'w');
         if (!$buffer) {
-            throw new FlysystemException("Could not create temporary file");
+            throw new Exception("Could not create temporary file");
         }
 
         // Transfer from given stream
         Util::rewindStream($stream);
         stream_copy_to_stream($stream, $buffer);
         if (!fclose($buffer)) {
-            throw new FlysystemException("Could not write stream to temporary file");
+            throw new Exception("Could not write stream to temporary file");
         }
 
         return $file;
@@ -1113,7 +1111,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
      * @param string $variant Variant to write
      * @param array $config Write options. {@see AssetStore}
      * @return array Tuple associative array (Filename, Hash, Variant)
-     * @throws FlysystemException
+     * @throws InvalidArgumentException
      */
     protected function writeWithCallback($callback, $filename, $hash, $variant = null, $config = [])
     {
@@ -1198,10 +1196,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         }
 
         // Submit and validate result
-        $result = $callback($fs, $parsedFileID->getFileID());
-        if (!$result) {
-            throw new FlysystemException("Could not save {$filename}");
-        }
+        $callback($fs, $parsedFileID->getFileID());
 
         return $parsedFileID->getTuple();
     }
@@ -1233,7 +1228,15 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         // If `applyToFileOnFilesystem` calls our closure we'll know for sure that a file exists
         return $this->applyToFileOnFilesystem(
             function (ParsedFileID $parsedFileID, Filesystem $fs) {
-                return $fs->getMetadata($parsedFileID->getFileID());
+                $path = $parsedFileID->getFileID();
+                return [
+                    'timestamp' => $fs->lastModified($path),
+                    'fileExists' => $fs->fileExists($path),
+                    'directoryExists' => $fs->directoryExists($path),
+                    'type' => $fs->mimeType($path),
+                    'size' => $fs->fileSize($path),
+                    'visibility' => $fs->visibility($path),
+                ];
             },
             new ParsedFileID($filename, $hash, $variant)
         );
@@ -1244,7 +1247,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         // If `applyToFileOnFilesystem` calls our closure we'll know for sure that a file exists
         return $this->applyToFileOnFilesystem(
             function (ParsedFileID $parsedFileID, Filesystem $fs) {
-                return $fs->getMimetype($parsedFileID->getFileID());
+                return $fs->mimetype($parsedFileID->getFileID());
             },
             new ParsedFileID($filename, $hash, $variant)
         );
@@ -1551,14 +1554,14 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
     protected function createResponseFor(Filesystem $flysystem, $fileID)
     {
         // Block directory access
-        if ($flysystem->get($fileID) instanceof Directory) {
+        if ($flysystem->directoryExists($fileID)) {
             return $this->createDeniedResponse();
         }
 
         // Create streamable response
         $stream = $flysystem->readStream($fileID);
-        $size = $flysystem->getSize($fileID);
-        $mime = $flysystem->getMimetype($fileID);
+        $size = $flysystem->fileSize($fileID);
+        $mime = $flysystem->mimetype($fileID);
         $response = HTTPStreamResponse::create($stream, $size)
             ->addHeader('Content-Type', $mime);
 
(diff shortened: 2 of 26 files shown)


src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1198,9 +1198,10 @@ function (
         }
 
         // Submit and validate result
-        $result = $callback($fs, $parsedFileID->getFileID());
-        if (!$result) {
-            throw new FlysystemException("Could not save {$filename}");
+        try {
+            $callback($fs, $parsedFileID->getFileID());
+        } catch (UnableToWriteFile $exception) {
+            throw new UnableToWriteFile("Could not save {$filename}");
```
emteknetnz:
```suggestion
        $result = $callback($fs, $parsedFileID->getFileID());
        if (!$result) {
            throw new UnableToWriteFile("Could not save {$filename}");
```

If there's a `UnableToWriteFile` exception thrown from `$callback()` there's no need to catch it only to re-throw a new `UnableToWriteFile` exception

src/Flysystem/GeneratedAssets.php
```diff
@@ -102,21 +102,20 @@ protected function checkOrCreate($filename, $callback = null)
...
-                ->put($filename, $content);
-
-        if (!$result) {
-            throw new Exception("Error regenerating file \"{$filename}\"");
+        try {
+            $this
+            ->getFilesystem()
+            ->write($filename, $content);
+        } catch (UnableToWriteFile $exception) {
+            throw new UnableToWriteFile("Error regenerating file \"{$filename}\"");
```
emteknetnz:
```suggestion
        $result = $this
                ->getFilesystem()
                ->put($filename, $content);

        if (!$result) {
            throw new UnableToWriteFile("Error regenerating file \"{$filename}\"");
```
No need to create a new `UnableToWriteFile` after catching

src/Util.php
```diff
@@ -0,0 +1,20 @@
...
+class Util
+{
+    public static function rewindStream($resource)
+    {
+        if (ftell($resource) !== 0 && static::isSeekableStream($resource)) {
+            rewind($resource);
+        }
+    }
+
+    public static function isSeekableStream($resource)
```
emteknetnz:
```suggestion
    public static function isSeekableStream(resource $resource): void
```

src/Util.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace SilverStripe\Assets;
+
+class Util
+{
+    public static function rewindStream($resource)
```
emteknetnz:
```suggestion
    public static function rewindStream(resource $resource): void
```

src/Util.php
```diff
@@ -0,0 +1,20 @@
...
+        if (ftell($resource) !== 0 && static::isSeekableStream($resource)) {
+            rewind($resource);
+        }
+    }
+
+    public static function isSeekableStream($resource)
+    {
+        $metadata = stream_get_meta_data($resource);
+
+        return $metadata['seekable'];
```
emteknetnz:
```suggestion
        return stream_get_meta_data($resource)['seekable'];
```

tests/php/ProtectedFileControllerTest.php
```diff
@@ -321,7 +321,7 @@ public function testFolders()
 
         // Flysystem reports root folder as not present
         $result = $this->get('assets');
-        $this->assertResponseEquals(404, null, $result);
+        $this->assertResponseEquals(403, null, $result);
```
emteknetnz: A 403 seems like a security downgrade from a 404, as it acknowledges to the attacker that this file exists, it's just that they cannot access it.  A 404 response does not lot the attacker know the file even exists.

Can we change this behavior so that a 404 is returned instead of a 403?

composer.json
```diff
@@ -24,7 +24,7 @@
         "silverstripe/vendor-plugin": "^2",
         "symfony/filesystem": "^6.1",
         "intervention/image": "^2.7.2",
-        "league/flysystem": "^1.1.9"
+        "league/flysystem": "^3.0"
```
emteknetnz:
```suggestion
        "league/flysystem": "^3.9.0"
```
Latest release is 3.9.0 https://github.com/thephpleague/flysystem/releases/tag/3.9.0

src/Flysystem/ExtendedFilesystem.php
```diff
@@ -0,0 +1,29 @@
...
+        array $config = [],
+        PathNormalizer $pathNormalizer = null
+    ) {
+        $this->adapter = $adapter;
+        parent::__construct($adapter, $config, $pathNormalizer);
+    }
+
+    /**
+     * @return FilesystemAdapter
+     */
```
emteknetnz:
```suggestion
```

src/Flysystem/ExtendedLocalFilesystemAdapter.php
```diff
@@ -0,0 +1,32 @@
...
+use League\Flysystem\PathPrefixer;
+use League\Flysystem\UnixVisibility\VisibilityConverter;
+use League\MimeTypeDetection\MimeTypeDetector;
+
+class ExtendedLocalFilesystemAdapter extends LocalFilesystemAdapter
+{
+    /**
+     * @var PathPrefixer
+     */
+    private $pathPrefixer;
```
emteknetnz:
```suggestion
    private PathPrefixer $pathPrefixer;
```

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1233,7 +1234,15 @@ public function getMetadata($filename, $hash, $variant = null)
         // If `applyToFileOnFilesystem` calls our closure we'll know for sure that a file exists
         return $this->applyToFileOnFilesystem(
             function (ParsedFileID $parsedFileID, Filesystem $fs) {
-                return $fs->getMetadata($parsedFileID->getFileID());
+                $path = $parsedFileID->getFileID();
+                return [
+                    'timestamp' => $fs->lastModified($path),
```
emteknetnz, line 1233: We should be using `$fs->getAdapter()->getMetaData()`? similar to 1.x? https://github.com/thephpleague/flysystem/blob/1.x/src/Filesystem.php#L357

src/Util.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace SilverStripe\Assets;
+
+class Util
```
emteknetnz, line 5: The methods on this class are only used by `FlysystemAssetStore`, so you may as well just move this methods to that class. And then delete this class

You can update the existing `FlystemAssetStore::isSeekableStream()` to the logic currently in this class

tests/php/Dev/Tasks/SecureAssetsMigrationHelperTest.php
```diff
@@ -112,7 +112,7 @@ public function dataMigrate()
             ],
             'Unprotected' => [
                 'unprotected',
-                null,
+                '',
```
emteknetnz, line 115: Why has this changed?

emteknetnz (changes requested):
Could you remove the `Extended` prefix on the new classes `ExtendedFilesystem` and `ExtendedLocalFilesystemAdapter` and instead prefix `League` to the import statement on the league class e.g.

`use League\Flysystem\Filesystem as LeagueFilesystem;`

We do a similar thing with `Email` - https://github.com/silverstripe/silverstripe-framework/blob/5/src/Control/Email/Email.php#L23

src/Flysystem/FlysystemAssetStore.php
sabina-talipova: Since new filesystem operations no longer return boolean value, if filesystem operation was a success or not. Now they throw `UnableTo*File` exceptions and we should process these exceptions. 
I implemented this changes following this Docs ["No more success result booleans" paragraph](https://flysystem.thephpleague.com/docs/what-is-new/)  and 
[Filesystem API](https://flysystem.thephpleague.com/docs/usage/filesystem-api/)

sabina-talipova: DONE

src/Flysystem/GeneratedAssets.php
sabina-talipova: DONE

src/Flysystem/ExtendedFilesystem.php
sabina-talipova: DONE

src/Flysystem/ExtendedLocalFilesystemAdapter.php
sabina-talipova: DONE

src/Flysystem/FlysystemAssetStore.php
sabina-talipova: `getMetadata` method doesn't exist anymore.
Currently, it was replaced with few new methods for each metadata attribute.
[Filesystem::getMetadata is removed](https://flysystem.thephpleague.com/docs/upgrade-from-1.x/)
[Metadata getters are renamed](https://flysystem.thephpleague.com/docs/upgrade-from-1.x/)

tests/php/Dev/Tasks/SecureAssetsMigrationHelperTest.php
sabina-talipova: `TypeError: League\Flysystem\Filesystem::write(): Argument #2 ($contents) must be of type string, null given`

Method `write` in `LocalFilesystemAdapter class` expects to get string as 2nd param.

src/Util.php
sabina-talipova: We use static method `rewindStream` in `Sha1FileHashingService` and `FlysystemAssetStore`.
And `isSeekableStream` in `FlysystemAssetStore` and `Util` uses this method itself.

src/Util.php
sabina-talipova: DONE

src/Util.php
sabina-talipova: DONE

sabina-talipova: I cannot add `resource` as a type hint.
`Warning: "resource" is not a supported builtin type and will be interpreted as a class name. `
See https://wiki.php.net/rfc/scalar_type_hints :
`No type hint for resources is added, as this would prevent moving from resources to objects for existing extensions, which some have already done (e.g. GMP).`

src/Util.php
sabina-talipova: DONE 
I cannot add `resource` as a type hint.
`Warning: "resource" is not a supported builtin type and will be interpreted as a class name. `
See https://wiki.php.net/rfc/scalar_type_hints :
`No type hint for resources is added, as this would prevent moving from resources to objects for existing extensions, which some have already done (e.g. GMP).`

composer.json
sabina-talipova: If we use `^` before version number, composer should install a latest release `>=3.0 < 4.0`.
I checked the version after install and it's 3.9.0.
Or should we use `3.9.0` version only?

tests/php/ProtectedFileControllerTest.php
sabina-talipova: Fix it by implementing old version of `Filesystem` `has()` method in `SilverStripe\Assets\Flysystem\Filesystem` class.

sabina-talipova: DONE

composer.json
emteknetnz: Use `^3.9.0`

The reason is that we run a `composer update --prefer-lowest` job in our CI - [example](https://github.com/silverstripe/silverstripe-framework/actions/runs/3321789438/jobs/5489921340).  Under this job, `3.0.0` would be installed.  I see no reason to support older versions so we may as well set the minimum to `3.9.0`

src/Util.php
emteknetnz: Since this is in the Util class, may as well just add the following (adapted from [this](https://stackoverflow.com/a/38430606)):

```
private static function checkIsResource($resource): void
{
    if (!is_resource($resource)) {
        throw new \InvalidArgumentException('$resource argument is not a valid resource');
    }
}
```

And then type check within `rewindStream()` and `isSeekableStream()`

I don't this we need to worry about the `object` scenario since [stream_get_meta_data()](https://www.php.net/manual/en/function.stream-get-meta-data.php) is type hinted for `resource`

src/Util.php
emteknetnz: As above, check use a new `checkIsResource()` method to type check

emteknetnz (changes requested):
### A few more code changes requested

- See comments in the Files changed tab

### Some possible references to removed methods may still exist

This list is largely for my own testing.

Looking at the list of methods removed in https://flysystem.thephpleague.com/docs/upgrade-from-1.x/

I tried searching for the following strings in `./vendor/silverstripe/**/*.php` - the idea being that there may be some classes that haven't yet been updated and either weren't covered by unit tests in the assets module, or, they are called from different modules and possibly would get caught if we ran that PR against `silverstripe/installer` (more about that in the section below).

Looks like most things have been correctly updated. 

**->getMetaData(**
- (none I think, though there's also multiple `getMetaData()` methods on other silverstripe classes)

**['basename']**
- HashFileIDHelper::parseFileID()
- LegacyFileIDHelper::parseFileID()
- LegacyFileIDHelper::parseSilverStripe30VariantFileID()
- NaturalFileIDHelper::parseFileID()

**->rename(**
- (none I think, though there's FlysystemAssetStore::rename())

**->createDir(**
- (none)

**->update(**
- (none I think, though there's also DataObject::update(), MemoryConfigCollection::update() and Config_ForClass::update())

**->updateStream(**
- (none)

**->put(**
- (none)

**->putStream(**
- (none)

**->getTimestamp(**
- (none I think, though there DBDate::getTimestamp())

**->has(**
- (large number of uses, though it doesn't look like it's been depracated or removed in 3.x)

**->getMimetype(**
- (none I think, though there is DBFile::getMimeType())

**->getSize(**
- (none I think, though there is File::getSize())

**->getVisibility(**
- (none I think, though there is File::getVisibility() and FlystemAssetStore::getVisibility())


### May be worth testing with silverstripe/installer before merge

It may be worth running this through the `silverstripe/installer` CI using [vendor-code-patcher](https://github.com/emteknetnz/vendor-code-patcher)

Unlike the last time we tried this, adding `"league/flysystem": "3.9.0 as 1.99.99"` to composer.json on the new branch of `silverstripe/installer` should work since it doesn't have any conflicting subdependencies

I'm not sure how much other modules such as asset-admin actually call league/flysystem methods directly, so there is some risk with merging this without it going via installer.  However If you don't think it's worth doing this though, I'm OK with skipping this step

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1198,9 +1197,10 @@ function (
...
 
         // Submit and validate result
-        $result = $callback($fs, $parsedFileID->getFileID());
-        if (!$result) {
-            throw new FlysystemException("Could not save {$filename}");
+        try {
+            $callback($fs, $parsedFileID->getFileID());
+        } catch (UnableToWriteFile $exception) {
+            throw $exception;
         }
```
emteknetnz:
```suggestion
        $callback($fs, $parsedFileID->getFileID());
```
If all we're doing is re-throwing the exception after catching it, we may as well not bother catching it

src/Util.php
sabina-talipova: DONE

src/Util.php
sabina-talipova: DONE

composer.json
sabina-talipova: DONE

src/Flysystem/FlysystemAssetStore.php
sabina-talipova: DONE

sabina-talipova:
@emteknetnz, I went through your list and this is a small report about my investigation:

Before I've started to work on upgrading new deps, I've checked do we use `League\Flysystem` classes anywhere else. Only `silverstripe\assets` uses this module.

- `['basename']` in the following methods `HashFileIDHelper::parseFileID()`, `LegacyFileIDHelper::parseFileID()`, `LegacyFileIDHelper::parseSilverStripe30VariantFileID()`, `NaturalFileIDHelper::parseFileID()` is a part of result of `preg_match()` php function. `basename` is a named subpattern. 
These parts are unrelated of changes that were made. 

- `getMetaData()` methods in another classes related to `SilverStripe\Assets\Storage\AssetStore` interface and  
`SilverStripe\Assets\Storage\AssetContainer` interface.
   - `class FlysystemAssetStore implements AssetStore` implements `getMetaData` method and uses new methods in `League\Flysystem\Filesystem` class that replaced old `League\Flysystem\Filesystem` `getMetaData()` method.
   - `class File extends DataObject implements AssetContainer` in `getMetaData()` method uses `getMetaData()` method in `DBFile` class, that in its order uses `getMetaData()` existing in `FlysystemAssetStore`
   
- `rename()` is a method of `FlysystemAssetStore` class and it used `League\Flysystem` method `rename()` before. I've replaced it with new method `move()`

- `update()` methods are unrelated to new changes.

- `getTimestamp()` method is unrelated to new changes. 

- `has()`. Most of the places where we invoke `has` method, we invoke them either in implementation of `CacheInterface` or in `Injector` class. 

- Most of the invoke of `has()` method in `silverstripe-assets`.  I implemented part of old `has()` method logic in new `SilverStripe\Assets\Flysystem\Filesystem` class.
 
- `getMimetype()` the similar situation as with `getMetaData()`. 

- `getSize()` in `File` class has its own logic, unrelated to new changes.
 
- `getVisibility()` the similar situation as with `getMetaData()`

emteknetnz (changes requested):
I tried running this PR using installer, there's single failing unit test on framework - https://github.com/emteknetnz/silverstripe-installer/actions/runs/3343823971/jobs/5537884602

I've fixed this on my local - could you please update `GeneratedAssets::removeContent()`:

```php
    public function removeContent($filename)
    {
        $filesystem = $this->getFilesystem();
        if ($filesystem->directoryExists($filename)) {
            $filesystem->deleteDirectory($filename);
        } elseif ($filesystem->fileExists($filename)) {
            $filesystem->delete($filename);
        }
    }
```

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1071,22 +1069,22 @@ protected function getStreamSHA1($stream)
...
+     * @throws Exception
      */
     protected function getStreamAsFile($stream)
     {
         // Get temporary file and name
         $file = tempnam(sys_get_temp_dir(), 'ssflysystem');
         $buffer = fopen($file ?? '', 'w');
         if (!$buffer) {
-            throw new FlysystemException("Could not create temporary file");
+            throw new Exception("Could not create temporary file");
```
emteknetnz, line 1080: Why was this changed? ([diff](https://github.com/silverstripe/silverstripe-assets/compare/11d4da5..88b62be)) It wasn't causing any errors in CI? If there isn't a compelling reason for changing this can we please revert this change?

emteknetnz (changes requested):

src/Flysystem/FlysystemAssetStore.php
sabina-talipova: `League\Flysystem\Exception` class doesn't exist anymore in new version of Flysystem. I removed this `use League\Flysystem\Exception as FlysystemException`

michalkleiner: Perhaps this should throw `League\Flysystem\UnableToWriteFile`?

emteknetnz: I'm kind on inclined to not use `UnableToWriteFile` here, mainly because that's thrown exclusively by league code, makes sense to me to no use the league exception for exception in our own logic. I don't think that errors with writing to temporary files really warrants a new custom Exception, so plain old Exception is fine IMO

michalkleiner: We're catching and rethrowing it from our adapter, but fair enough as it's not really related to the actual file manipulations. Actually made me look up flysystem's exceptions so some good still came out of the discussion.

emteknetnz (approved):

PR #530, opened 2022-11-28
API Remove deprecated code

sabina-talipova:
### Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10501


src/Dev/TestAssetStore.php
```diff
@@ -160,36 +160,11 @@ public static function getLocalPath(AssetContainer $asset, $forceProtected = fal
...
-    {
-        return parent::parseFileID($fileID);
-    }
-
-    public function getOriginalFilename($fileID)
-    {
-        return parent::getOriginalFilename($fileID);
-    }
-
-    public function getFilesystemFor($fileID)
```
emteknetnz, line 183: These methods shouldn't be removed - FlysystemAssetStore::getFilesystemFor() is a protected method, while this is a public method

emteknetnz (changes requested):

src/Dev/TestAssetStore.php
sabina-talipova: Done

emteknetnz (approved):

PR #531, opened 2022-11-28
API Stop using deprecated API

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10545


sabina-talipova (commented):
Should we mark as deprecated all other `*MigrationHelper `classes in `test/php/Dev/Tasks` folder as well?
Because they don't have any test cases, that can be cause of additional warnings.

emteknetnz:
We don't need to mark the unit test files in `test/php/Dev/Tasks` as deprecated as there's no need to deprecate unit tests

Issue #534, opened 2022-12-15
FileIDHelperResolutionStrategy::findVariants() ordering issue


emteknetnz:
findVariants() seems to be having some issue with the order of results returned

This was first noticed in https://github.com/silverstripe/silverstripe-assets/commit/349508447b66591671b8930a25cd3fc1f78fb97e#diff-0668b7adbf9e875b66c47d35627b4ed054921175c1c8583e407607e4c4629353R526 where the order was changed

I experienced it again locally, and fixed this unit test to be order agnostic https://github.com/silverstripe/silverstripe-assets/pull/535

Investigate to see if this is something that we should even be worried about, and if there's a fix

PR #535, opened 2022-12-15
MNT Make asset variant unit tests order agnostic

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10403

Fixes `1) SilverStripe\Assets\Tests\FilenameParsing\FileIDHelperResolutionStrategyTest::testFindVariant with data set #0 (SilverStripe\Assets\FilenameParsing\FileIDHelperResolutionStrategy Object (...), SilverStripe\Assets\FilenameParsing\ParsedFileID Object (...))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Folder/FolderFile.pdf'
+'Folder/SubFolder/SubFolderFile.pdf'`
https://github.com/emteknetnz/silverstripe-installer/actions/runs/3700324160/jobs/6268641143

Possible due to the upgrade of the Flysystem dependency, there is an ordering issue that was attempted to be fixed in https://github.com/silverstripe/silverstripe-assets/commit/349508447b66591671b8930a25cd3fc1f78fb97e#diff-0668b7adbf9e875b66c47d35627b4ed054921175c1c8583e407607e4c4629353R526

From memory this unit test was showing different results on local and CI.  This unit test was failing for me locally as well, so I've made it order agnostic.

I've raised a new issue to further investigate what's going on [here](https://github.com/silverstripe/silverstripe-assets/issues/534)

### DIFF
--- a/tests/php/FilenameParsing/FileIDHelperResolutionStrategyTest.php
+++ b/tests/php/FilenameParsing/FileIDHelperResolutionStrategyTest.php
@@ -530,14 +530,13 @@ class FileIDHelperResolutionStrategyTest extends SapphireTest
         $variantGenerator = $strategy->findVariants($tuple, $this->fs);
 
         /** @var ParsedFileID $parsedFileID */
+        $c = 0;
         foreach ($variantGenerator as $parsedFileID) {
-            $this->assertNotEmpty($expectedPaths);
-            $expectedPath = array_shift($expectedPaths);
-            $this->assertEquals($expectedPath, $parsedFileID->getFileID());
+            $this->assertTrue(in_array($parsedFileID->getFileID(), $expectedPaths));
             $this->assertEquals('mockedvariant', $parsedFileID->getVariant());
+            $c++;
         }
-
-        $this->assertEmpty($expectedPaths);
+        $this->assertSame(count($expectedPaths), $c);
     }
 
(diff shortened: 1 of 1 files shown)


GuySartorelli:
You've linked to a diff for a PR where you say "there is an ordering issue that was attempted to be fixed" - but there's no clear indication in that link, nor comments in the original PR, of what the "ordering issue" is, or what was attempted to fix it.... it seems like there's some context here that I'm missing.

It looks to me like this PR just alters the test to say that the bad behaviour is actually good (in which case IMO it's not a fix at all and should not be merged) - but I might be missing something.

GuySartorelli:
@emteknetnz Can you please respond about my above comment?

emteknetnz:
In the diff I linked to, the ordering changed because for reasons unknown, so the unit test was updated to account for that.  From memory Sabina was experiencing a different ordering on her local vs CI, so the order was thus unpredictable.

I experienced the opposite ordering again in this PR, so I made the unit test order agnostic

GuySartorelli:
Ahhhhhh okay. So the test was relying on a specific order, but the _underlying functionality_ is still correct. That's okay then.

GuySartorelli (approved):

PR #536, opened 2023-01-10
API Removed ReturnTypeWillChange annotation

maxime-rainville:
## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/645

### DIFF
--- a/src/Storage/DefaultAssetNameGenerator.php
+++ b/src/Storage/DefaultAssetNameGenerator.php
@@ -120,7 +120,7 @@ class DefaultAssetNameGenerator implements AssetNameGenerator
         return Config::inst()->get(__CLASS__, 'version_prefix');
     }
 
-    public function current(): mixed
+    public function current(): string
     {
         $version = $this->version;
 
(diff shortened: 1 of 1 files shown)


src/Storage/DefaultAssetNameGenerator.php
```diff
@@ -120,8 +120,7 @@ protected function getPrefix()
         return Config::inst()->get(__CLASS__, 'version_prefix');
     }
 
-    #[\ReturnTypeWillChange]
-    public function current()
+    public function current(): mixed
```
ChloeSartorelli:
```suggestion
    public function current(): string
```

src/Storage/DefaultAssetNameGenerator.php
```diff
@@ -146,26 +145,22 @@ public function current()
         return $filename;
     }
 
-    #[\ReturnTypeWillChange]
-    public function key()
+    public function key(): mixed
```
ChloeSartorelli:
```suggestion
    public function key(): int
```

GuySartorelli (changes requested):

GuySartorelli (approved):

Issue #541, opened 2023-02-23
Still using a slightly legacy asset resolution strategy by default


GuySartorelli:
Installer overrides the default assets file resolution strategy - this is a holdover from when this stuff was introduced in CMS 4.4

The assets module should use the optimal strategy by default, and installer should stop overriding it at all.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/542
- https://github.com/silverstripe/silverstripe-installer/pull/350

maxime-rainville:
All merged.

PR #542, opened 2023-02-23
ENH Use the optimal file resolution strat by default

GuySartorelli:
This change reflects the config that installer was already overriding it with, so new projects were using this strategy anyway. It should just be the default outright.

## Parent issue
- https://github.com/silverstripe/silverstripe-assets/issues/541

### DIFF
--- a/_config/asset.yml
+++ b/_config/asset.yml
@@ -38,7 +38,6 @@ SilverStripe\Core\Injector\Injector:
       DefaultFileIDHelper: '%$SilverStripe\Assets\FilenameParsing\HashFileIDHelper'
       ResolutionFileIDHelpers:
         - '%$SilverStripe\Assets\FilenameParsing\HashFileIDHelper'
-        - '%$SilverStripe\Assets\FilenameParsing\NaturalFileIDHelper'
       VersionedStage: Stage
 ---
 Name: assetscore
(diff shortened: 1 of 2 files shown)


maxime-rainville (changes requested):
Good job picking this up. Your build failure will be legit because the unit test assumes that NaturalFileIDHelper will be there.

I would suggest either:
- removing the test or 
- updating the test to add the `NaturalFileIDHelper` back in. 

We still need to test that can validate that fallback resolution strategy still work. So option two might be preferable if there's no other test worrying about that possibility.

GuySartorelli:
> We still need to test that can validate that fallback resolution strategy still work

Do we though? This is a fallback on a legacy strategy for a legacy setup. No new projects will be using this configuration anyway. Any projects which have been around since this was needed (and therefore use this strategy for protected files) will throw deprecation warnings when they upgrade to 4.13 (see https://github.com/silverstripe/silverstripe-framework/issues/10692).

I would argue it's more appropriate to remove that from the test - if it _does_ work and people use it as a workaround then fine, we're not going out of our way to break it per se, but we shouldn't _actively_ support legacy fallbacks IMO.

That said, after reading this if you still think option 2 is preferable I'll do that.

GuySartorelli:
Nevermind, it turns out these tests are kinda complex and it's easiest to just add that config back in for them

maxime-rainville (approved):
LGTM

PR #543, opened 2023-02-24
API Explicitly deprecate legacy file resolution

GuySartorelli:
Can't be done the same way as other deprecation behaviours because of the problem mentioned in https://github.com/silverstripe/silverstripe-framework/pull/10702#discussion_r1116403808 
So instead of interrogating the configuration, we're only emitting this deprecation warning if we _actually_ find files which are in a legacy file structure.

This does mean the deprecation warning will be somewhat intermittent (only the oldest files in a project are likely to trip this, so it relies on developers happening to hit endpoints that call for those files) I think this is the best we can do.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10692

### DIFF
--- a/src/FilenameParsing/LegacyFileIDHelper.php
+++ b/src/FilenameParsing/LegacyFileIDHelper.php
@@ -132,18 +132,34 @@ class LegacyFileIDHelper implements FileIDHelper
             return null;
         }
 
+        // This helper is only used if the natural and hash helpers have already failed to parse the ID
+        // So if we can parse the ID here it means the project has a deprecated legacy file structure.
+        $deprecationMessage = 'Use of legacy file resolution strategies is deprecated.'
+        . ' See https://docs.silverstripe.org/en/4/developer_guides/files/file_migration/.'
+        . ' After migrating your files, change your file resolution configuration to match the defaults:'
+        . ' https://github.com/silverstripe/silverstripe-installer/blob/4/app/_config/assets.yml';
+
+
         // Can't have a resampled folder without a variant
         if (empty($matches['variant']) && strpos($fileID ?? '', '_resampled') !== false) {
-            return $this->parseSilverStripe30VariantFileID($fileID);
+            $parsed = $this->parseSilverStripe30VariantFileID($fileID);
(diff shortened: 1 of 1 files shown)


src/FilenameParsing/LegacyFileIDHelper.php
```diff
@@ -132,18 +132,34 @@ public function parseFileID($fileID)
...
+        . ' After migrating your files, change your file resolution configuration to match the defaults:'
+        . ' https://github.com/silverstripe/silverstripe-installer/blob/4/app/_config/assets.yml';
+
+
         // Can't have a resampled folder without a variant
         if (empty($matches['variant']) && strpos($fileID ?? '', '_resampled') !== false) {
-            return $this->parseSilverStripe30VariantFileID($fileID);
+            $parsed = $this->parseSilverStripe30VariantFileID($fileID);
+            if ($parsed) {
+                Deprecation::notice('4.13.0', $deprecationMessage, Deprecation::SCOPE_GLOBAL);
```
emteknetnz, line 147: Should we be adding what the fileID is if it `$parsed` to the Deprecation message?  $fileID it will be a part of URL right?  If these messages show up in their logs it would be helpful to try and track down what's actually linking to them?

emteknetnz (changes requested):

src/FilenameParsing/LegacyFileIDHelper.php
ChloeSartorelli: I'm not sure it would be helpful... the migration docs should take care of any files that are in a legacy setup without having to deal with each one individually, and we de-dup deprecation warnings based on the message so if a project has a fully legacy assets system with thousands of assets they'll end up with a FLOOD in their logs with one message for each file.

I'll add it in if you still think it'd be useful with that context in mind.

emteknetnz: Yeah probably leave ... a single message should be enough.  Likely they'd get a bunch when they're navigating around asset-admin

emteknetnz (approved):

Issue #550, opened 2023-05-17
Wrong permissions set on directories created through SS v5


lars-lemon8:
It seems that new 'asset' directories created by SS getting a 'private' visibility. (0700 permissions instead of 0775). 
This makes the directory unreachable from the web if apache runs as a different user. The exact same server does not have this problem with SS4. 

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/551

GuySartorelli:
Can you please provide more details? Especially specific directories involved (I ask because in discussions on slack you explicitly mentioned folders related to tinymce) and steps to reproduce this from a fresh installation.

derkgort:
First you run dev/build, this will remove the _tinymce dir in the assets folder.
Then edit a page, this will create the _tinymce folder.
Now the _tinymce has the permissions: 0700
`drwx------ 2 test TB23-001     38 May 17 11:33 _tinymce
`
The same for when you create a new folder in the assets and publish an asset in there:

`drwx------ 2 test TB23-001    140 May 12 12:38 test`

derkgort:
Also note, in vagrant this cannot be reproduced. That's because there is some magic happening in the filesystem layer.

lozcalver:
I can reproduce this, just published an image in a directory named `test`:

```
-rw-rw-r--+ 1 www-data     www-data  822 Mar 14 16:07 error-404.html
-rw-rw-r--+ 1 www-data     www-data  717 Mar 14 16:07 error-500.html
-rw-rw-r--+ 1 www-data     www-data 1226 May 11 11:01 .htaccess
drwx--S---+ 2 www-data     www-data 4096 May 17 10:40 test
```

Permissions should be inherited from this config, but I guess that’s not happening for some reason:

https://github.com/silverstripe/silverstripe-assets/blob/de229ea2a80e1489029c309a97a2d5d27e0228fd/src/Flysystem/AssetAdapter.php#L39-L54

lozcalver:
Okay it looks like those config settings are being picked up when the adapter is constructed, I wonder if this could actually be related to this issue: https://github.com/thephpleague/flysystem/issues/1584

lars-lemon8:
Yeah, this sounds fishy indeed: 

"The problem is that setVisibility() will only be called if the Filesystem instance Config key visibility is set"

emteknetnz:
I debugged this to work out what's going on in both CMS 4.13 and CMS 5.0

Both versions use the same config

```php
[...]

PR #551, opened 2023-05-17
FIX Use 0775 permissions for directories

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-assets/issues/550

### DIFF
--- a/src/Flysystem/AssetAdapter.php
+++ b/src/Flysystem/AssetAdapter.php
@@ -39,6 +39,10 @@ class AssetAdapter extends LocalFilesystemAdapter
     /**
      * Config compatible permissions configuration
      *
+     * dir.private is defaulted to 0775 rather than 0700 to mimic the behaviour of CMS 4 where
+     * league/flysystem v1 would use the 'public' 0775 visibility for all directories, while still using
+     * the 'private' 0600 visibility for draft files
+     *
      * @config
      * @var array
      */
@@ -49,7 +53,7 @@ class AssetAdapter extends LocalFilesystemAdapter
         ],
         'dir' => [
             'public' => 0775,
-            'private' => 0700,
+            'private' => 0775,
         ]
     ];
 
--- a/tests/php/Flysystem/AssetAdapterTest.php
+++ b/tests/php/Flysystem/AssetAdapterTest.php
@@ -20,18 +20,6 @@ class AssetAdapterTest extends SapphireTest
     protected function setUp(): void
     {
         parent::setUp();
-
-        AssetAdapter::config()->set('file_permissions', [
-            'file' => [
-                'public' => 0644,
-                'private' => 0600,
-            ],
-            'dir' => [
-                'public' => 0755,
-                'private' => 0700,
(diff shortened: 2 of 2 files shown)


michalkleiner:
My gut feel is that this isn't the best solution, but at the same time I don't have that much time to dig deeper into the original issue and see for myself what could be done differently (if anything) in our code around the case when a file is published and the folder is not changed to be using the `dir.public` permissions at that point. Also if the permissions are the same as with CMS 4 then it's not that big of a deal.

GuySartorelli:
I'm on the "if the permissions are the same as with CMS 4 then it's not that big of a deal" side of things - even though it was technically a bug causing those permissions, they're what people have been using for many years now to the extent that those permissions are effectively a feature and now that it's using the "correct" permissions that's causing problems for people.

I think the best solution is, as this PR does, making it so that the CMS 4 behaviour is intentionally kept. Especially since files _do_ follow the more secure behaviour, and ultimately files are what matters in that regard.

GuySartorelli (commented):
Works well locally. It seems concerning this wasn't picked up in a test though. I think it'd be appropriate to add a unit test to avoid a regression in the future.

emteknetnz:
Added a unit test to validate the default permissions. Beyond duplicating the config in a unit test I really don't know what else to do.

It's up to league/filesystem to actually apply the permissions to newly created files/folders, so not unit testing that.

GuySartorelli:
> It's up to league/filesystem to actually apply the permissions to newly created files/folders, so not unit testing that.

I'd argue that It's up to us to ensure that our configuration is used to apply those permissions.

It's one thing to validate that the configuration is set which is what your new test does.
It's another thing to validate that the configuration is _actually used_ to set the _correct_ permissions based on that configuration. That's what I think we should have here. A test that validates that if the configuration changes, the permissions change to match (or, at the _very_ least, validate that the configuration is correctly passed through to the upstream library, but I think that'll be harder to test and result in a lest robust regression test in any case)

That said, I don't feel super strongly about it - if you give a good reason not to add a test like that I'll be happy to merge.

emteknetnz:
> It's another thing to validate that the configuration is actually used to set the correct permissions based on that configuration.

To me that's very much up to league/flysystem and we shouldn't test that. We should trust that dependencies actually do what they're supposed to do

> A test that validates that if the configuration changes, the permissions change to match (or, at the very least, validate that the configuration is correctly passed through to the upstream library

AssetAdapter is a subclass on league/flysystem  LocalFilesystemAdapter.  We'd be unit testing that we're passing the correct 2nd param to __parent::construct() - https://github.com/silverstripe/silverstripe-assets/blob/2/src/Flysystem/AssetAdapter.php#L67

This seems like it would be a pretty unusual unit test and I don't think we do this elsewhere. I don't think we should do it.

GuySartorelli:
I still think we should have a test for this - but I don't feel strongly enough about it to stop this PR from being merged, so I'll merge as is.

GuySartorelli (approved):

PR #556, opened 2023-06-15
add check for specific user inherited permission

andrewandante:
Connected to https://github.com/silverstripe/silverstripe-framework/pull/10819

Allows file/folder access to be restricted on a user-by-user basis, rather than having to use groups

## Issue
- silverstripe/silverstripe-framework#10817

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -397,6 +397,14 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
             return $member->inGroups($this->ViewerGroups());
         }
 
+        // Specific users can view this file
+        if ($this->CanViewType === InheritedPermissions::ONLY_THESE_MEMBERS) {
+            if (!$member) {
+                return false;
+            }
+            return $this->ViewerMembers()->filter('ID', $member->ID)->count() > 0;
+        }
+
         // Check default root level permissions
         return $this->getPermissionChecker()->canView($this->ID, $member);
     }
@@ -423,7 +431,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         }
 
         // Delegate to parent if inheriting permissions
-        if ($this->CanEditType === 'Inherit' && $this->ParentID) {
+        if ($this->CanEditType === InheritedPermissions::INHERIT && $this->ParentID) {
             return $this->getPermissionChecker()->canEdit($this->ParentID, $member);
         }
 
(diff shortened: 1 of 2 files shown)


src/File.php
```diff
@@ -397,6 +397,14 @@ public function canView($member = null)
             return $member->inGroups($this->ViewerGroups());
         }
 
+        // Specific users can view this file
+        if ($this->CanViewType === InheritedPermissions::ONLY_THESE_MEMBERS) {
+            if (!$member) {
+                return false;
+            }
+            return \in_array($member->ID, $this->ViewerMembers()->column('ID'), true);
```
ChloeSartorelli: as per https://github.com/silverstripe/silverstripe-cms/pull/2862#discussion_r1253699612
```suggestion
            return $this->ViewerMembers()->filter('ID', $member->ID)->count() > 0;
```

src/File.php
```diff
@@ -423,10 +431,26 @@ public function canEdit($member = null)
...
+        }
+
+        // Specific users can edit this file
+        if ($this->CanEditType === InheritedPermissions::ONLY_THESE_MEMBERS) {
+            if (!$member) {
+                return false;
+            }
+            return \in_array($member->ID, $this->EditorMembers()->column('ID'), true);
+        }
+
```
ChloeSartorelli: These shouldn't be necessary - the permission checker canedit call below should cover both of these scenarios.

GuySartorelli (changes requested):

src/File.php
ChloeSartorelli: Just tested locally - can confirm that this chunk of code isn't necessary.
Note that because inherited permissions is cached, you may need to either flush or log out and back in again to see the effect of permissions changes.

GuySartorelli (commented):
One thing that's a little confusing is if you have previously set a group, and then you decide to swap to setting explicit users, the selected group is still present.

This is solved for pages by hiding the field, but as I mentioned in one of these PRs that may be out of scope for this PR... the other way to solve it would be to _clear_ the groups if the permission type is set to something else (and same for users).... can you foresee any problems if we implemented that?

andrewandante:
>can you foresee any problems if we implemented that?

Hmm. It makes "changing back" pretty tricky, which is a bit annoying. I'd prefer to go with the "hide what is irrelevant" approach - though I could also see a "Reset to defaults"-style action which pushes everything back to Inherit and clears all the relationships. I agree it looks a bit.. mediocre 😅

andrewandante:
@GuySartorelli presumably this needs an updated `silverstripe/framework: ^5.1` constraint too?

GuySartorelli:
Ahh, yes please! I thought I checked it for that but apparently I didn't 😅

GuySartorelli (approved):
LGTM

Issue #557, opened 2023-06-22
FileLinkTracking::augmentSyncLinkTracking should not make make queries if there are no tracked fields


lekoala:
Recently found this by chance

It seems that since FileLinkTracking is attached to the dataobject, it runs for classes that are not relevant (eg: LoginSession)

<img width="1909" alt="image" src="https://github.com/silverstripe/silverstripe-assets/assets/250762/58db0080-5f30-4aa0-9e51-967c9bc32a6f">

By checking if there are actually any DBHTMLText, the query is not happening anymore

```php
        // Build a list of HTMLText fields, merging all linked pages together.
        $allFields = DataObject::getSchema()->fieldSpecs($this->owner);
        $linkedPages = [];
        $anyBroken = false;
        $hasTrackedFields = false;
        foreach ($allFields as $field => $fieldSpec) {
            $fieldObj = $this->owner->dbObject($field);
            if ($fieldObj instanceof DBHTMLText) {
                $hasTrackedFields = true;
                // Merge links in this field with global list.
                $linksInField = $this->trackLinksInField($field, $anyBroken);
                $linkedPages = array_merge($linkedPages, $linksInField);
            }
        }

        if (!$hasTrackedFields) {
            return;
        }
```

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/558
- https://github.com/silverstripe/silverstripe-cms/pull/2863

lekoala:
same thing for SiteTreeLinkTracking

maybe i'm missing something, but these changes seem to save two queries :-)

GuySartorelli:
Both PRs are merged. Thanks for that.

PR #558, opened 2023-06-22
Don't make query if not needed

lekoala:
setByIDList will make a query even though there are no DBHTMLText for this DataObject

Checking the presence of these fields allow to return early and prevent the needless query

Fixes https://github.com/silverstripe/silverstripe-assets/issues/557

### DIFF
--- a/src/Shortcodes/FileLinkTracking.php
+++ b/src/Shortcodes/FileLinkTracking.php
@@ -142,6 +142,7 @@ class FileLinkTracking extends DataExtension
             }
         }
 
+        // We cannot rely on linkedPages being empty, because we need to remove them if there was any
         if (!$hasTrackedFields) {
             return;
         }
--- a/src/Shortcodes/FileLinkTracking.php
+++ b/src/Shortcodes/FileLinkTracking.php
@@ -131,13 +131,19 @@ class FileLinkTracking extends DataExtension
         $allFields = DataObject::getSchema()->fieldSpecs($this->owner);
         $linkedPages = [];
         $anyBroken = false;
+        $hasTrackedFields = false;
         foreach ($allFields as $field => $fieldSpec) {
             $fieldObj = $this->owner->dbObject($field);
             if ($fieldObj instanceof DBHTMLText) {
+                $hasTrackedFields = true;
                 // Merge links in this field with global list.
                 $linksInField = $this->trackLinksInField($field, $anyBroken);
                 $linkedPages = array_merge($linkedPages, $linksInField);
             }
+        }
+
+        if (!$hasTrackedFields) {
+            return;
         }
 
         // Soft support for HasBrokenFile db field (e.g. SiteTree)


src/Shortcodes/FileLinkTracking.php
```diff
@@ -131,15 +131,21 @@ public function augmentSyncLinkTracking()
...
                 // Merge links in this field with global list.
                 $linksInField = $this->trackLinksInField($field, $anyBroken);
                 $linkedPages = array_merge($linkedPages, $linksInField);
             }
         }
 
+        if (!$hasTrackedFields) {
+            return;
+        }
+
```
michalkleiner: Can use the `$linkedPages` variable as that gets populated only when there are DBHTMLText fields

```suggestion
        foreach ($allFields as $field => $fieldSpec) {
            $fieldObj = $this->owner->dbObject($field);
            if ($fieldObj instanceof DBHTMLText) {
                $hasTrackedFields = true;
                // Merge links in this field with global list.
                $linksInField = $this->trackLinksInField($field, $anyBroken);
                $linkedPages = array_merge($linkedPages, $linksInField);
            }
        }

        if (empty($linkedPages)) {
            return;
        }

```

lekoala: @michalkleiner i don't think so, because maybe you had a linked page, but you don't anymore, and you want to remove them

michalkleiner: True, you could iterate over fields and still end up with an empty array, good call.

lekoala:
@michalkleiner  any other wishes for this PR ?

michalkleiner (approved):
I think this looks fine. What does @GuySartorelli think?

GuySartorelli:
Arguably this is an enhancement and shouldn't be targeting CMS 4 - but given the very low risk of regression it's probably fine.
I have no problem with this change, feel free to merge

src/Shortcodes/FileLinkTracking.php
```diff
@@ -131,15 +131,21 @@ public function augmentSyncLinkTracking()
...
             $fieldObj = $this->owner->dbObject($field);
             if ($fieldObj instanceof DBHTMLText) {
+                $hasTrackedFields = true;
                 // Merge links in this field with global list.
                 $linksInField = $this->trackLinksInField($field, $anyBroken);
                 $linkedPages = array_merge($linkedPages, $linksInField);
             }
         }
 
+        if (!$hasTrackedFields) {
```
ChloeSartorelli, line 146: Please add a comment here about why we don't just check `$linkedPages` - I was about to suggest the same thing that @michalkleiner had suggested in https://github.com/silverstripe/silverstripe-assets/pull/558#discussion_r1238430544

GuySartorelli (changes requested):

src/Shortcodes/FileLinkTracking.php
lekoala: done

GuySartorelli (approved):
Looks good, thanks

lekoala:
@GuySartorelli do you also apply this to the ss5 version or do i need to make a pr for that as well?

GuySartorelli:
@lekoala it'll get merged up, eventually. We're working on getting merge ups to be automated but for now they're just done ad hoc. If you need this immediately let me know and I'll make a point to do merge ups sooner rather than later.

lekoala:
@GuySartorelli i'm in the upgrade process to ss5 for my main client and that should go to production at the end of the summer, so if it's in a month or so that works for me :-)

Issue #559, opened 2023-06-22
AssetControlExtension should run less queries


lekoala:
Since it's an extension applied to all dataobject, extra caution should be taken in onBeforeWrite

https://github.com/silverstripe/silverstripe-assets/blob/e57099b0f1bfa63509307fe8da290c1dfe7a4ed1/src/AssetControlExtension.php#L72-L83

This kind of code can lead to this

<img width="1551" alt="image" src="https://github.com/silverstripe/silverstripe-assets/assets/250762/4ef49f56-717d-472d-87e2-256c09b6f371">

If i understand this correctly, it means we run an uncached select query, on each write, for any dataobject ?
that can't be right ? right ?

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/560

lekoala:
example solution provided as PR

GuySartorelli:
PR merged - it will be automatically tagged when CI has finished running on the branch.

PR #560, opened 2023-06-22
AssetControlExtension should run less queries

lekoala:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/559

### DIFF
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -217,12 +217,7 @@ class AssetControlExtension extends DataExtension
         }
     }
 
-    /**
-     * Check if the owner has assets
-     *
-     * @return bool
-     */
-    private function hasAssets()
+    private function hasAssets(): bool
     {
         $fields = DataObject::getSchema()->fieldSpecs($this->owner);
         foreach ($fields as $field => $db) {
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -221,7 +221,7 @@ class AssetControlExtension extends DataExtension
     {
         $fields = DataObject::getSchema()->fieldSpecs($this->owner);
         foreach ($fields as $field => $db) {
-            $fieldObj = $this->owner->$field;
+            $fieldObj = $this->owner->dbObject($field);
             if ($fieldObj instanceof DBFile) {
                 return true;
             }
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -222,7 +222,7 @@ class AssetControlExtension extends DataExtension
      *
      * @return bool
      */
-    protected function hasAssets()
+    private function hasAssets()
     {
         $fields = DataObject::getSchema()->fieldSpecs($this->owner);
         foreach ($fields as $field => $db) {
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -52,6 +52,10 @@ class AssetControlExtension extends DataExtension
      */
     public function onAfterDelete()
     {
+        if (!$this->hasAssets()) {
+            return;
+        }
+
         // Prepare blank manipulation
         $manipulations = new AssetManipulationList();
 
@@ -71,6 +75,10 @@ class AssetControlExtension extends DataExtension
      */
     public function onBeforeWrite()
     {
+        if (!$this->hasAssets()) {
+            return;
+        }
+
         // Prepare blank manipulation
         $manipulations = new AssetManipulationList();
 
@@ -207,6 +215,23 @@ class AssetControlExtension extends DataExtension
         foreach ($assets as $asset) {
             $manipulation->addAsset($asset, $state);
         }
+    }
+
+    /**
+     * Check if the owner has assets
+     *
+     * @return bool
+     */
+    protected function hasAssets()
+    {
+        $fields = DataObject::getSchema()->fieldSpecs($this->owner);
+        foreach ($fields as $field => $db) {
+            $fieldObj = $this->owner->$field;
+            if ($fieldObj instanceof DBFile) {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**


src/AssetControlExtension.php
```diff
@@ -209,6 +217,23 @@ protected function addAssetsFromRecord(AssetManipulationList $manipulation, Data
         }
     }
 
+    /**
+     * Check if the owner has assets
+     *
+     * @return bool
+     */
+    protected function hasAssets()
```
emteknetnz: This will need to be a private method so that it's not an API additin, which we can't do in 1.13 and we're not planning on releasing 1.14 now that CMS 5 is out.  You should also use a strongly typed :bool return and rather then the docbloxk

lekoala: i made it private. regarding the return type, that would be the only function with a strongly typed in the whole class, which i think is out of the scope. if the team wants to make specific style adjustements or syntax adjustement, i'll leave it up to you ;-)
i had a similar discussion here https://github.com/silverstripe/silverstripe-framework/pull/10834#pullrequestreview-1494413433 and i don't get why the pr should not follow the overall current style of the class. either the class use strong return type, or it doesn't. but mixing it doesn't feel right :-)

ChloeSartorelli: When adding new methods, properties, etc, our preference is that they be strongly typed where possible.  

In general if a request like that is made you'll find less friction to having your pr be merged if you make the requested change - yes it's a little inconsistent with existing API, but it's also more robust. Just because the existing code doesn't follow our current best practice doesn't mean the new code shouldn't either. 

We can't make everything strongly typed outside of a major release and we didn't have time to do that for CMS 5. We can add strong typing to new methods as we add them, though, so that's what we prefer to do. It will ultimately reduce the burden when we do go through with a comb and strongly type everything.

lekoala: ok ;-) works for me as well, i've updated the pr accordingly

GuySartorelli:
Have you done any profiling or other quantitative analysis that shows this as being a bottleneck, or which shows the PR as providing actual performance benefits?
This feels like the sort of thing that _could_ lead to regressions, so I'd like to see that it's actually providing a clear benefit before taking it any further.

lekoala:
@GuySartorelli i just saw it leads to unecessary queries

for example:
currently, the session-manager middleware update the LastAccessed property on each request (including ajax requests, so that's a ton of updates queries each time)
this extension applies to all dataobjects, even when they don't have assets, so it performs a select on **each write, for all dataobjects**. so, currently, we have one select and one update, for each request. i don't see how an extra select on each write cannot slow things down.

there is a similar issue with the SitetreeTracking and FileTracking extensions. they are applied to all dataobjects, even when there are no fields using shortcode in it.

any dataobject without files or htmltext fields are currently doing extra selects before writing for this reason, including base models like LoginSession.

i don't have a strict benchmark, but with all the recent PR i've made, pages are displaying twice as fast on average for my custom admin module.
(the login session being updated on each request being one really big offender)

michalkleiner:
Just a thought that crossed my mind now — it could be good to have some additional metadata around data models (or classes in general) maybe within the config system or the manifest where we could hold information like this (has files? has fields that can have shortcodes? other stuff that can only be changed via adding/removing extensions or via dev/build) and then use it for similar optimisations? Could be a heavy weight cannon for what's needed here (in this PR), but had to write it down.

lekoala:
@michalkleiner love the idea! from what I can see, the pattern is this one: some extensions need to be applied globally since you don't know which db fields would be used.
ideally, it would be great that based on the fields/relations, extension get applied (or not) to a model. that would allow to keep "lightweight" (all things considered :p) models without extensions that do not apply to them.

src/AssetControlExtension.php
```diff
@@ -209,6 +217,18 @@ protected function addAssetsFromRecord(AssetManipulationList $manipulation, Data
...
 
+    private function hasAssets(): bool
+    {
+        $fields = DataObject::getSchema()->fieldSpecs($this->owner);
+        foreach ($fields as $field => $db) {
+            $fieldObj = $this->owner->$field;
+            if ($fieldObj instanceof DBFile) {
+                return true;
+            }
+        }
```
ChloeSartorelli: This logic is sound for skipping `addAssetsFromRecord()` where `$this->owner` is passed in - it mirrors `findAssets()`.

However, `addAssetsFromOtherStages()` passes in a specific version of the record, which (correct me if I'm wrong) could have a `DBFile` instance on some field where the version represented by `$this->owner` has null.

`onBeforeWrite()` also has a clear distinction between `$this->owner` and the "prior record" fetched directly from the database - and in that case as well I imagine there could be some scenarios where the "current" record has a `null` where the "prior" record has a `DBFile` instance.

GuySartorelli (commented):
Looked a little deeper and again I'm worried about regressions here... it really does seem like there's risk with no quantitative reward. I'd feel a lot better about this if there was some quantitative analysis showing that this change does actually provide some clear, definable, and quantifiable benefits.

GuySartorelli:
@michalkleiner I like that idea a lot, as a general concept. I'm guessing that'd belong on `DataObjectSchema`?

michalkleiner:
> @michalkleiner I like that idea a lot, as a general concept. I'm guessing that'd belong on `DataObjectSchema`?

Yeah, probably, depends on what data it would hold and where it would live etc.
Could be config maybe, too, since what db fields and what relations are on each data model is defined via config.
Could be a build generated file that would get loaded similarly to graphql schema or similar.. just brainstorming. We probably should create an RFC ticket for that and there we might end up not having enough use cases for it and then it will end up forgotten or just a bolt on dataobjectschema.

src/AssetControlExtension.php
lekoala: @GuySartorelli i may be wrong here, but as far as i understand, fieldSpecs do not take into consideration versions, so checking the schema should return the same thing regardless

if we want to be 100% sure, we can also simply only wrap this part in my new hasAssets method

        // Mark overwritten object as deleted
        if ($this->owner->isInDB()) {
            $priorRecord = DataObject::get(get_class($this->owner))->byID($this->owner->ID);
            if ($priorRecord) {
                $this->addAssetsFromRecord($manipulations, $priorRecord, AssetManipulationList::STATE_DELETED);
            }
        }

this way, we get rid of the priorRecord call, which is the main issue, and add a todo for later to optimize things even further

lekoala:
For reference
https://github.com/silverstripe/silverstripe-framework/issues/10850

Do you want to proceed with my current approach ? as an alternative, we can simply avoid the priorRecord call instead of returning early, but i don't think it's necessary.

i don't see how my approach could go wrong : a dataobject with no assets should not make asset manipulations...

lekoala:
@GuySartorelli let me know if you need anything else from my end on this

GuySartorelli:
Sorry, I just haven't had time to look at this recently. It's on my radar but I can't give an estimate of when I'll have time to look at it again. If @michalkleiner of another core committer checks out your response to my concern and decides that they're happy with it, I'll have no concerns with them merging it - otherwise it'll just have to wait until someone makes time to look at it I'm afraid.

src/AssetControlExtension.php
ChloeSartorelli: `fieldSpecs()` will always return the same thing, but we're not checking against that, we're checking if `$this->owner->$field` is an instance of `DBFile` - in other words does the version of the record we currently have in memory have a `DBFile` as the _value_ for that field. It's checking what's stored in the db for a given version of a record, not just checking directly against the static schema.

If something like this works, it would make me a lot more at ease about it:

```php
private function hasAssets(): bool
{
    $fields = DataObject::getSchema()->fieldSpecs($this->owner);
    foreach ($fields as $field => $fieldSpec) {
        $fieldClass = strtok($fieldSpec ?? '', '(');
        if (singleton($fieldClass) instanceof DBFile) {
            return true;
        }
    }
    return false;
}
```

The above is based mostly from memory, so may need some tweaking - but basically instead of checking the value of the field on the class, just check purely what's in the field specs.

~~However, while writing this I realised another potential regression this could cause - definitely an edge case, but something to consider nevertheless. If someone puts a `DBFile` field wrapped inside a `DBComposite` for some reason (not likely but definitely not impossible), this will always say there's no assets, even when there is.
I'm not sure what the current behaviour is in that scenario though, so maybe that already doesn't work?~~
Nevermind, in the case of a `DBComposite`, `findAssets()` won't find the asset anyway. So the above should be fine in that regard.

ChloeSartorelli: @lekoala just `@`ing you here in case you missed the above comment.

GuySartorelli (changes requested):

src/AssetControlExtension.php
lekoala: @GuySartorelli i see your point. Using singleton creates other issues (here, it will fail on DBPrimaryKey because it has no arguments). I still think it's fine to use the current instance as a reference. dbObject will return an empty object anyway based on the schema specs, so this should end up being the same, with little to no extra work involved. i've changed my code to call dbObject directly to make sure we do get a proper class based on the specs

ChloeSartorelli: > dbObject will return an empty object anyway based on the schema specs

It won't necessarily be an _empty_ object - if the field has a value it just returns the field including the value. But it does importantly return a spec-based object regardless of whether the record we're calling it on has a value so I'm okay with this. Thank you for making the change.

GuySartorelli (approved):
Tested locally.
Results in 3 less queries on write for records with no assets, and no observable change for records which _do_ have assets.

Thanks for submitting this and for your patience.

lekoala:
@GuySartorelli :-) glad to help. it was really killing me to see those useless queries being made.

Issue #545, opened 2023-03-14
Error when previewing page and asset doesn't exist


phptek:
"[Emergency] Uncaught TypeError: Return value of SilverStripe\Assets\Shortcodes\FileShortcodeProvider::getCachedMarkup() must be of the type string, null returned"

Granted, this is a feature of `silverstripe/cms`, however the culprit is logic found in `silverstripe/assets`, specifically in `FileShortcodeProvider::getCachedMarkup()`

`return $item['markup'];` <-- can return null. A simple null coalesce operator "fixes" it ala `return $item['markup'] ?? ''`, but there may be issues in `silverstripe/cms` or `silverstripe/framework` where this array value should never be `null`. I think this has just surfaced when an explicit `:string` return type was naively added to the method.

Using Silverstripe `4.12.3` on PHP 7.4

PR
- https://github.com/silverstripe/silverstripe-assets/pull/562

GuySartorelli:
Thanks for reporting this.
Can you please give a clear set of steps to reproduce this error? You mention previewing a page when an asset doesn't exist - but you haven't described for example how the asset relates to the page?

GuySartorelli:
PR merged and tagged as [1.13.3](https://github.com/silverstripe/silverstripe-assets/releases/tag/1.13.3) and [2.0.4](https://github.com/silverstripe/silverstripe-assets/releases/tag/2.0.4)

PR #561, opened 2023-07-07
FIX getCachedMarkup returns NULL instead of string if file doesn't exist

sabina-talipova:
## Description
- Check if file exist before return cached link for this file from `getCachedMarkup` method.
- Method `handle_shortcode` should return string.  

## Test steps
- Set `FileShortcodeProvider::allow_session_grant = true`;
- Create new Page or open existing one in CMS;
- Click `Insert Link to a file` in Content field;
- Select any existing file and Save page;
- Remove file from backend or change link in DB record;
- Open Page in CMS again;
- You should not see any Errors;
- Shortcode link should lead to "Home" page;  
 
## Parent issue
- https://github.com/silverstripe/silverstripe-assets/issues/545

### DIFF
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -102,6 +102,8 @@ class FileShortcodeProvider implements ShortcodeHandler, Flushable
             $url = $record->Link();
         }
 
+        $url = $url ?? '';
+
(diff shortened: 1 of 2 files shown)


GuySartorelli:
The bug was reported for CMS 4 - please retarget this PR to the `1.13` branch.

PR #562, opened 2023-07-11
FIX getCachedMarkup returns NULL instead of string if file doesn't exist

sabina-talipova:
## Description
- Check if file exist before return cached link for this file from `getCachedMarkup` method.
- Method `handle_shortcode` should return string.  

## Test steps
- Set `FileShortcodeProvider::allow_session_grant = true`;
- Create new Page or open existing one in CMS;
- Click `Insert Link to a file` in Content field;
- Select any existing file and Save page;
- Remove file from backend or change link in DB record;
- Open Page in CMS again;
- You should not see any Errors;
- Shortcode link should lead to "Home" page;  
 
## Parent issue
- https://github.com/silverstripe/silverstripe-assets/issues/545

### DIFF
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -140,7 +140,7 @@ class FileShortcodeProvider implements ShortcodeHandler, Flushable
     {
         $item = $cache->get($cacheKey);
         $assetStore = Injector::inst()->get(AssetStore::class);
-        if ($item && !empty($item['filename']) && $item['markup']) {
+        if ($item && $item['markup'] && !empty($item['filename'])) {
             // Initiate a protected asset grant if necessary
             $allowSessionGrant = static::getGrant(null, $arguments);
             if ($allowSessionGrant && $assetStore->exists($item['filename'], $item['hash'])) {
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -139,11 +139,15 @@ class FileShortcodeProvider implements ShortcodeHandler, Flushable
     protected static function getCachedMarkup($cache, $cacheKey, $arguments): string
     {
         $item = $cache->get($cacheKey);
-        if ($item && !empty($item['filename'])) {
+        $assetStore = Injector::inst()->get(AssetStore::class);
+        if ($item
+            && !empty($item['filename'])
+            && $assetStore->exists($item['filename'], $item['hash'])
+            && $item['markup']) {
             // Initiate a protected asset grant if necessary
             $allowSessionGrant = static::getGrant(null, $arguments);
             if ($allowSessionGrant) {
-                Injector::inst()->get(AssetStore::class)->grant($item['filename'], $item['hash']);
+                $assetStore->grant($item['filename'], $item['hash']);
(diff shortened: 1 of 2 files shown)


src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -102,6 +102,8 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
             $url = $record->Link();
         }
 
+        $url = $url ?? '';
+
```
ChloeSartorelli: I've tested without this and it doesn't seem to change anything... seems to not be needed?

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -139,11 +141,12 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
     protected static function getCachedMarkup($cache, $cacheKey, $arguments): string
     {
         $item = $cache->get($cacheKey);
-        if ($item && !empty($item['filename'])) {
+        $assetStore = Injector::inst()->get(AssetStore::class);
+        if ($item && !empty($item['filename']) && $assetStore->exists($item['filename'], $item['hash'])) {
```
ChloeSartorelli:
```suggestion
        if ($item && !empty($item['filename'])) {
```
Move this below (see https://github.com/silverstripe/silverstripe-assets/pull/562#discussion_r1259092280) - we shouldn't hit the filesystem if we're not even allowing the grant.

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -139,11 +141,12 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
...
             // Initiate a protected asset grant if necessary
             $allowSessionGrant = static::getGrant(null, $arguments);
             if ($allowSessionGrant) {
```
ChloeSartorelli:
```suggestion
            if ($allowSessionGrant && $assetStore->exists($item['filename'], $item['hash'])) {
```
Moved as per https://github.com/silverstripe/silverstripe-assets/pull/562#discussion_r1259092223

GuySartorelli (changes requested):

src/Shortcodes/FileShortcodeProvider.php
sabina-talipova: `File::getURL()` returns null https://github.com/silverstripe/silverstripe-assets/blob/7605cda151919e8638c3b300e4864d6c3748c349/src/File.php#L896.
And then we set this value in `$markup` in cache. After we return $item['markup'] in `getCachedMarkup` method.
So probably, there are some cases where we can get null from `getCachedMarkup`
I'll add checking in the `getCachedMarkup` method.

sabina-talipova: Updated

src/Shortcodes/FileShortcodeProvider.php
sabina-talipova: Updated

src/Shortcodes/FileShortcodeProvider.php
sabina-talipova: Updated

GuySartorelli (approved):

PR #565, opened 2023-08-07
FIX Multi HTML entities in image shortcodes

sabina-talipova:
## Description
- Turn off double_encode for image attributes in ImageShortcodeProvider
## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10451

### DIFF
--- a/src/Shortcodes/ImageShortcodeProvider.php
+++ b/src/Shortcodes/ImageShortcodeProvider.php
@@ -5,8 +5,6 @@ namespace SilverStripe\Assets\Shortcodes;
 use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Image;
-use SilverStripe\Assets\Storage\AssetStore;
-use SilverStripe\Core\Convert;
 use SilverStripe\Core\Flushable;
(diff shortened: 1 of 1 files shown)


src/Shortcodes/ImageShortcodeProvider.php
ChloeSartorelli, line 1: Also remove the `use SilverStripe\Core\Convert;` at the top of the file, since it isn't being used anymore.

GuySartorelli (changes requested):
This change looks okay and seems to be necessary. Just one small change here.

GuySartorelli:
Please also add a behat test for this - or a unit test if you can find a way to do that cleanly.

src/Shortcodes/ImageShortcodeProvider.php
sabina-talipova: Updated. 
I removed `use SilverStripe\Assets\Storage\AssetStore;` we don't use it as well.

sabina-talipova:
Add behat test in asset-admin https://github.com/silverstripe/silverstripe-asset-admin/pull/1376

GuySartorelli (approved):
LGTM, will merge with the related PRs.

PR #567, opened 2023-08-10
FIX Double encoding image shortcode attributes

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10451


GuySartorelli (approved):
Looks good, all works as expected.

Issue #512, opened 2022-09-01
No prevention of duplicated file names when moving a file to a new folder


sb-relaxt-at:
I have encountered the following issue (tested in 1.10 but should be the same in 1.11): 

1. Create a folder "foo"
2. Upload a file a.jpg to assets (id 2)
3. Upload a different file a.jpg to assets/foo (id 3)
4. Move the file from assets/a.jpg to assets/foo/a.jpg (for example via setting a new parent in the user interface)

As long as both files are not published everything is fine (as they have different hashes), but as soon as the files get published it will no longer work and one of the files will be overwritten. We will have the following db entries, whereas only one will generate a working URL:

* ID 3: Name "a.jpg", FileFilename: "foo/a.jpg"
* ID 4: Name "a-v2.jpg" FileFilename: "foo/a.jpg"

In the filesystem only "foo/a.jpg" exists.

If both files do have the same content it is almost the same, but both files will generate a working URL. 

From my understanding the file name should be also made unique if the ParentID has changed, adding a condition in [File#onBeforeWrite](https://github.com/silverstripe/silverstripe-assets/blob/1.11/src/File.php#L693).

## Acceptance criteria
- When a content author changes the location of a file via the location dropdown it should give hem the same results as moving the file with the mouse.

## Note
- If a file is already in this state, we don't currently have a good way of fixing it. We've decided that this is not something we can practically fix.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/568

maxime-rainville:
Validated this.

It looks like if you drag tho folder around with the mouse, assets is smart enough to rename the file for you.

However, if you do it via the "Location" dropdown, you end up wit two files with the same name.

sb-relaxt-at:
Thank you for looking into it, I almost forgot about it…

GuySartorelli:
PR tagged. This will be automatically tagged when CI has finished running on the branch.

PR #568, opened 2023-08-15
FIX Ensure filenames are not duplicated when moving

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-assets/issues/512

Created new files test-image-without-title.png + test-image-metacharacters.png because the new logic was causing the duplicate files to end up with -v2.png / -v3.png suffixes and breaking other unrelated unit tests


michalkleiner (approved):

GuySartorelli (approved):
Looks good, works great locally.

PR #579, opened 2023-10-20
ENH add optional SiteConfig root level permissions

### LINKED ISSUE #113: Root-level assets permissions missing from SiteConfig
`SiteTree` has a root-level permission configurable in `SiteConfig`:

image

There should be a similar field for assets - in particular as any assets stored in the root have `Inherit` permissions set by default and are inheriting nothing by default.

Would make sense for it to match the per-file permissions:

image

andrewandante:
Fixes #113

Offers a SiteConfig-based approach to Root file permissions, in a similar vein to Pages.

Aims to have the same starting point as the current `FileDefaultPermissions` - that is, view by anyone, edit only by those with the Permission `File::EDIT_ALL`.

Can be configured with the following yaml (caveat, maybe not all that is needed, IDK):

```yml
---
Name: app-assetspermissions
After:
 - assetspermissions
---
  SilverStripe\Security\PermissionChecker.file:
    class: SilverStripe\Security\InheritedPermissions
    constructor:
      BaseClass: SilverStripe\Assets\File
      CacheService: '%$Psr\SimpleCache\CacheInterface.InheritedPermissions'
    properties:
      DefaultPermissions: '%$SilverStripe\Assets\SiteConfigFilePermissions'
      GlobalEditPermissions:
        - CMS_ACCESS
  SilverStripe\Security\InheritedPermissionFlusher:
    properties:
      Services:
        - '%$SilverStripe\Security\PermissionChecker.file'
SilverStripe\SiteConfig\SiteConfig:
  extensions:
    - SilverStripe\Assets\RootLevelAccessSiteConfigExtension
```

## TODO

- probably remove SiteConfig as a dependency, this shouldn't be invoked unless it is installed
- hide things properly on the page (doing that thing where I need to adjust some React code or all the options are open
- update Readme/Docs to offer this option

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "require": {
         "php": "^8.1",
         "silverstripe/framework": "^5.1",
+        "silverstripe/siteconfig": "^5.1",
         "silverstripe/vendor-plugin": "^2",
         "symfony/filesystem": "^6.1",
         "intervention/image": "^2.7.2",
--- a/src/File.php
+++ b/src/File.php
@@ -105,6 +105,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
      * Permission for edit all files
      */
     const EDIT_ALL = 'FILE_EDIT_ALL';
+    const GRANT_ACCESS = 'FILE_GRANT_ACCESS';
 
     private static $default_sort = "\"Name\"";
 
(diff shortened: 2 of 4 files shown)


GuySartorelli:
I'm not sure if SiteConfig is the right place for this - it would make more sense to me if instead there was a way to configure this on the root 'folder' in asset admin itself. There's no reason for this to be tied to SiteConfig other than that's where sitetree handles _its_ root permissions, which makes sense in that context because:
1. SiteConfig is already required by the CMS module
2. For pages, those permissions are about access to the site

Asset admin shouldnt have any dependency on SiteConfig, and access to files I think is sufficiently separate from site access that it should be contained withing the asset admin directly.

andrewandante:
Fair - from a user's perspective, I think that's where I'd expect it to be, but equally we can't put it there without SiteConfig, so I think in this case it should go somewhere else. Thinking here? 

image

Could pop it out like the current "folder" Edit Form, with only a permissions "tab"?

GuySartorelli:
Yup, that'd make sense to me.
Or perhaps just add an edit icon next to "Files" like when you're editing a folder - since that's essentially what we're doing here.
edit button next to folder name
And then as you say, only have the permissions tab in the form.

andrewandante:
Oh yeah that's clean, I like that a lot. Now to figure out how to do it 😉

GuySartorelli:
In the meantime I'll close this PR - please create a new one when/if you have a way forward on the new approach.

Issue #572, opened 2023-10-10
Uploading unknown filetypes is no longer possible


edwilde:
This appears to be caused by `league/flysystem` v3 changing to throwing exceptions rather than returning `null` - see [a similar issue in laravel](https://github.com/laravel/framework/pull/42761).

I have added the appropriate configurations to allow `brf` (braille format) files, [as per the documentation](https://docs.silverstripe.org/en/5/developer_guides/files/allowed_file_types/#mime-type-validation).

This works just fine in CMS 4.13, but fails in CMS 5.0 with the following error:
```
error-log.ERROR: Uncaught Exception League\Flysystem\UnableToRetrieveMetadata: "Unable to retrieve the mime_type for file at location: c7481fdf07/brf-sample.brf. " at /var/www/mysite/www/vendor/league/flysystem/src/UnableToRetrieveMetadata.php line 49 {"exception":"[object] (League\\Flysystem\\UnableToRetrieveMetadata(code: 0): Unable to retrieve the mime_type for file at location: c7481fdf07/brf-sample.brf.  at /var/www/mysite/www/vendor/league/flysystem/src/UnableToRetrieveMetadata.php:49)"} []
```

The exception is raised on [the flysystem adapter](https://github.com/thephpleague/flysystem/blob/bd4c9b26849d82364119c68429541f1631fba94b/src/Local/LocalFilesystemAdapter.php#L397) and unhandled in the [Silverstripe Asset store class](https://github.com/silverstripe/silverstripe-assets/blob/2/src/Flysystem/FlysystemAssetStore.php#L1101).

That was fixed, released, and then rolled back. It is now an opt-in to fix that behaviour by passing a new argument to the constructor.

## Acceptance Criteria
- Pass `true` to the new opt-in argument in the `FallbackMimeTypeDetector` constructor.

## NOTE
Once this has been resolved, we should raise a new PR to add https://github.com/silverstripe/silverstripe-assets/issues/573 to the changelog.
No sense adding it until this is resolved.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/582

GuySartorelli:
Turns out this is a bug upstream: https://github.com/thephpleague/flysystem/issues/1468
I've raised a PR to fix it: https://github.com/thephpleague/flysystem/pull/1710

I'll keep this issue open in the meantime - if they decide they don't consider it to be a bug, we may need to work around it.

GuySartorelli:
[...]

PR #580, opened 2023-11-21
NEW: Add the ability to specify the MimeTypeDetector for the LocalFilesystemAdapter

madmatt:
The `AssetAdapter` class has a really weird constructor which has made this super confusing.

`SilverStripe\Assets\Flysystem\AssetAdapter` extends `SilverStripe\Assets\Flysystem\LocalFilesystemAdapter` which in turn extends `League\Flysystem\Local\LocalFilesystemAdapter` but the constructors are totally different.

 * `AssetAdapter` takes arguments `string? $root`, `int $writeFlags`, `int $linkHandling`
 * The Flysystem `LocalFilesystemAdapter` takes arguments `string $location`, `VisibilityConverter $visibility`, `int $writeFlags`, `int $linkHandling`, `MimeTypeDetector $mimeTypeDetector`, `bool $lazyRootCreation`.

Basically, I have a weird file type (`*.brf` files which is a Braille file format used for physical Braille devices that convert words into Braille. It's essentially a fancy ASCII document, but `finfo` detects it as `text/plain` which according to Flysystem is an "inconclusive file type" and therefore is not a valid mime type.

Not sure if this requires documentation or not, happy to make a PR to add docs to the `developer-docs` repo if you think it does and I can link them here.

Here's how you use it (to allow text/plain for example):

In app/_config/assets.yml:
```yml
# Use the FinfoMimeTypeDetector directly instead of the Fallback. We remove text/plain
SilverStripe\Assets\Flysystem\AssetAdapter:
  default_mime_type_detector: '%$League\MimeTypeDetection\FinfoMimeTypeDetector'
  League\MimeTypeDetection\FinfoMimeTypeDetector:
    constructor:
      MagicFile: ''
      ExtensionMap: null
      BufferSampleSize: null
      InconclusiveMimeTypes:
        - 'application/x-empty'
        - 'text/x-asm'
        - 'application/octet-stream'
        - 'inode/x-empty'
```

### DIFF
--- a/src/Flysystem/AssetAdapter.php
+++ b/src/Flysystem/AssetAdapter.php
@@ -5,10 +5,12 @@ namespace SilverStripe\Assets\Flysystem;
 use League\Flysystem\Config as FlysystemConfig;
 use League\Flysystem\UnableToWriteFile;
 use League\Flysystem\UnixVisibility\PortableVisibilityConverter;
+use League\MimeTypeDetection\MimeTypeDetector;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Filesystem;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\SSViewer;
@@ -35,6 +37,16 @@ class AssetAdapter extends LocalFilesystemAdapter
      * @var string
      */
     private static $default_server = 'apache';
+
+    /**
+     * Default mime type detector to use. If not defined, the default mime type detector from Flysystem will be used.
+     * If you want to define one, this must be the name of a class that implements the interface
+     * League\MimeTypeDetection\MimeTypeDetector
+     *
+     * @config
+     * @var string
+     */
+    private static $default_mime_type_detector;
 
     /**
      * Config compatible permissions configuration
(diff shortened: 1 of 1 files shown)


madmatt:
Also, I can't figure out a good way to write a test for this to be honest, without injecting some custom file into the repo which feels a bit overkill. Thoughts welcome on how to test that, or if it's worth it to do so.

michalkleiner:
Is there not a simple way how to treat a file based on file extension and skip all the guesswork?

madmatt:
By default Flysystem uses the `finfo` mimetype checker. There's a separate `ExtensionMimeTypeDetector` which does it based on file extension but that is just doing a 'dumb' lookup between file extension and mime type so ultimately it's worse than `finfo` in the majority of cases because it doesn't actually look at the file contents, only the extension, which I'd say is a pretty big security risk.

michalkleiner:
Yeah, I was thinking just for the inconclusives as a fallback, not to have it as the primary/default detector. But that might be even more complicated than what you've already achieved.

GuySartorelli:
For your specific use case of the brf file, see https://github.com/silverstripe/silverstripe-assets/issues/572 - a bug in flysytem resulted in "inconclusive" file types returning `null` instead of the detected file type, which resulted in an exception. A PR to fix that has been accepted and merged, and once that is tagged I think that will solve your use case.

I do agree though that improved flexibility here could be potentially beneficial. I haven't looked properly at this PR, just thought you'd like to know that there _is_ a solution waiting to be tagged for the use case you mentioned in the description.

GuySartorelli:
@madmatt https://github.com/silverstripe/silverstripe-assets/pull/582 has just been merged and tagged - do you still have a use case for this change?

madmatt:
Nope I don't think so, thanks @GuySartorelli! I'll look at updating to the tag and make sure it still works on our side. Thanks for investigating!

PR #582, opened 2023-12-18
FIX Allow inconclusive mimetypes

GuySartorelli:
I don't want to bump the dependency in a patch release so we can't use a named argument for the new `useInconclusiveMimeTypeFallback` parameter since it isn't available in flysystem until [3.23.0](https://github.com/thephpleague/flysystem/compare/3.22.0...3.23.0).

Passing an extra argument doesn't cause errors so it's safe if the project doesn't have flysystem updated - but passing missing named arguments _does_ cause errors.

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/572

### DIFF
--- a/src/Flysystem/LocalFilesystemAdapter.php
+++ b/src/Flysystem/LocalFilesystemAdapter.php
@@ -19,7 +19,8 @@ class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter
         MimeTypeDetector $mimeTypeDetector = null
     ) {
         $this->pathPrefixer = new PathPrefixer($location);
-        parent::__construct($location, $visibility, $writeFlags, $linkHandling, $mimeTypeDetector);
+
+        parent::__construct($location, $visibility, $writeFlags, $linkHandling, $mimeTypeDetector, false, true);
     }
 
     public function prefixPath(string $path): string


emteknetnz (changes requested):
If I understand this correctly, if mimetype cannot be worked out from the file contents, then it will fallback to using the file extension to determine the mimetype?  Isn't this introducing a bit of a security risk since it would provide a vector to upload files that don't match the file extenion?

GuySartorelli:
It tries to get a mimetype from the file contents - but even if it _gets_ a mimetype _and it's a valid mimetype_, it will ignore that type if it's deemed to be "inconclusive".
If then checks the file name - and if it can't get any mimetype from the file name (inconclusive or otherwise), an exception is thrown.

So for `.brf` files for example, the file contents look like `text/plain`. That's considered inconclusive, so it moves on to check the file type, and there is no registered mime type for `.brf` in flysystem, so it throws an exception.

If it returned `text/plain` (which this PR will make it do), that mimetype is _still validated_ afterwards. In the case of `.brf` files, the [mimevalidator was updated](https://github.com/silverstripe/silverstripe-mimevalidator/pull/68) with the correct mime type, which is one of `text/plain` or `application/octet-stream`. If the mime type detected isn't one of those, it'll still fail validation.

So long as the mimetype validator is set up correctly for a given file type (which AFAIK is required to include it as uploadable) there is no risk of vulnerabilities of that type.

emteknetnz (approved):
OK sounds good

Issue #488, opened 2022-05-03
Allow file variant with different extension


maxime-rainville:
As a developer, I want to be able to create variant of files with different extensions.

## Acceptance criteria
- [x] File URL of variants can still be resolved back to their parent file regardless of whether they have a different extension. (View FileIDParser)
- [x] You can still have file with the same name and a different extension and avoid collision.
- [x] Normal file operation like session grant, deletion, move, publication or archival apply to variant with different extensions as well.
- [x] There's an API for creating variant with alternative extensions.
- [x] Variants with different extensions can be created for regular files as well as image.
- [x] All Image manipulation functions work for variants with alternative extensions if they map to an image format.
- [x] All Image manipulation functions can be chained when applied to an image variant with a different extensions

## Follow up task 
We've decided te spin off the follow AC as it's own card:
- Variants of Variants are handled in a sensible manner and avoid needless duplication.

This is meant to illustrate what "Variant of Variant" means. Note that those image conversion methods won't exist right away.
  - `$JpegImage.Png().Jpeg()` doesn't throw an error or create a duplicate of the original image.
  - `$JpegImage.Png().WebP()` doesn't throw an error, doesn't use an intermediary step and creates the same file has calling `$JpegImage.WebP()`.
  - `$JpegImage.FitMax(100,100).WebP()` results in exactly the same file as `$JpegImage.WebP().FitMax(100,100)`

## Notes
- There's a [proof of concept](https://github.com/silverstripe/silverstripe-assets/pull/411) that implements all of these concepts.
  - There's a low level `manipulate` API for creating variant with alternative extensions (something similar to [`manipulateImage`](https://github.com/silverstripe/silverstripe-assets/blob/e5ca733eaaab580278e5e4325c55a3ef7b7fa368/src/ImageManipulation.php#L852))
- What happens when you upload `image1.jpg`, which has a webp variant, and then you upload a separate `image1.webp`? When you try to download `https://example.com/assets/image1.webp` which of the files do you get? Or does the existence of `image1.jpg` stop you from uploading `image1.webp`?
  - The URL for the variant webp is not just `assets/image1.webp` so there should be no conflict, but this should be validated.

## Excluded
- Generic `ToImage` API that would attempt to create Thumbnails for any file type.
- Higher level Converter service API ... that will come later (see https://github.com/silverstripe/silverstripe-assets/issues/489)

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/587
[...]

PR #585, opened 2024-01-16
NEW Allow file variants with different extensions

GuySartorelli:
## Description

Provides a low-level API for generating a file variant which has a different extension than the original file.
e.g. could be used for:
- Generating thumbnails for videos/documents/etc
- Converting .png or .jpg to .webp
- Converting .docx to .pdf

## Manual testing steps

1. Upload a video file.
1. Upload a document file.
1. Use this code for `PageController`.
    Replace `/var/www/html/.eddev/samples/snickers.jpg` with the absolute path to some arbitrary image file your site can access (or replace that logic with something to actually generate a file from the video - I'm taking some shortcuts in that regard)
    Replace `/var/www/html/.eddev/samples/client-pdf2.pdf` with the absolute path to some arbitrary pdf file your site can access (or replace that logic with something to actually generate a pdf from the document - I'm taking some shortcuts in that regard)
    ```php
    use SilverStripe\Assets\Storage\AssetStore;
    use SilverStripe\Assets\File;
    use SilverStripe\Assets\Image_Backend;
    use SilverStripe\CMS\Controllers\ContentController;
    use SilverStripe\Core\Injector\Injector;
    
    class PageController extends ContentController
    {
        public function getVideoImage()
        {
            $file = File::get()->find('FileFilename:EndsWith', File::get_category_extensions('video'));
            return $file->manipulateExtension('jpg', function (AssetStore $store, string $filename, string $hash, string $variant) {
                $backend = Injector::inst()->create(Image_Backend::class);
                $backend->loadFrom('/var/www/html/.eddev/samples/snickers.jpg');
                $tuple = $backend->writeToStore($store, $filename, $hash, $variant, ['conflict' => AssetStore::CONFLICT_USE_EXISTING]);
                return [$tuple, $backend];
            });
        }

        public function getPdfFromDoc()
        {
            $file = File::get()->find('FileFilename:EndsWith', File::get_category_extensions('document'));
            return $file->manipulateExtension('pdf', function (AssetStore $store, string $filename, string $hash, string $variant) {
                $tuple = $store->setFromString(
                    file_get_contents('/var/www/html/.eddev/samples/client-pdf2.pdf'),
                    $filename,
                    $hash,
                    $variant,
                    ['conflict' => AssetStore::CONFLICT_USE_EXISTING]
                );
                return [$tuple, null];
            });
        }
    }
    ```
1. Add this to a page template:
    ```
    <a href="$VideoImage.ScaleMaxHeight(200).CropHeight(50).Link">Link to the resized, cropped thumbnail</a> - ID=$VideoImage.ID
    <br>
    <a href="$PdfFromDoc.Link">Link to the "converted" pdf</a> - ID=$PdfFromDoc.ID
    ```
1. Visit a page and click on the links.
    - For the image, you should see an appropriately scaled and cropped copy of your dummy thumbnail image.
    - For the pdf, you should see the pdf file you pointed your code to.
    - The URL should be obviously a variant of the original file
    - The displayed ID numbers should be the original file IDs

You can also try the [documentation example](https://github.com/silverstripe/developer-docs/pull/441) of converting one image to another image (e.g. a `.jpg` to `.webp`) and see that it also works correctly, even with chained image manipulations.

## Issues

- https://github.com/silverstripe/silverstripe-assets/issues/488

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/FilenameParsing/AbstractFileIDHelper.php
+++ b/src/FilenameParsing/AbstractFileIDHelper.php
@@ -2,11 +2,28 @@
 
 namespace SilverStripe\Assets\FilenameParsing;
 
+use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Path;
 
 abstract class AbstractFileIDHelper implements FileIDHelper
 {
     use Injectable;
+
+    /**
+     * A variant type for encoding a variant filename with a different extension than the original.
+     */
+    public const EXTENSION_REWRITE_VARIANT = 'ExtRewrite';
+
+    /**
+     * Use the original file's extension
+     */
+    protected const EXTENSION_ORIGINAL = 0;
+
+    /**
+     * Use the variant file's extension
+     */
+    protected const EXTENSION_VARIANT = 1;
 
     public function buildFileID($filename, $hash = null, $variant = null, $cleanfilename = true)
     {
@@ -21,6 +38,10 @@ abstract class AbstractFileIDHelper implements FileIDHelper
         // Since we use double underscore to delimit variants, eradicate them from filename
         if ($cleanfilename) {
             $filename = $this->cleanFilename($filename);
+        }
+
+        if ($variant) {
+            $filename = $this->swapExtension($filename, $variant, self::EXTENSION_VARIANT);
         }
 
         $name = basename($filename ?? '');
@@ -51,6 +72,48 @@ abstract class AbstractFileIDHelper implements FileIDHelper
         }
 
         return $fileID;
+    }
+
+    /**
+     * Get the original file's filename with the extension rewritten to be the same as either the original
+     * or the variant extension.
+     *
+     * @param string $filename Original filename without variant
+     * @param int $extIndex One of self::EXTENSION_ORIGINAL or self::EXTENSION_VARIANT
+     */
+    protected function swapExtension(string $filename, string $variant, int $extIndex): string
+    {
+        // If there's no variant at all, we can rewrite the filenmane
+        if (empty($variant)) {
+            return $filename;
+        }
+
+        // Split variant string in variant list
+        $subVariants = explode('_', $variant);
+
+        // Split our filename into a filename and extension part
+        $fileParts = pathinfo($filename);
+        if (!isset($fileParts['filename']) || !isset($fileParts['extension'])) {
+            return $filename;
+        }
+        $dirname = $fileParts['dirname'] !== '.' ? $fileParts['dirname'] : '';
+        $filenameWithoutExtension = Path::join($dirname, $fileParts['filename']);
+        $extension = $fileParts['extension'];
+
+        // Loop our variant list until we find our special file extension swap variant
+        // Reverse the list first so the variant extension we find is the last extension rewrite variant in a chain
+        $extSwapVariant = preg_quote(self::EXTENSION_REWRITE_VARIANT, '/');
+        foreach (array_reverse($subVariants) as $subVariant) {
+            if (preg_match("/^$extSwapVariant(?<base64>.+)$/", $subVariant, $matches)) {
+                // This array always contain 2 values: The original extension at index 0 and the variant extension at index 1
+                /** @var array $extensionData */
+                $extensionData = Convert::base64url_decode($matches['base64']);
+                $extension = $extensionData[$extIndex];
+                break;
+            }
+        }
+
+        return $filenameWithoutExtension . '.' . $extension;
     }
 
     public function cleanFilename($filename)
--- a/src/FilenameParsing/HashFileIDHelper.php
+++ b/src/FilenameParsing/HashFileIDHelper.php
@@ -30,10 +30,16 @@ class HashFileIDHelper extends AbstractFileIDHelper
         }
 
         $filename = $matches['folder'] . $matches['basename'] . $matches['extension'];
+        $variant = $matches['variant'] ?: '';
+
+        if ($variant) {
+            $filename = $this->swapExtension($filename, $variant, self::EXTENSION_ORIGINAL);
+        }
+
         return new ParsedFileID(
             $filename,
             $matches['hash'],
-            isset($matches['variant']) ? $matches['variant'] : '',
+            $variant,
             $fileID
         );
     }
--- a/src/FilenameParsing/NaturalFileIDHelper.php
+++ b/src/FilenameParsing/NaturalFileIDHelper.php
@@ -2,8 +2,6 @@
 
 namespace SilverStripe\Assets\FilenameParsing;
 
-use SilverStripe\Core\Injector\Injectable;
-
 /**
  * Parsed Natural path URLs. Natural path is the same hashless path that appears in the CMS.
  *
@@ -23,10 +21,16 @@ class NaturalFileIDHelper extends AbstractFileIDHelper
         }
 
         $filename = $matches['folder'] . $matches['basename'] . $matches['extension'];
+        $variant = $matches['variant'] ?: '';
+
+        if ($variant) {
+            $filename = $this->swapExtension($filename, $variant, self::EXTENSION_ORIGINAL);
+        }
+
         return new ParsedFileID(
             $filename,
             '',
-            isset($matches['variant']) ? $matches['variant'] : '',
+            $variant,
             $fileID
         );
     }
--- a/src/ImageBackendFactory.php
+++ b/src/ImageBackendFactory.php
@@ -37,9 +37,15 @@ class ImageBackendFactory implements Factory
      */
     public function create($service, array $params = [])
     {
-        /** @var AssetContainer $assetContainer */
+        /** @var AssetContainer|null $assetContainer */
         $assetContainer = reset($params);
-        if (!$assetContainer instanceof AssetContainer) {
+
+        // If no asset container was passed in, create a new uncached image backend
+        if (!$assetContainer) {
+            return $this->creator->create($service, $params);
+        }
+
+        if (!($assetContainer instanceof AssetContainer)) {
             throw new BadMethodCallException("Can only create Image_Backend for " . AssetContainer::class);
         }
 
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Assets;
 
 use InvalidArgumentException;
 use LogicException;
+use SilverStripe\Assets\FilenameParsing\AbstractFileIDHelper;
 use SilverStripe\Assets\Storage\AssetContainer;
 use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Assets\Storage\DBFile;
@@ -841,12 +842,24 @@ trait ImageManipulation
     }
 
     /**
+     * Wrapper for manipulate() that creates a variant file with a different extension than the original file.
+     *
+     * @return DBFile|null The manipulated file
+     */
+    public function manipulateExtension(string $newExtension, callable $callback)
+    {
+        $pathParts = pathinfo($this->getFilename());
+        $variant = $this->variantName(AbstractFileIDHelper::EXTENSION_REWRITE_VARIANT, $pathParts['extension'], $newExtension);
+        return $this->manipulate($variant, $callback);
+    }
+
+    /**
      * Wrapper for manipulate that passes in and stores Image_Backend objects instead of tuples
      *
      * @param string $variant
      * @param callable $callback Callback which takes an Image_Backend object, and returns an Image_Backend result.
      * If this callback returns `true` then the current image will be duplicated without modification.
-     * @return DBFile The manipulated file
+     * @return DBFile|null The manipulated file
      */
     public function manipulateImage($variant, $callback)
     {
@@ -914,7 +927,7 @@ trait ImageManipulation
      * This callback will be passed the backend, filename, hash, and variant
      * This will not be called if the file does not
      * need to be created.
-     * @return DBFile The manipulated file
+     * @return DBFile|null The manipulated file
      */
     public function manipulate($variant, $callback)
     {
@@ -999,6 +1012,7 @@ trait ImageManipulation
      * For legacy reasons, there's no delimiter between this part, and the encoded arguments.
      * This means we have to use a whitelist of "known formats", based on methods
      * available on the {@link Image} class as the "main" user of this trait.
+     * The one exception to this is the variant for swapping file extensions, which is explicitly allowed.
      * This class is commonly decorated with additional manipulation methods through {@link DataExtension}.
      *
      * @param $variantName
@@ -1007,10 +1021,11 @@ trait ImageManipulation
      */
     public function variantParts($variantName)
     {
-        $methods = array_map('preg_quote', singleton(Image::class)->allMethodNames() ?? []);
+        $allowedVariantTypes = array_map('preg_quote', singleton(Image::class)->allMethodNames() ?? []);
+        $allowedVariantTypes[] = preg_quote(AbstractFileIDHelper::EXTENSION_REWRITE_VARIANT);
 
         // Regex needs to be case insensitive since allMethodNames() is all lowercased
-        $regex = '#^(?<format>(' . implode('|', $methods) . '))(?<encodedargs>(.*))#i';
+        $regex = '#^(?<format>(' . implode('|', $allowedVariantTypes) . '))(?<encodedargs>(.*))#i';
         preg_match($regex ?? '', $variantName ?? '', $matches);
 
         if (!$matches) {
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -1037,7 +1037,7 @@ trait ImageManipulation
             throw new InvalidArgumentException('Invalid variant name arguments: ' . $variantName);
         }
 
-        return array_merge([$matches['format']], $args[0]);
+        return array_merge([$matches['format']], $args);
     }
 
     /**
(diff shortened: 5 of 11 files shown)


src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -14,70 +13,13 @@
  *
  * e.g.: `Uploads/a1312bc34d/sam__ResizedImageWzYwLDgwXQ.jpg`
  */
-class HashFileIDHelper implements FileIDHelper
+class HashFileIDHelper extends AbstractFileIDHelper
```
ChloeSartorelli: This is BC safe - anything checking for types of `FileIDHelper` will still pass because `AbstractFileIDHelper` implements it.

This is similar to when we added an `AbstractGridfieldComponent` class.

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -14,70 +13,13 @@
...
 {
-    use Injectable;
```
ChloeSartorelli: Moved `use Injectable` into the abstract class

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -14,70 +13,13 @@
...
-
     /**
      * Default length at which hashes are truncated.
      */
     const HASH_TRUNCATE_LENGTH = 10;
 
-    public function buildFileID($filename, $hash = null, $variant = null, $cleanfilename = true)
```
ChloeSartorelli: Moved this method to the abstract class. There were only a very few small places where this implementation differed from `NaturalFileIDHelper`, and those are handled by calling new `protected` methods in the abstract class's implementation of this method.

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -14,70 +13,13 @@
...
-
-        // Add extension
-        if ($extension) {
-            $fileID .= $extension;
-        }
-
-        return $fileID;
-    }
-
-    public function cleanFilename($filename)
```
ChloeSartorelli: Moved this method to the abstract class as it's identical to the implementation in `NaturalFileIDHelper`

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -124,9 +84,4 @@ private function truncate($hash)
     {
         return substr($hash ?? '', 0, self::HASH_TRUNCATE_LENGTH);
     }
-
-    public function lookForVariantRecursive(): bool
```
ChloeSartorelli: Moved this method to the abstract class as it's identical to the implementation in `NaturalFileIDHelper`

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public function lookForVariantRecursive(): bool
+    {
+        return false;
+    }
+
+    abstract protected function getFileIDBase($shortFilename, $fullFilename, $hash, $variant): string;
+
+    abstract protected function validateFileParts($filename, $hash, $variant): void;
```
ChloeSartorelli: These accept more arguments than the current concrete classes use to make it easier to have additional subclasses with their own validation or fileIDBase implementations.

src/ImageBackendFactory.php
```diff
@@ -37,9 +37,15 @@ public function __construct(Factory $creator)
...
     {
-        /** @var AssetContainer $assetContainer */
+        /** @var AssetContainer|null $assetContainer */
         $assetContainer = reset($params);
-        if (!$assetContainer instanceof AssetContainer) {
+
+        // If no asset container was passed in, create a new uncached image backend
+        if (!$assetContainer) {
+            return $this->creator->create($service, $params);
+        }
```
ChloeSartorelli, line 46: Needed so that developers can get an instance of the injected `Image_Backend` when they don't have a valid `AssetContainer` instance that contains an image file. The example in the "manual testing steps" section of the PR description simply won't work without this.

src/ImageBackendFactory.php
```diff
@@ -37,9 +37,15 @@ public function __construct(Factory $creator)
...
+
+        if (!($assetContainer instanceof AssetContainer)) {
```
ChloeSartorelli, line 48: This used to be `if (!$assetContainer instanceof AssetContainer)` - which actually checks `if ((!$assetContainer) instanceof AssetContainer)`. In other words it was checking if a boolean was an instance of a class, which will always evaluate to false.

src/ImageManipulation.php
```diff
@@ -840,6 +841,18 @@ public function isHeight($height)
         return $this->getHeight() === $height;
     }
 
+    /**
+     * Wrapper for manipulate() that creates a variant file with a different extension than the original file.
+     *
+     * @return DBFile The manipulated file
+     */
+    public function manipulateExtension(string $newExtension, callable $callback)
```
ChloeSartorelli, line 849: Note we can't add an explicit return typehint here because we're just using `return $this->manipulate(/**/);` which _should_ return `DBFile` but _could_ be overridden to return any arbitrary thing.

src/InterventionBackend.php
```diff
@@ -366,8 +366,10 @@ public function writeToStore(AssetStore $assetStore, $filename, $hash = null, $v
                 throw new BadMethodCallException("Cannot write corrupt file to store");
             }
 
+            // Make sure we're using the extension of the variant file, which can differ from the original file
+            $url = $assetStore->getAsURL($filename, $hash, $variant, false);
+            $extension = pathinfo($url, PATHINFO_EXTENSION);
             // Save file
-            $extension = pathinfo($filename ?? '', PATHINFO_EXTENSION);
```
ChloeSartorelli, line 370: This change is vital if using `InterventionImage::writeToStore()` to store (and encode) the variant. Without this, it will think it's meant to store it using the original file extension which will be wrong at best and outright fail if the original file wasn't an image.

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Get the original file's filename with the extension rewritten to be the same as either the original
+     * or the variant extension.
+     *
+     * @param string $filename Original filename without variant
+     * @param int $extIndex One of self::EXT_ORIGINAL or self::EXT_VARIANT
+     */
+    private function swapExtension(string $filename, string $variant, int $extIndex): string
```
ChloeSartorelli: Swapping the extension is necessary so we can correctly identify the file on the filesystem. Without this, the original file extension is used when searching for the variant file in the filesystem.

e.g. you have an original file `myfile.mp4`, and a variant `myfile__extRewriteWyJtcDQiLCJqcGciXQ.jpg`.
Without this method, if would look for `myfile__extRewriteWyJtcDQiLCJqcGciXQ.mp4` on the filesystem, and obviously it wouldn't find it.

tests/php/ImageTest.php
ChloeSartorelli, line 1: See the description of https://github.com/silverstripe/silverstripe-assets/pull/585/commits/c8f0c29748eff1ac8711e9ed52ab9fd9b0cce8e6 for changes in this file

src/ImageManipulation.php
```diff
@@ -1024,7 +1039,7 @@ public function variantParts($variantName)
             throw new InvalidArgumentException('Invalid variant name arguments: ' . $variantName);
         }
 
-        return array_merge([$matches['format']], $args[0]);
+        return array_merge([$matches['format']], $args);
```
ChloeSartorelli, line 1040: See the description of https://github.com/silverstripe/silverstripe-assets/pull/585/commits/c8f0c29748eff1ac8711e9ed52ab9fd9b0cce8e6

emteknetnz (changes requested):
I'm finding this too hard too peer review despite it being split into different commits.  I'm jumping between the 'commits' tab and the 'files changed' tab where I want to put my comments and loosing track of things.

Please split this PR into two different PRs, one for refactoring and one for the original issue. In the future please refrain from doing large amount of refactoring in PRs that also contain new functionality / fix a bug .

GuySartorelli:
Moved the refactoring into https://github.com/silverstripe/silverstripe-assets/pull/587

Marking this as draft because it'll have conflicts once that is merged.

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -2,8 +2,6 @@
 
 namespace SilverStripe\Assets\FilenameParsing;
 
-use SilverStripe\Core\Injector\Injectable;
```
ChloeSartorelli, line 5: Missed this in the refactor PR - it's not used anymore

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -53,6 +73,66 @@ public function buildFileID($filename, $hash = null, $variant = null, $cleanfile
...
+        // If there's no variant at all, we can rewrite the filenmane
+        if (empty($variant)) {
+            return $filename;
+        }
+
+        // Split variant string in variant list
+        $subVariants = explode('_', $variant);
+
+        // Split our filename into a filename and extension part
+        if (!preg_match('/(?<basename>.+)\.(?<ext>[a-z\d]+)$/i', $filename, $matches)) {
```
emteknetnz: Use pathinfo() instead of your own regex, there may be some weird filenames this doesn't work on

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -23,10 +21,16 @@ public function parseFileID($fileID)
         }
 
         $filename = $matches['folder'] . $matches['basename'] . $matches['extension'];
+        $variant = isset($matches['variant']) ? $matches['variant'] : '';
+
+        if (isset($variant)) {
```
emteknetnz:
```suggestion
        $variant = $matches['variant'] ?: '';

        if ($variant) {
```

src/ImageManipulation.php
emteknetnz: Not sure that's a valid concern?  Who in the real world would subclass ImageManipulation and change manipulate()?  And even if they did they can just change the return type on this method?  We don't normally care about project override core classes. Am I missing something here?

Seems like return type should be `: ?DBFile` since manipulate() can return `null`

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -53,6 +73,66 @@ public function buildFileID($filename, $hash = null, $variant = null, $cleanfile
...
+     */
+    private function swapExtension(string $filename, string $variant, int $extIndex): string
+    {
+        // If there's no variant at all, we can rewrite the filenmane
+        if (empty($variant)) {
+            return $filename;
+        }
+
+        // Split variant string in variant list
+        $subVariants = explode('_', $variant);
```
emteknetnz, line 92: What happens when the original filename has underscores in it? You'll need to add a unit test for this.

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -2,12 +2,28 @@
...
 use SilverStripe\Core\Injector\Injectable;
 
 abstract class AbstractFileIDHelper implements FileIDHelper
 {
     use Injectable;
 
+    /**
+     * A variant type for encoding a variant filename with a different extension than the original.
+     */
+    public const EXTENSION_REWRITE_VARIANT = 'extRewrite';
```
emteknetnz:
```suggestion
    public const EXTENSION_REWRITE_VARIANT = 'ExtRewrite';
```
All the other variants types have initial caps

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -53,6 +73,66 @@ public function buildFileID($filename, $hash = null, $variant = null, $cleanfile
...
+    }
+
+    /**
+     * Get the original file's filename with the extension rewritten to be the same as either the original
+     * or the variant extension.
+     *
+     * @param string $filename Original filename without variant
+     * @param int $extIndex One of self::EXT_ORIGINAL or self::EXT_VARIANT
+     */
+    private function swapExtension(string $filename, string $variant, int $extIndex): string
```
emteknetnz: The naming on these three methods seems really inconsistent, the words they start with are

"rewrite"
"restore"
"swap"

Yet they all do the same thing?

Just get rid of the first two methods just make this method public and also the EXT_* constants public

src/FilenameParsing/AbstractFileIDHelper.php
```diff
@@ -2,12 +2,28 @@
...
+
+    /**
+     * use the original file's extension
+     */
+    private const EXT_ORIGINAL = 0;
+
+    /**
+     * use the variant file's extension
+     */
+    private const EXT_VARIANT = 1;
```
emteknetnz:
```suggestion
    /**
     * Use the original file's extension
     */
    private const EXTENSION_ORIGINAL = 0;

    /**
     * Use the variant file's extension
     */
    private const EXTENSION_VARIANT = 1;
```
Make the naming consistent with other contant above and also make it slightly clearer. Also initial caps for "Use"

Will also need to update references to consts

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -30,10 +30,16 @@ public function parseFileID($fileID)
         }
 
         $filename = $matches['folder'] . $matches['basename'] . $matches['extension'];
+        $variant = isset($matches['variant']) ? $matches['variant'] : '';
+
+        if (isset($variant)) {
```
emteknetnz:
```suggestion
        $variant = $matches['variant'] ?: '';

        if ($variant) {
```

emteknetnz (changes requested):

src/FilenameParsing/AbstractFileIDHelper.php
ChloeSartorelli: Done

src/FilenameParsing/NaturalFileIDHelper.php
ChloeSartorelli: This isn't new code (it's just moved from below) but I'll validate and make a change here to avoid PR ping pong

ChloeSartorelli: Done

src/ImageManipulation.php
ChloeSartorelli: It is a valid concern, because we have a whole public API document that assumes people will change anything that's public. The whole point is that we should assume people will change anything that's public, and that way we never run into people saying "You broke this public API that I was using in a bizarre but totally semver-friendly way!"

> And even if they did they can just change the return type on this method?

They can't just change the return type of this method if the return type they want isn't covariant (or contravariant, whichever the right one is here) with the return type we specify.

> We don't normally care about project override core classes. Am I missing something here?

We have a whole [definition of public API](https://docs.silverstripe.org/en/5/project_governance/public_api/) which assumes people will override core classes.

> Seems like return type should be : ?DBFile since manipulate() can return null

Updated the PHPDoc for each of the manipulate methods to reflect this.

src/FilenameParsing/AbstractFileIDHelper.php
ChloeSartorelli: That's just a fluke because the method names start with a capital letter - but I'll change this to avoid PR ping pong

ChloeSartorelli: Done

src/FilenameParsing/AbstractFileIDHelper.php
ChloeSartorelli: Just lifted these straight from the POC as they were, since they did what they needed to do.

Done - but made protected since they're only used by subclasses (and used reflection in tests)

src/FilenameParsing/AbstractFileIDHelper.php
ChloeSartorelli: Done

src/FilenameParsing/HashFileIDHelper.php
ChloeSartorelli: https://github.com/silverstripe/silverstripe-assets/pull/585#discussion_r1466955600

Done

src/FilenameParsing/AbstractFileIDHelper.php
ChloeSartorelli: It won't match the variant regex and so it will be ignored.

ChloeSartorelli: I've added underscores to file names in the test scenarios.

src/ImageManipulation.php
emteknetnz: OK make sense now, it's purely because manipulate() doesn't a have a strong return type - which you'd already mentioned :-)

emteknetnz (changes requested):
Tested locally, works good

Won't block merging on this, though just a question really - is it expected to have intermediatory files stay on the file system, for instance I followed the DOC PR instructions for creating a webp image `$MyFile.format('webp').ScaleWidth(150)`, though I ended up with this in the filesystem

image

Ideally we wouldn't have the unused webp image persisted instead only have the webp+ScaleWidth image, is it viable to have the intermediatory image not persist? OK to make this a new card if it's viable though complex

GuySartorelli:
I'm not even sure if that's viable, to be honest. Each variant is a file. Here you have one variant for the file conversion, and another variant for scaling the image - and because those are discrete steps, they create discrete files.

I'm pretty sure the same thing would happen if you scaled and then rotated an image - you'd have one variant (and therefore one file) for the scaling operation and one for the rotation operation.

I don't think changing that is really feasible - or at least with my current limited understanding of the assets system I can't think of a way to tackle it.

emteknetnz (approved):

Issue #489, opened 2022-05-03
API/Service for registering "converters" that will create file variant under a different extensions


maxime-rainville:
As a CMS Developer, I want clear and simple direction when implementing logic that converts files between formats.

## Acceptance criteria
- [x] There's a high-level API for implementing file format converters that [wraps around the lower level API](https://github.com/silverstripe/silverstripe-assets/issues/488).
- [x] Converters can validate the presence of required libraries, binaries, PHP modules or dependencies.
- [x] The converter API makes it clear that missing dependencies should throw a suitable Exception.
- [x] If a user attempts to convert a file to a format that is not available, a suitable Exception value is thrown.
- [x] Multiple converters for the same format can be defined with the potential to use alternative converters if the first one can not handle a specific source format. (e.g. You could have a converter that can turn any image to JPEG. But it might not be able to take a PDF and create a JPEG thumbnail ... so you add an alternative converter.)
- [x] Developer doc for how to use the converter API is well defined.

## Excludes
- The creation of any specific converters ... this is just about the API.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/595
- https://github.com/silverstripe/developer-docs/pull/491

michalkleiner:
Do we want to consider a file variant under the same file extension/same type of file? I guess the question is what is a file variant in this piece of work. Would e.g. adding a watermark and saving as the same file type constitute a variant? Or should it be a different file name and that will not be seen as a variant?

GuySartorelli:
#488 mentions that calling `Jpeg()` on a file that is already a `.jpg` or `.jpeg` file shouldn't create a duplicate, but @michalkleiner has a good point that there might be a use case for using the converter API to generate, instead, a different variant of the jpeg.... I think however that the API as defined in #488 would be too vague to cleanly support that.

I think the converter API should not explicitly say "this is already a jpeg so I'm not touching it" - i.e. developers could use the converter API to define their own converter that adds a watermark to a jpg and outputs a jpg - but they should define their own method in an extension on the `Image` class (e.g. `Watermark()`) which would then engage the converter to generate that image.

I think it would make sense for the result of that to be a variant, rather than its own file - but maybe there could be a way for developers to choose whether the output of conversion is a variant or a new `File` db entry?

maxime-rainville:
[...]

PR #595, opened 2024-04-04
NEW File converter API

GuySartorelli:
## Description

Provides:
- An API for file converters, so the community can provide their own file converters and expect them to work consistently with anything we build ourselves
- A file converter that uses Intervention Image for converting from one image format to another
- A manager for file conversion so you can just try a conversion and if there's any converter which supports it, your conversion will be successful
- A method on `File` and `DBFile` to leverage available converters

Note that one AC is self-fulfilling:
> Converters can validate the presence of required libraries, binaries, PHP modules or dependencies.

Converters can validate whatever the heck they want. If they don't have the stuff they need, they should just return `false` from the method that checks if they support a given conversion.

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/489


src/Conversion/FileConversionManager.php
```diff
@@ -0,0 +1,35 @@
...
+
+use SilverStripe\Assets\Storage\DBFile;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injector;
+
+class FileConversionManager
+{
+    use Configurable;
+
+    private static $converters = [];
```
ChloeSartorelli: > [!NOTE]
> I haven't decided how to organise the converters yet so I'm seeking feedback on this specific implementation detail.

1. The easiest way to arrange this is just a regular array of converters - and you just dump all of the converters in here and loop through all of them until you either find one that supports the conversion or run out of converters.
This is fine when there are only a few converters, assuming the logic in `supportsConversion()` doesn't have to do any serious computation.
This will _not_ scale well if there ends up being a lot of converters.
1. An alternative would be to register based on the category of file that the converter either converts _from_ or _to_, e.g:
    ```yml
    SilverStripe\Assets\Conversion\FileConversionManager:
      converters:
        image:
          SilverStripe\Assets\Conversion\InterventionImageConverter
          App\Conversion\PdfToImageConverter
        image/supported:
          SilverStripe\Assets\Conversion\InterventionImageConverter
          App\Conversion\PdfToImageConverter
        document:
          App\Conversion\WordToPdfConverter
    ```
    - The above uses the category names from [`File.app_categories` configuration](https://github.com/silverstripe/silverstripe-assets/blob/0cee400bbecc908b2d75c1254ee37bf5963625f9/src/File.php#L216-L242) - but note that image converters need to be doubled up, because there's two separate image categories.
    - Note it's also a) not clear whether these are what the converter supports converting _to_ or _from_ (e.g. `PdfToImageConverter` is in the `image` and `image_supported` categories because it converts _to_ those - but developers could easily put it in the `document` category accidentally thinking they're meant to put them in the category they convert _from_ (or vice versa)
1. Similar to 2 above, except they could be listed everywhere they convert *from* and everywhere they convert *to*. This would ensure we only ever check converters that convert from-and-to the requested formats - but it's even more confusing to configure. I would definitely not recommend this unless we find people are making thousands of converters.
1. Similar to 3 above, but flip the config on its head like so:
    ```yml
    SilverStripe\Assets\Conversion\FileConversionManager:
      converters:
          SilverStripe\Assets\Conversion\InterventionImageConverter:
            from: ['image', 'image/supported']
            to: ['image', 'image/supported']
          App\Conversion\PdfToImageConverter:
            from: 'document'
            to: ['image', 'image/supported']
          App\Conversion\WordToPdfConverter
    ```
    - Any converter with no listed "from" or "to" will either throw an exception for bad config, or more likely just always check those, but check them after everything else.
    - Any converter which _does_ list a "from" and a "to" category can be skipped early if the format to convert from-or-to isn't in their list.
1. ??? I'm sure there are other ways we could organise this - and if someone has a great idea I'd be happy to hear it.

Right now I'm torn between option 1 (the simple approach, since it's unlikely there will ever be more than a handful of converters in a given project) and option 4 (which I think has a nice trade-off between being easy to understand and outright skipping converters we know won't handle conversions without even asking)

> [!NOTE]
> In the above suggestions I use categories instead of explicit file extensions because:
> - Sometimes the file extensions a converter supports will be dynamic based on what's available at runtime. We already see that with `InterventionImageConverter` in this PR.
> - Listing all of the formats in config is prone to error - especially simply errors like listing `jpg` but forgetting `jpeg`

maxime-rainville: I would lean toward option 1. I think the entire point of converters is to make this process as easy and transparent as possible for the end developer. 

If someone has very difficult and complex needs, that's the point where maybe you want to use the lower level API directly instead of using a converter.

Maybe we need a way to apply priorities so people can override converters or force their converter to take precedence over another, but I'm not sure.

ChloeSartorelli: > If someone has very difficult and complex needs, that's the point where maybe you want to use the lower level API directly instead of using a converter.

I think this misses the point a bit - it's just about avoiding asking converters if they can handle a conversion if the conversion is way out of its scope. e.g. don't ask an image converter if it can convert .doc to .pdf.

> Maybe we need a way to apply priorities so people can override converters or force their converter to take precedence over another.

Overriding converters should be done via the injector.
Precedence does make some sense though. Someone might have a hyper-efficient converter for some specific format (for example converting to webp) - but which doesn't handle anything else. It would make sense to want that to come before any other image converters.

I'll see what I can do in that regard, I think that makes a lot of sense.

ChloeSartorelli: Actually nevermind - priority isn't needed. People should just use the `After` and `Before` configuration directives.

ChloeSartorelli: For future reference: Since both Max and I leaned towards option 1 that's what I'll go with.

src/Conversion/FileConversionException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace SilverStripe\Assets\Conversion;
+
+use LogicException;
+
+/**
+ * An exception that represents a failure to convert a file in a FileConverter class.
+ */
+class FileConversionException extends LogicException
```
ChloeSartorelli: This allows us and developers to catch exceptions that were intentionally thrown for conversion failures - while still correctly getting errors in their faces if there are *unexpected* exceptions.

src/Conversion/FileConversionManager.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Convert the file to the given format using the first available converter that can perform the conversion.
+     *
+     * @param string $toFormat The file extension you want to convert to - e.g. "webp".
+     * @param array $options Any options defined for this converter which should apply to the conversion.
+     * Note that if a converter supports the conversion generally but doesn't support these options, that converter will not be used.
+     * @throws FileConversionException if the conversion failed or there were no converters available.
+     */
+    public function convert(DBFile $from, string $toFormat, array $options = []): DBFile
```
ChloeSartorelli: Providing `$options` here gives a higher level of control for developers (and converters) which might require it.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,199 @@
...
+                continue;
+            }
+            if (!is_int($value)) {
+                $problems[] = "quality value must be an integer";
+            }
+        }
+        return $problems;
+    }
+
+    private function supportedByIntervention(string $from, string $to, string $driver): bool
```
ChloeSartorelli: Intervention Image doesn't currently tell you what it supports. [It might in the future](https://github.com/Intervention/image/issues/1326) but for now we have to check for ourselves.

src/ImageManipulation.php
```diff
@@ -709,6 +711,33 @@ public function ThumbnailURL($width, $height)
...
+    /**
+     * Convert the file to another format if there's a registered converter that can handle it.
+     *
+     * @param string $toFormat The file extension you want to convert to - e.g. "webp".
+     * @param bool $returnNullOnFailure If true and the conversion fails, `null` is returned instead.
+     * This is useful if calling this method from a template, for example, where you don't want to be throwing
+     * exceptions for end users.
+     * @throws FileConversionException If the conversion fails and $returnNullOnFailure is false.
+     */
+    public function convert(string $toFormat, bool $returnNullOnFailure = false): ?AssetContainer
```
ChloeSartorelli: We intentionally don't expose `$options` here because this is intended to be primarily used from templates (e.g. `$MyFile.convert('webp')`). Passing an array of options in templates is I think unsupported but at best would be complicated to do. If developers really need that level of control of their file conversions, they can use the `FileConversionManager` in their controller or model.

ChloeSartorelli: I originally had `$returnOriginalOnFailure` instead of `$returnNullOnFailure`, and it would return `$this` if the conversion failed. But `null` is more consistent with how other methods in this file behave, and in many cases it would be more desirable to display *nothing* in the template instead of *the wrong thing*.

ChloeSartorelli: We could optionally remove that parameter and just **always** return `null` on failure. I'll leave it as is for now pending review.

The main benefit of removing the parameter is that it is more consistent with other methods here. We'd probably want to at least log the failure though if we did that.
The main benefit of keeping it is that silent failing is an explicit decision developers have to make, which makes it safer.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+                    return [$tuple, $backend];
+                }
+            );
+        } catch (ImageException $e) {
+            throw new FileConversionException('Failed to convert: ' . $e->getMessage(), $e->getCode(), $e);
+        }
+        // This is very unlikely but the API for `manipulateExtension()` allows for it
+        if ($result === null) {
+            throw new FileConversionException('File conversion resulted in null. Check whether original file actually exists.');
+        }
```
ChloeSartorelli: I can't find a way to test either of these - I think all of the scenarios that would normally result in these are being explicitly validated against already. This is just here as some defensive programming.

src/ImageManipulation.php
ChloeSartorelli: Note that this is in the `ImageManipulation` trait because this trait is shared in both the `File` and `DBFile` class. The issue originally had some comments about this that I wanted to have as part of the ACs but it was decided that they [were not relevant](https://github.com/silverstripe/silverstripe-assets/issues/489#issuecomment-1945698768) - so we're stuck with putting a file-thing into an image trait.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+        if (!is_a($backend, InterventionBackend::class)) {
+            return false;
+        }
+        return $this->supportedByIntervention($fromFormat, $backend) && $this->supportedByIntervention($toFormat, $backend);
+    }
+
+    public function convert(DBFile $from, string $toFormat, array $options = []): DBFile
+    {
+        // Do some basic validation up front for things we know aren't supported
+        $problems = $this->validateOptions($options);
```
michalkleiner, line 36: Above we call the variable `unsupportedOptions` which sounds a bit better, however it's not clear that a method called `validateSomething` would return a list of exclusions/unsupported items only. Could we consider something like `filterInvalidOptions` or return an array of valid and invalid options etc.

ChloeSartorelli: `filterInvalidOptions` suggests it will filter out anything that's invalid and return the valid options (if there were any) as a result. I have no problem renaming the method or the variable but I think that name suggests different functionality.

The idea here is that there's a clear exception if any unsupported, invalid, etc options are passed in so developers can set their options correctly. No unsupported/invalid options will ever be passed in here unless the developer explicitly chooses to use this converter directly - the manager would have checked `supportsValidation()` first, which would have returned `false`.

> however it's not clear that a method called validateSomething would return a list of exclusions/unsupported items only.

It's returning an array of problems (or exception messages if you prefer?) so we can tell the developer exactly what they did wrong. I'm not sure what method name could convey that. `checkIfOptionsAreValidAndReturnClearMessagesAboutThingsTheDeveloperNeedsToFix()` is a bit too verbose, for example.

It's a private method anyway so it's not as important as a public method would be, and IMO it's pretty clear from its usage what it's doing - I can add a PHPdoc if it really needs clarifying, or if somsone can think of a really good name I'm happy to change it.

src/ImageManipulation.php
michalkleiner: If we need to pass in options from templates in projects we pass them as json_encoded array or object, and access it using `func_get_args` as it can't have a default value.

Which brings me to the second param here which may not be able to be overridden from a template and would always be false when called from a template.

For consistency with other methods accessible from templates, would `Convert` (with capital C` be better?

ChloeSartorelli: > If we need to pass in options from templates in projects we pass them as json_encoded array or object, and access it using func_get_args as it can't have a default value.

Sorry, I'm confused about what you mean here... this method doesn't exist outside of this PR... Where are you passing options to in templates?

> Which brings me to the second param here which may not be able to be overridden from a template and would always be false when called from a template.

You can call `$MyFile.convert('ext', true)` in a template. I have tested this and it works as expected.

> For consistency with other methods accessible from templates, would Convert (with capital C` be better?

The other methods do not follow the current best practices in our [PHP Coding Conventions](https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/) - but we should adhere to best practices with all new code, I think.

src/Conversion/InterventionImageFileConverter.php
michalkleiner: Fair enough re: all the points, thanks for taking the time to explain that!

src/ImageManipulation.php
michalkleiner: > > If we need to pass in options from templates in projects we pass them as json_encoded array or object, and access it using func_get_args as it can't have a default value.
> 
> Sorry, I'm confused about what you mean here... this method doesn't exist outside of this PR... Where are you passing options to in templates?

That was just an example how one can access extra parameters passed in, sorry for the confusion.

> > Which brings me to the second param here which may not be able to be overridden from a template and would always be false when called from a template.
> 
> You can call `$MyFile.convert('ext', true)` in a template. I have tested this and it works as expected.

Would it also work if you did `<% with $MyFile %><a href="$convert('ext', true)"><% end_with %>`?
 
> > For consistency with other methods accessible from templates, would Convert (with capital C` be better?
> 
> The other methods do not follow the current best practices in our [PHP Coding Conventions](https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/) - but we should adhere to best practices with all new code, I think.

That was my guess right after I asked.

ChloeSartorelli: > That was just an example how one can access extra parameters passed in, sorry for the confusion.

Ahh okay, fair enough.
Honestly I think the times people will want explicit options will be quite uncommon (any given option will usually only be supported by one specific converter anyway) - and in those cases there's always the option of handling this in their model or controller.

I personally would prefer to err on the side of a simpler API at this stage, and revisit if we find this is becoming a barrier for people - but I'm happy to change it if the consensus is that we should allow JSON formatted strings from templates here.

> Would it also work if you did `<% with $MyFile %><a href="$convert('ext', true)"><% end_with %>`?

Don't see why not, the syntax is still the same. If that doesn't work it's a bug in the parser which needs to be reported and fixed.

michalkleiner: I'll try to reproduce the default param not being able to be overridden from a template on a fresh install. Might be just related to global template providers, not to methods on objects like here.

michalkleiner:
✅ I haven't done any functional testing so won't do a full PR approval but in general this looks good to me.

src/ImageManipulation.php
ChloeSartorelli: Could also be related to getters - `$MyGetter('some param')` won't actually pass the values through to a `getMyGetter()` method, you have to use the full method name in the template like `$getMyGetter('some param')`

emteknetnz: I think remove the param and just return null

You have written this "This is useful if calling this method from a template, for example, where you don't want to be throwing for end users."

That sounds like the main scenario this method will be called in? In which case let's not force the developer to have to put in a param, which they don't have to add in for any other template methods.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+                if ($driver === 'gd') {
+                    return false;
+                }
+                // always supported by imagick
+                return true;
+            case 'webp':
+                if ($driver === 'gd' && !function_exists('imagewebp')) {
+                    return false;
+                }
+                if ($driver === 'imagick' && !\Imagick::queryFormats('WEBP')) {
```
emteknetnz:
```suggestion
                if ($driver === 'imagick' && !Imagick::queryFormats('WEBP')) {
```

Import Imagick up the top.  Remove the trailing slash for other examples in this file

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+        if ($format === '') {
+            return false;
+        }
+
+        // GD and Imagick support different things.
+        // This follows the logic in intervention's AbstractEncoder::process() method
+        // and the various methods in the Encoder classes for GD and Imagick,
+        // excluding checking for strings that were obviously mimetypes
+        switch (strtolower($format)) {
+            case 'gif':
```
emteknetnz, line 108: Backlog an issue linking to https://github.com/Intervention/image/issues/1326 to refactor this part

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+        if (!$this->supportedByIntervention($toFormat, $originalBackend)) {
+            throw new FileConversionException("Convertion to format '$toFormat' is not suported.");
+        }
+
+        $quality = $options['quality'] ?? null;
+        // Clone the backend if we're changing quality to avoid affecting other manipulations to that original image
+        $backend = $quality === null ? $originalBackend : clone $originalBackend;
+        // Pass through to invervention image to do the conversion for us.
+        try {
+            $result = $from->manipulateExtension(
```
emteknetnz, line 54:
```suggestion
            $to = $from->manipulateExtension(
```

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+                    $config = ['conflict' => AssetStore::CONFLICT_USE_EXISTING];
+                    $tuple = $backend->writeToStore($store, $filename, $hash, $variant, $config);
+                    return [$tuple, $backend];
+                }
+            );
+        } catch (ImageException $e) {
+            throw new FileConversionException('Failed to convert: ' . $e->getMessage(), $e->getCode(), $e);
+        }
+        // This is very unlikely but the API for `manipulateExtension()` allows for it
+        if ($result === null) {
```
emteknetnz, line 69:
```suggestion
        if ($to === null) {
```

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -0,0 +1,181 @@
...
+            throw new FileConversionException('File conversion resulted in null. Check whether original file actually exists.');
+        }
+        return $result;
```
emteknetnz, line 72:
```suggestion
        return $to;
```

emteknetnz (changes requested):
As usual my main thing is naming :-)

- Change FileConversionManager to FileConverterManager
- Change FileConversionException to FileConverterException
- Change $fromFormat to $fromExtension
- Change $toFormat to $toExtension
- convert() should be Convert() to match convention when used from templates

Also make corresponding naming changes in https://github.com/silverstripe/developer-docs/pull/491

Also, I had this in my template: `$MyFile.convert('webp')`  - however the html rendered had the original png file as the alt tag. I would have expected a webp file in the alt tag.

src/ImageManipulation.php
ChloeSartorelli: I don't like that personally but I don't feel strongly enough about it to have peer review ping pong about it.

> That sounds like the main scenario this method will be called in?

There are all PHP methods and can be called from PHP. Any time you're dealing with an image (or file in this case) in PHP for any reason, you'll call these methods. File conversion will I suspect be just as common in PHP as it will in templates, for things like converting files before attaching to emails, etc.

That said, I've made the requested change and logged the error as a compromise.

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: Done

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: I'm not going to raise an issue to make a change based on the possibility that intervention image **might** have this functionality in a future release.

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: This is very subjective. To me `$result` makes a lot more sense, because it is the result of the conversion. `$to` as a variable name to me implies that the conversion has not yet happened, and this variable stores information about what the conversion will be made *to*.

With that context, if you still insist on me making the change, I'll make it at that point purely to avoid additional ping pong - but I am resisting for now because 
1. it's a subjective change that we have different views on - there is no objectively better name here
2. I prefer the name I already am using for the reasons discussed above

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: Any discussion about this can be in https://github.com/silverstripe/silverstripe-assets/pull/595#discussion_r1558795888

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: Any discussion about this can be in https://github.com/silverstripe/silverstripe-assets/pull/595#discussion_r1558795888

GuySartorelli:
> Change FileConversionManager to FileConverterManager
> Change FileConversionException to FileConverterException
> Change $fromFormat to $fromExtension
> Change $toFormat to $toExtension

Done, though I think the first two are not as good as the original names. Naming is always subjective though so 🤷‍♀️

> convert() should be Convert() to match convention when used from templates

This was suggested in https://github.com/silverstripe/silverstripe-assets/pull/595#discussion_r1556894605 and I provided reasoning why I didn't do that in that comment thread. Please read that - and if you still want this change after you have that context, I'll make the change at that stage to avoid additional ping pong.


> Also, I had this in my template: $MyFile.convert('webp') - however the html rendered had the original png file as the alt tag. I would have expected a webp file in the alt tag.

This is unrelated to this API - that's a bug with the low-level API and will need to be logged as a separate issue.

emteknetnz:
> convert() should be Convert() to match convention when used from templates

Yes it's not following our own best practice for PHP methods ... however when we wrote those we probably didn't consider methods intended to called from template.  The convention there very much is initial caps for method names unfortunately and we should follow the existing convention. ImageManipulation is full of public methods named like that.  End users will be expecting to use `$MyFile.Convert('webp')` rather than `$MyFile.convert('webp')` because they're already calling `$MyFile.ThumbnailURL(300, 200)`

GuySartorelli:
Methods are just methods. Some methods _can_ be called from templates. Especially in this case, it is not exclusively intended to be used in templates.

Nevertheless, I will rename this method to avoid additional unnecessary ping pong - but in the future I will be more resistant to this sort of change because it does go against our best practices (which do not and *should* not make exceptions for methods which can be called from templates).

Done.

emteknetnz (approved):

PR #596, opened 2024-04-29
Align rendering of image genercated by ImageShortCodeProvider with generic Image generation

maxime-rainville:
This PR aligns how ImageShortcodeProvider generates its HTML markup to be aligned with how regular images are generated.

This reduces complexity and makes it easy to customise the HTML markups.

## Parent issue
- https://github.com/silverstripe/.github/issues/230

## Depends on
- https://github.com/silverstripe/silverstripe-framework/pull/11217
  - [Review combined build](https://github.com/creative-commoners/recipe-kitchen-sink/actions/workflows/ci.yml?query=branch%3Apulls%2F5%2Falt-attribute-html)

### DIFF
--- a/src/ImageManipulation.php
+++ b/src/ImageManipulation.php
@@ -1145,8 +1145,8 @@ trait ImageManipulation
         $attributes = [];
         if ($this->getIsImage()) {
             $attributes = [
-                'width' => $this->getWidth(),
-                'height' => $this->getHeight(),
+                'width' => $this->getWidth() ?: null,
+                'height' => $this->getHeight() ?: null,
                 'alt' => $this->getTitle(),
                 'src' => $this->getURL(false)
             ];
@@ -1171,11 +1171,19 @@ trait ImageManipulation
 
         $attributes = array_merge($defaultAttributes, $this->attributes);
 
-        // We need to suppress the `loading="eager"` attribute after we merge the default attributes
-        if (isset($attributes['loading']) && $attributes['loading'] === 'eager') {
-            unset($attributes['loading']);
-        }
+        if (isset($attributes['loading'])) {
+            // Check if the dimensions for the image have been set
+            $dimensionsUnset = (
+                empty($attributes['width']) ||
+                empty($attributes['height'])
+            );
 
+            // We need to suppress the `loading="eager"` attribute after we merge the default attributes
+            // or if dimensions are not set
+            if ($attributes['loading'] === 'eager' || $dimensionsUnset) {
+                unset($attributes['loading']);
+            }
+        }
         $this->extend('updateAttributes', $attributes);
 
         return $attributes;
--- a/src/Shortcodes/ImageShortcodeProvider.php
+++ b/src/Shortcodes/ImageShortcodeProvider.php
@@ -7,6 +7,7 @@ use SilverStripe\Assets\File;
 use SilverStripe\Assets\Image;
 use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\View\Parsers\ShortcodeHandler;
 use SilverStripe\View\Parsers\ShortcodeParser;
 
@@ -73,26 +74,43 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
             return null; // There were no suitable matches at all.
         }
 
+        // Grant access to file if necessary
+        if (static::getGrant($record)) {
+            $record->grantFile();
+        }
+
         // Check if a resize is required
+        $manipulatedRecord = $record;
         $width = null;
         $height = null;
-        $grant = static::getGrant($record);
-        $src = $record->getURL($grant);
         if ($record instanceof Image) {
             $width = isset($args['width']) ? (int) $args['width'] : null;
             $height = isset($args['height']) ? (int) $args['height'] : null;
+
+            // Resize the image if custom dimensions are provided
             $hasCustomDimensions = ($width && $height);
             if ($hasCustomDimensions && (($width != $record->getWidth()) || ($height != $record->getHeight()))) {
-                $resized = $record->ResizedImage($width, $height);
+                $resized = $manipulatedRecord->ResizedImage($width, $height);
                 // Make sure that the resized image actually returns an image
                 if ($resized) {
-                    $src = $resized->getURL($grant);
+                    $manipulatedRecord = $resized;
                 }
+            }
+
+            // If only one of width or height is provided, explicitly unset the other
+            if ($width && !$height) {
+                $args['height'] = false;
+            } elseif (!$width && $height) {
+                $args['width'] = false;
             }
         }
 
-        // Determine whether loading="lazy" is set
-        $args = ImageShortcodeProvider::updateLoadingValue($args, $width, $height);
+        // Set lazy loading attribute
+        if (!empty($args['loading'])) {
+            $loading = strtolower($args['loading']);
+            unset($args['loading']);
+            $manipulatedRecord = $manipulatedRecord->LazyLoad($loading !== 'eager');
+        }
 
         // Build the HTML tag
         $attrs = array_merge(
@@ -101,7 +119,7 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
             // Use all other shortcode arguments
             $args,
             // But enforce some values
-            ['id' => '', 'src' => $src]
+            ['id' => '', 'src' => '']
         );
 
         // If file was not found then use the Title value from static::find_error_record() for the alt attr
@@ -111,11 +129,14 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
 
         // Clean out any empty attributes (aside from alt) and anything not whitelisted
         $whitelist = static::config()->get('attribute_whitelist');
-        $attrs = array_filter($attrs ?? [], function ($v, $k) use ($whitelist) {
-            return in_array($k, $whitelist) && (strlen(trim($v ?? '')) || $k === 'alt');
-        }, ARRAY_FILTER_USE_BOTH);
+        foreach ($attrs as $key => $value) {
+            if (in_array($key, $whitelist) && (strlen(trim($value ?? '')) || in_array($key, ['alt', 'width', 'height']))) {
+                $manipulatedRecord = $manipulatedRecord->setAttribute($key, html_entity_decode($value));
+            }
+        }
 
-        $markup = ImageShortcodeProvider::createImageTag($attrs);
+        // We're calling renderWith() with an explicit template in case someone wants to use a custom template
+        $markup = $manipulatedRecord->renderWith(self::class . '_Image');
 
         // cache it for future reference
         if ($fileFound) {
@@ -131,9 +152,12 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
 
     /**
      * Construct and return HTML image tag.
+     *
+     * @deprecated 2.3.0
      */
     public static function createImageTag(array $attributes) : string
     {
+        Deprecation::notice('2.3.0', 'Will be removed without equivalent functionality to replace it.');
         $preparedAttributes = '';
         foreach ($attributes as $attributeKey => $attributeValue) {
             if (strlen($attributeValue ?? '') > 0 || $attributeKey === 'alt') {
@@ -208,30 +232,5 @@ class ImageShortcodeProvider extends FileShortcodeProvider implements ShortcodeH
         return Image::create([
             'Title' => _t(__CLASS__ . '.IMAGENOTFOUND', 'Image not found'),
         ]);
-    }
-
-    /**
-     * Updated the loading attribute which is used to either lazy-load or eager-load images
-     * Eager-load is the default browser behaviour so when eager loading is specified, the
-     * loading attribute is omitted
-     *
-     * @param array $args
-     * @param int|null $width
-     * @param int|null $height
-     * @return array
-     */
-    private static function updateLoadingValue(array $args, ?int $width, ?int $height): array
-    {
-        if (!Image::getLazyLoadingEnabled()) {
-            return $args;
-        }
-        if (isset($args['loading']) && $args['loading'] == 'eager') {
-            // per image override - unset the loading attribute unset to eager load (default browser behaviour)
-            unset($args['loading']);
-        } elseif ($width && $height) {
-            // width and height must be present to prevent content shifting
-            $args['loading'] = 'lazy';
-        }
-        return $args;
     }
 }
--- a/tests/php/Shortcodes/ImageShortcodeProviderTest.php
+++ b/tests/php/Shortcodes/ImageShortcodeProviderTest.php
@@ -15,6 +15,7 @@ use SilverStripe\Assets\Shortcodes\FileShortcodeProvider;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Security\InheritedPermissions;
 use SilverStripe\Security\Member;
+use DOMDocument;
 
 class ImageShortcodeProviderTest extends SapphireTest
 {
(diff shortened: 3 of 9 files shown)


src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -75,10 +75,10 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
         }
 
         // Check if a resize is required
+        $manipulatedRecord = $record;
```
maxime-rainville, line 83: The old approach was basically building the `` tag by hand.

I'm using standard Image manipulation to assign the correct size and attribute to a manipulated image instead.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -87,7 +87,7 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
                 $resized = $record->ResizedImage($width, $height);
                 // Make sure that the resized image actually returns an image
                 if ($resized) {
-                    $src = $resized->getURL($grant);
```
maxime-rainville, line 89: The grant isn't needed here. Session grants apply to the original file and all its variant.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -102,7 +102,7 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
             // Use all other shortcode arguments
             $args,
             // But enforce some values
-            ['id' => '', 'src' => $src]
+            ['id' => '', 'src' => '']
```
maxime-rainville, line 122: The `src` attirbutes need to be unset here because we want the resize picture to pick up its own URL.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -112,11 +112,14 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
...
-        }, ARRAY_FILTER_USE_BOTH);
+        foreach ($attrs as $key => $value) {
+            if (in_array($key, $whitelist) && (strlen(trim($value ?? '')) || $key === 'alt')) {
+                $manipulatedRecord = $manipulatedRecord->setAttribute($key, $value);
+            }
+        }
 
-        $markup = self::createImageTag($attrs);
+        // We're calling renderWith() with an explicit template in case someone wants to use a custom template
+        $markup = $manipulatedRecord->renderWith(self::class . '_Image');
```
maxime-rainville, line 139: We're using a `ImageShortcodeProvider` specific template here. This would allow you to customise each template individually.

maxime-rainville (commented):
## Template tree

`templates/DBFile_image.ss` is what was used to render the mark-up for images in templates. There's a related `templates/DBFile_download.ss` that renders plain file like.

I concluded that those files were misplaced. They should be following the namespace structure.

## What I did
- Created matching `DBFile` and `DBFile_Image` templates in the correct folder.
- ImageShortcodeProvider now generates its markup by calling its own template that includes the namespaced image template.
- I moved all the mark up from the original templates and moved it to the template in the correct namespace.
  - If you've customised image rendering by overriding `templates/DBFile_image.ss`, your custom template will still be used.
  - Because the ImageShortcodeProvider doesn't call `templates/DBFile_image.ss` at all, your custom template won't start bleeding into TinyMCE images.

## Things to change in CMS 6
Once this PR is merged, I'll do a follow up PR for CMS 6:
- This [method in DBFile](https://github.com/silverstripe/silverstripe-assets/blob/ff5e8a743eea9cdddb900579a6e8c3d77a48e8ba/src/Storage/DBFile.php#L157-L171) should reference the name spaced template.
- `templates/DBFile_image.ss` and `templates/DBFile_download.ss` will be deleted.
- `ImageShortcodeProvider::createImageTag()` will be deleted since it's now deprecated.

## Other
I haven't added any test because the output of this thing is meant to be identical to the original.

I guess the only test I could run is try to override some templates and validate that I still get the expected output. Not sure how much value that would provide.

src/Shortcodes/ImageShortcodeProvider.php
ChloeSartorelli: We still need to call `createImageTag()` because people may have subclassed the shortcode provider, replaced the original one with theirs, and updated the `createImageTag()` method. If we stop calling it, we throw away their customisation which is a breaking change.

Instead, update the implementation of `createImageTag()` to return markup which is generated via the template. Alternatively update it to return an empty string. Then we can call it and if it returns a truthy value we know it's a custom implementation and we should use that instead of the template.

src/ImageManipulation.php
```diff
@@ -1150,8 +1150,8 @@ protected function getDefaultAttributes(): array
         $attributes = [];
         if ($this->getIsImage()) {
             $attributes = [
-                'width' => $this->getWidth(),
-                'height' => $this->getHeight(),
+                'width' => $this->getWidth() ?: null,
```
maxime-rainville, line 1148: When the short code was rendering an non-existent image, `getWidth` and `getHeight` would return 0. This would lead to a `width="0"` attribute in the HTML.

Defaulting to null here fixes this.

src/ImageManipulation.php
```diff
@@ -1176,11 +1176,25 @@ public function getAttributes()
 
         $attributes = array_merge($defaultAttributes, $this->attributes);
 
-        // We need to suppress the `loading="eager"` attribute after we merge the default attributes
-        if (isset($attributes['loading']) && $attributes['loading'] === 'eager') {
-            unset($attributes['loading']);
+        if (isset($attributes['loading'])) {
+            // Check if the dimensions for the image have been set
+            $dimensionsUnset = (
```
maxime-rainville, line 1176: The shortcode had some logic to make sure that if no size was provided, the image would be eager loaded. This logic made more sense here in Image manipulation than in the original one.

--

The shortcode also had some logic to handle the scenario where only the height or only the width was provided. In that scenario, the missing attribute would be omitted.

Unfortunately, ImageManipulation really wants to read the size from the actual Image. This would lead to the aspect ratio of the image being changed. So I updated ImageShortcodeProvider to explicitely set the missing value to `auto` in those situation.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -75,25 +75,40 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
         }
 
         // Check if a resize is required
+        $manipulatedRecord = $record;
         $width = null;
         $height = null;
         $grant = static::getGrant($record);
-        $src = $record->getURL($grant);
```
maxime-rainville, line 80: TODO I don't need the `$src` here, but I do need to grant access to the file.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -75,25 +75,40 @@ public static function handle_shortcode($args, $content, $parser, $shortcode, $e
...
+                $args['height'] = 'auto';
+            }
+
+            if (!$width && $height) {
+                $args['width'] = 'auto';
+            }
         }
 
-        // Determine whether loading="lazy" is set
-        $args = self::updateLoadingValue($args, $width, $height);
```
maxime-rainville: ImageManipulation will take care of this from now on.

maxime-rainville: TODO Delete `updateLoadingValue` down below

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -46,33 +47,31 @@ protected function tearDown(): void
...
         $image = $this->objFromFixture(Image::class, 'imageWithTitle');
-        $parser = new ShortcodeParser();
-        $parser->register('image', [ImageShortcodeProvider::class, 'handle_shortcode']);
-
-        $this->assertEquals(
-            sprintf(
-                '<img src="%s" alt="">',
-                $image->Link()
-            ),
-            $parser->parse(sprintf('[image id=%d]', $image->ID))
```
maxime-rainville, line 57: Most of these test were throwing errors because `ImageManipulation` was returning attributes in a different order.

`ImageManipulation` does a few things a bit differently which also led to slightly different result in some context.

Because none of these test actually included messages, it was really hard to figure out what each test was meant to be doing. 

So I basically ended rewriting all the test with more descriptive message and to uso a utility assert method.

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -101,65 +100,79 @@ public function testShorcodeRegenrator()
...
 
     public function testShortcodeHandlerDoesNotResampleToNonIntegerImagesSizes()
     {
         $image = $this->objFromFixture(Image::class, 'imageWithoutTitle');
-        $parser = new ShortcodeParser();
-        $parser->register('image', [ImageShortcodeProvider::class, 'handle_shortcode']);
-
-        $this->assertEquals(
-            sprintf(
-                '<img src="%s" alt="" width="50%%" height="auto">',
```
maxime-rainville, line 138: This confused me for a bit. The way to escape a `%` in `sprintf` is to double it.

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -380,4 +482,39 @@ public function testCreateImageTag(string $expected, array $attributes)
...
         $this->assertEquals($expected, ImageShortcodeProvider::createImageTag($attributes));
     }
+
+    /**
+     * This method will assert that the $tag will contain an image with specific attributes and values
+     *
+     * @param array $attrs Key pair values of attributes and values.
+     *                     If the value is FALSE we assume that it is not present
+     */
+    private function assertImageTag(array $attrs, string $shortcode, $message = ""): void
```
maxime-rainville, line 516: Instead of testing whether the HTML generated by a specific shortcode exactly matches the expected HTML, this method will parse the HTML and check the individual attributes.

This is better because we shouldn't care about the order of the attributes. What we care about is whether the attribute is there or not and whether it has the correct value.

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -380,4 +482,39 @@ public function testCreateImageTag(string $expected, array $attributes)
...
+        $doc = new DOMDocument();
+        $doc->loadHTML($tag);
+        $node = $doc->getElementsByTagName('img')->item(0);
+        $nodeAttrs = $node->attributes;
+
+        foreach ($attrs as $key => $value) {
+            $nodeAttr = $nodeAttrs->getNamedItem($key);
+            if ($value === false) {
+                if ($nodeAttr !== null) {
+                    $this->fail("$message\nImage should not contain attribute '$key'\n$tag");
```
maxime-rainville, line 531: Instead of doing straight assertion here, I decided to directly fail the test. This has a few advantages:
- When I was asserting something about the node it would give a very unpleasant about what was failing which made it difficult to figure out what was actually wrong.
- I'm building much more specific messages here:
  - I'm specifying what key is expected and what its expected value is.
  - I'm printing out the generated image tag.
  - I'm also including the overarching message coming from the original test case to better understand what's the actual scenario we are testing.

tests/php/Shortcodes/ImageShortcodeProviderTest.php
```diff
@@ -380,4 +482,39 @@ public function testCreateImageTag(string $expected, array $attributes)
...
+                }
+            } else {
+                if ($nodeAttr === null) {
+                    $this->fail("$message\nImage should contain attribute '$key'\n$tag");
+                }
+                if ($nodeAttr->nodeValue !== $value) {
+                    $this->fail("$message\nImage attribute '$key' should have value '$value'\n$tag");
+                }
+            }
+            $this->assertTrue(true, 'Suppress warning about not having any assertion');
```
maxime-rainville, line 541: I need an actual assertion, otherwise PHPUnit thinks the test is broken.

maxime-rainville (commented):
Still a WIP, but here's some pointers while I have them in my mind.

The results between the old way and new way should be functionally equivalent. However, because shortcode and plain images were being rendered differently they are a couple situation where you will get slightly different results.

## No size in short code

`[img id=1]` would be converted to ``

Because `ImageManipulation` autodetects the size of the image the output will now be ``,

In practice, it should not be possible to not provide sizes for images unless you bypass the UI. So this should be a very marginal use case.

In any case, I'm calling this an improvement. An image in this scenario would now be lazy loaded.

## Height without width or Width without height
`[img id=1 width="300"]` would be converted to ``

It will now render to ``

`ImageManipulation` wants to auto-detect the size of the image if it's not provided. This could lead to a wrong aspect ratio.

So I updated the shortcode to explicitly set the missing value to `auto`. This should cause the browser to render the missing dimension to preserve the aspect ratio of the physical file.

It's worth pointing out that you can not lazy load the image in this situation because you can not predict the final size of the image.

## Rendering missing image 
ImageManipulation seem to assume that the image would always be there. It was making no effort to handle the scenario where the dimension of the image could not be determined.

I updated the logic so that if the size of an image can no be determined, ImageManipulation will now disable lazy loading on it.

In theory this could affect the scenario where you are trying to render a broken image.

src/Shortcodes/ImageShortcodeProvider.php
maxime-rainville: On the broader point, I don't think that think semantic versionning means we will always call the same methods in the same order.

On a more narrow point, because the `createImageTag` call uses `self` rather than `static`, if you subclass `ImageShortcodeProvider` and override `createImageTag`, calling `handle_shortcode` on your subclass will still end up calling `ImageShortcodeProvider::createImageTag()`

src/Shortcodes/ImageShortcodeProvider.php
maxime-rainville: Fixed.

src/Shortcodes/ImageShortcodeProvider.php
maxime-rainville: Done.

src/Shortcodes/ImageShortcodeProvider.php
```diff
@@ -209,29 +233,4 @@ protected static function find_error_record($errorCode)
...
-     * Updated the loading attribute which is used to either lazy-load or eager-load images
-     * Eager-load is the default browser behaviour so when eager loading is specified, the
-     * loading attribute is omitted
-     *
-     * @param array $args
-     * @param int|null $width
-     * @param int|null $height
-     * @return array
-     */
-    private static function updateLoadingValue(array $args, ?int $width, ?int $height): array
```
maxime-rainville, line 223: This is private so no need to deprecate the method. We can just nuke it.

src/Shortcodes/ImageShortcodeProvider.php
ChloeSartorelli: Ahh, I didn't catch that it uses self. I fully disagree with your reasoning for the first point but it's made moot by the use of self.

maxime-rainville:
The builds are red because they depend on the framework PR. I've created a branch of the sink to test this https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/9265818257

emteknetnz (changes requested):
Still have failing unit tests are merging the framework PR and rerunning

emteknetnz (changes requested):
Cannot use width/height="auto", it's invalid html

You can test that here https://www.freeformatter.com/html-validator.html

```html
<!DOCTYPE html>
<head>
<meta charset="utf-8"/>
<title>test</title>
</head>
<body>
<img src="http://example.com/abc.jpg" alt="" width="600" height="auto" />
</body>
```

`Bad value “auto” for attribute “height” on element “img”: Expected a digit but saw “a” instead.
From line 7, column 1 to line 7, column 73
Code Extract:
d>↩<body>↩</bo`

maxime-rainville:
Not sure what I was thinking there. I added a bit of logic to suppress the width/height attribute instead of outputting `auto`.

emteknetnz (changes requested):
Has the following integration failures in CI:

phpunit framework
- SilverStripe\Forms\Tests\HTMLEditor\HTMLEditorFieldTest::testReadonlyField

endtoend asset-admin
- vendor/silverstripe/asset-admin/tests/behat/features/insert-an-image.feature:108

endtoend cms
- vendor/silverstripe/cms/tests/behat/features/edit-a-page.feature:80

maxime-rainville:
https://github.com/silverstripe/silverstripe-framework/pull/11286 fixes the framework issue.
https://github.com/silverstripe/silverstripe-cms/pull/2971 fixes the cms issue.
My implementation was double escaping entities. I fixed it and added a unit test to cover this scenario 9d529c7a3c6793895f6da76a83b39d1a4dc814c4

emteknetnz (changes requested):
Merge conflict

emteknetnz (approved):

Issue #597, opened 2024-04-29
Shortcode are not being cached


maxime-rainville:
### Module version(s) affected

2.2.0, but probably all version of 2

### Description

The method that's meant to check the cache for `FileShortcodeProvider` and `ImageShortcodeProvider` will always return an empty string unless you allow session grant.

In CMS4, I think session grant was enabled by default so this wasn't a problem.

https://github.com/silverstripe/silverstripe-assets/blob/ff5e8a743eea9cdddb900579a6e8c3d77a48e8ba/src/Shortcodes/FileShortcodeProvider.php#L137-L150

### How to reproduce

- Add a `die('cache hit');` statement in this `if` clause https://github.com/silverstripe/silverstripe-assets/blob/ff5e8a743eea9cdddb900579a6e8c3d77a48e8ba/src/Shortcodes/ImageShortcodeProvider.php#L59-L64.
- Add an Image to a WYSIWYG and publish the page.
- Access the page twice.

Expected results: I hit the die statement.
Actual: My cache check always returns a blank string and I regenerate the short code each time.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Pull request
- https://github.com/silverstripe/silverstripe-assets/pull/598

GuySartorelli:
@maxime-rainville Please link to any revelant PRs since this is in the peer review column

maxime-rainville:
Added a link to PR https://github.com/silverstripe/silverstripe-assets/pull/598

GuySartorelli:
PR merged. Will be automatically tagged by GitHub actions

PR #598, opened 2024-04-29
BUG Make sure the cache shortcode value is used in FileShortcodeProvider

maxime-rainville:
## Parent issue
- https://github.com/silverstripe/silverstripe-assets/issues/597

### DIFF
--- a/src/Shortcodes/FileShortcodeProvider.php
+++ b/src/Shortcodes/FileShortcodeProvider.php
@@ -136,17 +136,32 @@ class FileShortcodeProvider implements ShortcodeHandler, Flushable
      */
     protected static function getCachedMarkup($cache, $cacheKey, $arguments): string
     {
+        // Retrieve the cache markup
         $item = $cache->get($cacheKey);
-        $assetStore = Injector::inst()->get(AssetStore::class);
-        if ($item && $item['markup'] && !empty($item['filename'])) {
-            // Initiate a protected asset grant if necessary
-            $allowSessionGrant = static::getGrant(null, $arguments);
-            if ($allowSessionGrant && $assetStore->exists($item['filename'], $item['hash'])) {
-                $assetStore->grant($item['filename'], $item['hash']);
-                return $item['markup'];
-            }
+        if (empty($item)) {
+            // This is a cache miss
+            return '';
         }
-        return '';
+
+        // Destructure our cache entry
+        $markup = $item['markup'] ?? '';
+        $filename = $item['filename'] ?? '';
+        $hash = $item['hash'] ?? '';
+        $assetStore = Injector::inst()->get(AssetStore::class);
+
+        // Validate the cache entry
+        if (empty($markup) || empty($filename) || empty($hash) || !$assetStore->exists($filename, $hash)) {
+            // Cache entry is wrong, delete it
+            $cache->delete($cacheKey);
+            return '';
+        }
+
+        // Shortcode can be configured to allow session grant to files even if the file is protected
+        if (static::getGrant(null, $arguments)) {
+            $assetStore->grant($filename, $hash);
+        }
+
+        return $markup;
     }
 
     /**
--- a/tests/php/Shortcodes/FileShortcodeProviderTest.php
+++ b/tests/php/Shortcodes/FileShortcodeProviderTest.php
@@ -44,6 +44,8 @@ class FileShortcodeProviderTest extends SapphireTest
             Config::inst()->set(SiteTree::class, 'create_default_pages', true);
             ErrorPage::singleton()->requireDefaultRecords();
         }
+
+        FileShortcodeProvider::getCache()->clear();
     }
 
     protected function tearDown(): void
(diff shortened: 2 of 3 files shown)


src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -136,17 +136,30 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
...
+        }
+
+        // Check the file exists
+        $filename = $item['filename'];
+        $hash = $item['hash'];
         $assetStore = Injector::inst()->get(AssetStore::class);
-        if ($item && $item['markup'] && !empty($item['filename'])) {
-            // Initiate a protected asset grant if necessary
-            $allowSessionGrant = static::getGrant(null, $arguments);
-            if ($allowSessionGrant && $assetStore->exists($item['filename'], $item['hash'])) {
```
maxime-rainville, line 144: `$allowSessionGrant` will be false for most project. So we basically never return the markup.

Checking the file exists is fundamentally a different concern, so I split this up in the updated implementation.

tests/php/Shortcodes/FileShortcodeProviderTest.php
```diff
@@ -179,4 +181,30 @@ public function testMarkupHasStringValue()
...
     }
+
+    public function testCacheHit()
+    {
+        $testFile = $this->objFromFixture(File::class, 'asdf');
+        $testFileID = $testFile->ID;
+        $tuple = $testFile->File->getValue();
+
+        // Prewarm cache with non-sense value
+        $markup = '<img src="cacheHit">';
```
maxime-rainville, line 192: This is a mildly dumb way of doing this, but short of re-factoring the shortcode handler to use dependency injection, it will have to do.

maxime-rainville (commented):
FileShortcodeProvider and ImageShortcodeProvider are using the same caching logic.

I only wrote test for FileShortcodeProvider because that's the class that contains the caching logic.

There's no security concern here about returning the cache markup for protected file:
- the previous behaviour for protected file was to query on generating the markup after a cache miss
- even if the markup reference a file you don't have access to, you'll be block once you actually try to view/download that file.

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -136,17 +136,30 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
      */
     protected static function getCachedMarkup($cache, $cacheKey, $arguments): string
     {
+        // Retrieve the cache markup
         $item = $cache->get($cacheKey);
+        if (empty($item) || empty($item['markup'])) {
```
ChloeSartorelli: If `$item` is not empty, but `$item['markup']` is empty, surely this is bad cache rather than a miss? Since it means **something** was cached with this key, but not something with markup?

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -136,17 +136,30 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
...
+            // This is a cache miss
+            return '';
+        }
+
+        // Check the file exists
+        $filename = $item['filename'];
+        $hash = $item['hash'];
```
ChloeSartorelli: We checked whether the `$item['markup']` index was empty above, but not these? Why do we need to check markup but we don't need to check these two? The original implementation did check if the `$item['filename']` index was empty.

Note that `empty($item['filename'])` is different from `$filename = $item['filename']; empty($filename)`

src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -136,17 +136,30 @@ public static function handle_shortcode($arguments, $content, $parser, $shortcod
...
+            $cache->delete($cacheKey);
+            return '';
         }
-        return '';
+
+        // Shortcode can be configured to session grant access to files even if the file is protected
+        $allowSessionGrant = static::getGrant(null, $arguments);
+        if ($allowSessionGrant) {
+            $assetStore->grant($filename, $hash);
+        }
```
ChloeSartorelli:
```suggestion
        if (static::getGrant(null, $arguments)) {
            $assetStore->grant($filename, $hash);
        }
```
We don't need that variable. We never use it outside of this condition anyway.

src/Shortcodes/FileShortcodeProvider.php
ChloeSartorelli: We don't need to turn these into variables, I'd personally prefer just using the array directly since it's clear what those represent without setting them to variables. That said, I won't hold up merging if you choose to leave them as is.
It would clear up the above point if you just reference them as array items instead of using these variables though, since we'll be checking `empty($item['filename'])` below, removing any chance of ambiguity.

tests/php/Shortcodes/FileShortcodeProviderTest.php
```diff
@@ -179,4 +181,30 @@ public function testMarkupHasStringValue()
...
         );
     }
+
+    public function testCacheHit()
+    {
+        $testFile = $this->objFromFixture(File::class, 'asdf');
+        $testFileID = $testFile->ID;
+        $tuple = $testFile->File->getValue();
+
+        // Prewarm cache with non-sense value
```
ChloeSartorelli:
```suggestion
        // Prewarm cache with nonsense value
```

GuySartorelli (changes requested):

src/Shortcodes/FileShortcodeProvider.php
maxime-rainville: Moved the markup check to be with the $filename and $hash check.

src/Shortcodes/FileShortcodeProvider.php
maxime-rainville: Switched to using the coalescing operator.

src/Shortcodes/FileShortcodeProvider.php
maxime-rainville: Done.

src/Shortcodes/FileShortcodeProvider.php
maxime-rainville: My preference is to keep them as variable to make the code to make the code more readable below.

tests/php/Shortcodes/FileShortcodeProviderTest.php
maxime-rainville: Corrected spelling.

src/Shortcodes/FileShortcodeProvider.php
ChloeSartorelli: I don't think it does make the code more readable - but it's really subjective so I'll let you leave it as is.

GuySartorelli (approved):
LGTM

PR #603, opened 2024-05-14
FIX Ensure public url slashes are forward slashes

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1459

Issue will only affect webservers running windows

### DIFF
--- a/src/Flysystem/PublicAssetAdapter.php
+++ b/src/Flysystem/PublicAssetAdapter.php
@@ -53,6 +53,7 @@ class PublicAssetAdapter extends AssetAdapter implements PublicAdapter
      */
     public function getPublicUrl($path)
     {
+        $path = Convert::slashes($path, '/');
         return Controller::join_links(Director::baseURL(), $this->parentUrlPrefix, $path);
     }
 
--- a/tests/php/PublicAssetAdapterTest.php
+++ b/tests/php/PublicAssetAdapterTest.php
@@ -28,4 +28,32 @@ class PublicAssetAdapterTest extends SapphireTest
             $adapter->getPublicUrl('dir/file.jpg')
         );
     }
+
+    public function provideGetPublicUrl(): array
+    {
+        return [
+            'filename' => [
+                'path' => 'lorem.jpg',
+                'expected' => '/baseurl/assets/lorem.jpg',
+            ],
+            'unixPath' => [
+                'path' => 'path/to/lorem.jpg',
+                'expected' => '/baseurl/assets/path/to/lorem.jpg',
+            ],
+            'windowsPath' => [
+                'path' => 'path\\to\\lorem.jpg',
+                'expected' => '/baseurl/assets/path/to/lorem.jpg',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideGetPublicUrl
+     */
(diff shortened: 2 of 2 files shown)


src/Flysystem/PublicAssetAdapter.php
ChloeSartorelli, line 1: Where is the `$path` variable coming from, and why does it have backslashes in the first place? It feels like this might need to be resolved somewhere further upstream to avoid any other possible issues we haven't spotted yet?

src/Flysystem/PublicAssetAdapter.php
```diff
@@ -53,6 +53,7 @@ protected function findRoot($root)
      */
     public function getPublicUrl($path)
     {
+        $path = Convert::slashes($path, '/');
```
ChloeSartorelli, line 56: If this *is* the right place to do this, and a filesystem path is what we expect to be passed in here, please use `Path::normalise()` instead.

GuySartorelli (changes requested):
Please also add a test so we catch regressions

src/Flysystem/PublicAssetAdapter.php
emteknetnz: Moving the fix upstream is the wrong direction specifically because it's likely to causing more issues rather than less

Context is that this fix is for windows where paths are backslashes not forward slashes.  If we add this further upstream then windows might starting looking for files in the local filesystem using forwards slashes instead of backslashes

This method is specifically for creating URLs, which use forward slashes

ChloeSartorelli: Right... but what does it *expect* to be passed into it? Does it expect a filesystem path? Or is the filesystem path being incorrectly passed in, and should have been normalised somewhere upstream?

ChloeSartorelli: In other words, as per my original comment:
> Where is the $path variable coming from, and why does it have backslashes in the first place?

src/Flysystem/PublicAssetAdapter.php
emteknetnz: As noted above, this method is for creating URLs, not filesystem paths

Path::normalise() will call Convert::slashes() will convert forward and backslashes to the php constant `DIRECTORY_SEPARATOR` which will be set to the correct slash of the operating system

src/Flysystem/PublicAssetAdapter.php
emteknetnz: PublicAssetAdapter::getPublicUrl() is called by

FlysystemAssetStore::getAsURL() where it is passed a "$fileID" (badly named variable, means filepath of the file on the filesytem.)

Also called by GeneratedAssets::getContentURL() where it is passed a "$filename"

Basically it's passed a file path from the filesystem and it's converted to a URL

Edit: bit of quick debugging, it's not an absolute filepath, it's a relative filepath. In a lot of cases it's just the filename.

emteknetnz:
Have added unit test

src/Flysystem/PublicAssetAdapter.php
ChloeSartorelli: Ahh, I didn't look as closely as I should have to that implementation. Good spot.

GuySartorelli (approved):
LGTM

Issue #609, opened 2024-06-04
No way to use `InterventionImageFileConverter` with third-party image drivers


GuySartorelli:
[libvips](https://www.libvips.org/) is reportedly a very efficient alternative for GD or Imagick, and has at least one implementation of a third-party driver for `intervention/image` in `osiemsiedem/intervention-image-vips-driver`.

There may be other third-party drivers for `intervention/image` as well.

Currently if the driver isn't either gd or imagick, `InterventionImageFileConverter` assumes it can't support any file conversions.
An extension hook should be provided so that developers opting to use a third-party drivers can define which file types are supported.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/610

PR #610, opened 2024-06-04
ENH Allow devs to define support for third-party intervention drivers

GuySartorelli:
Provides an extension hook for defining which file types are supported for file conversions with third-party intervention/image drivers.

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/609

### DIFF
--- a/src/Conversion/InterventionImageFileConverter.php
+++ b/src/Conversion/InterventionImageFileConverter.php
@@ -8,6 +8,7 @@ use SilverStripe\Assets\Image_Backend;
 use SilverStripe\Assets\InterventionBackend;
 use SilverStripe\Assets\Storage\AssetStore;
 use SilverStripe\Assets\Storage\DBFile;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injector;
 
 /**
@@ -16,6 +17,8 @@ use SilverStripe\Core\Injector\Injector;
  */
 class InterventionImageFileConverter implements FileConverter
 {
+    use Extensible;
+
     public function supportsConversion(string $fromExtension, string $toExtension, array $options = []): bool
     {
         $unsupportedOptions = $this->validateOptions($options);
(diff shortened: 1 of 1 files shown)


src/Conversion/InterventionImageFileConverter.php
```diff
@@ -90,16 +93,19 @@ private function validateOptions(array $options): array
...
         if ($format === '') {
             return false;
         }
 
+        // If the driver is somehow not GD or Imagick, we have no way to know what it might support
+        if ($driver !== 'gd' && $driver !== 'imagick') {
+            $supported = false;
+            $this->extend('updateSupportedByIntervention', $supported, $driver);
+            return $supported;
+        }
```
ChloeSartorelli: Note that there's no legitimate use case for updating the value or `$supported` if the driver is gd or imagick, since the logic below exactly mirrors the logic those drivers use internally.

emteknetnz (dismissed):
This reads a little non standard, seems like it should something more like

```php
$supportedDrivers =  ['gd', 'imagick'];
$this->extend('updateSupportedByIntervention', $supportedDrivers);

if (!in_array($driver, $supportedDrivers)) {
    return false;
}
```

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -90,21 +93,26 @@ private function validateOptions(array $options): array
...
         if ($format === '') {
             return false;
         }
 
+        $format = strtolower($format);
+
+        // If the driver is somehow not GD or Imagick, we have no way to know what it might support
+        if ($driver !== 'gd' && $driver !== 'imagick') {
+            $supported = false;
+            $this->extend('updateSupportedByIntervention', $format, $supported, $driver);
```
emteknetnz:
```suggestion
            $this->extend('updateSupportedByIntervention', $supported, $format, $driver);
```
Change around the params to make it a little more intentful what this is for

We're updated $supported so I think it needs to be the first param as in convention

I think $format is probably next important because it most cases there's probably only one third party library, so won't need to if/else it, so it's less important that $format.

emteknetnz (changes requested):

src/Conversion/InterventionImageFileConverter.php
ChloeSartorelli: The order of params for extension hooks isn't mentioned in [our coding conventions](https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/#extension-hooks)... maybe it should be?
In this case I don't have any problem changing this order so I'll do that.

ChloeSartorelli: Done

emteknetnz (approved):

Issue #608, opened 2024-05-30
Image convert('webp') conflicting with LazyLoad(false)


thisisannie:
### Module version(s) affected

5.x-dev#8886a3a93ddf5d9ad0b677f7403ca600894bbe00

### Description

Aim: To (selectively) disable lazy loading in SS template and convert the image to webp. 

Issue: When both attributes are applied, the tag loading="lazy" is incorrectly applied to the image tag.

### How to reproduce

In the .ss file:

```
<% include SideBar %>
<div class="content-container unit size3of4 lastUnit">

	<%-- regular image --%>
	0<br>
	$MyImage<br>
	0<br>

	<%-- as expected: this produces a webp and applies lazy loading tag by default --%>
	1<br>
	$MyImage.Convert('webp')<br>
	1<br>

	<%-- as expected: this correctly omits lazy loading tag  --%>
	2<br>
	$MyImage.LazyLoad(false)<br>
	2<br>

	<%-- problem: this produces a webp but incorrectly applies lazy loading tag --%>
	3<br>
	$MyImage.LazyLoad(false).Convert('webp')<br>
	3<br>

</div>
```

## Maybe related 
- https://github.com/silverstripe/silverstripe-assets/issues/513

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/612
- https://github.com/silverstripe/developer-docs/pull/526

emteknetnz:
Have replicated locally

Page.php
```php
<?php

use SilverStripe\Assets\File;
use SilverStripe\AssetAdmin\Forms\UploadField;
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Assets\Image;

class Page extends SiteTree
{
    private static $db = [];

    private static $has_one = [
        'MyImage' => Image::class,
    ];

    private static $owns = [
        'MyImage',
    ];

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        $fields->addFieldsToTab(
            'Root.Main',
            [
                UploadField::create('MyImage'),
            ]
        );
        return $fields;
    }
}

```

Page.ss
```html
<% include SideBar %>
<div class="content-container unit size3of4 lastUnit">
	<article>
		<h1>$Title</h1>
[...]

PR #612, opened 2024-06-04
FIX Ensure manipulations start with base file

GuySartorelli:
It turns out there's some special logic going on in manipulation that needs to take place starting with the `File` record. If you try to use the `DBFile` on its own without first performing a manipulation, the attributes applied to the file won't be passed down to subsequent manipulations.

Changing the API here is fine because it's new API that hasn't been released yet.

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/608

### DIFF
--- a/src/Conversion/FileConverter.php
+++ b/src/Conversion/FileConverter.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Assets\Conversion;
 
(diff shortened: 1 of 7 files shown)


emteknetnz (approved):
Tested locally, works well

PR #613, opened 2024-06-04
NEW Scaffold formfields for File and Image using new API

GuySartorelli:
Uses new API from https://github.com/silverstripe/silverstripe-framework/pull/11269 to scaffold formfields for `Image` and `File` has_one relations.

Note that there's no need to bump the framework dependency here - if installing this with framework 5.1 the old code path will be used instead, but the result will be the same.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11079

### DIFF
--- a/src/Image.php
+++ b/src/Image.php
@@ -2,6 +2,10 @@
 
 namespace SilverStripe\Assets;
 
+use SilverStripe\Forms\FileHandleField;
+use SilverStripe\Forms\FormField;
+use SilverStripe\ORM\DataObject;
+
 /**
  * Represents an Image
  */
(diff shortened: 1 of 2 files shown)


src/Image.php
```diff
@@ -40,6 +44,17 @@ public function __construct($record = null, $isSingleton = false, $queryParams =
...
     }
 
+    public function scaffoldFormFieldForHasOne(
+        string $fieldName,
+        ?string $fieldTitle,
+        string $relationName,
+        DataObject $ownerRecord
+    ): FormField&FileHandleField {
+        $field = parent::scaffoldFormFieldForHasOne($fieldName, $fieldTitle, $relationName, $ownerRecord);
+        $field->setAllowedFileCategories('image/supported');
```
emteknetnz, line 54:
```suggestion
        if ($field->hasMethod('setAllowedFileCategories')) {
            $field->setAllowedFileCategories('image/supported');
        }
```
Might as well do the same check as above since the FileHandleField could be changed with Injector

emteknetnz (changes requested):

src/Image.php
ChloeSartorelli: `setAllowedFileCategories()` is defined in the [`FileHandleField`](https://github.com/silverstripe/silverstripe-framework/blob/5/src/Forms/FileHandleField.php) interface, so we don't need to check for it.

`setAllowedMaxFileNumber()` isn't defined in that interface, which is why the condition for that method is there.

emteknetnz (approved):

PR #618, opened 2024-06-19
ENH Use appropriate formfields when auto-scaffolding files

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/262

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Assets;
 
 use InvalidArgumentException;
+use SilverStripe\AssetAdmin\Forms\UploadField;
 use SilverStripe\Assets\Shortcodes\FileLink;
 use SilverStripe\Assets\Shortcodes\FileLinkTracking;
 use SilverStripe\Assets\Shortcodes\FileShortcodeProvider;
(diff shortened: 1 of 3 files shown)


src/File.php
```diff
@@ -594,14 +595,44 @@ public function scaffoldFormFieldForHasOne(
         ?string $fieldTitle,
         string $relationName,
         DataObject $ownerRecord
-    ): FormField&FileHandleField {
+    ): FormField {
```
ChloeSartorelli, line 598: This is necessary for the change to `Folder`.

src/Folder.php
```diff
@@ -94,6 +98,35 @@ public static function find_or_make($folderPath)
...
+
+    public function scaffoldFormFieldForManyMany(
+        string $relationName,
+        ?string $fieldTitle,
+        DataObject $ownerRecord,
+        bool &$includeInOwnTab
+    ): FormField {
+        $includeInOwnTab = false;
+        return TreeMultiselectField::create($relationName, $fieldTitle, static::class);
+    }
```
ChloeSartorelli, line 128: You don't upload folders. This makes a lot more sense as a way to select folders, rather than using a file selection UI.

emteknetnz (changes requested):
Tested locally, works well, except that if I have a relation that is an Image::class, the auto-scaffolded upload field allows me to select regular files like PDFs which aren't images. Applies to has_one, has_many and many_many relations.

GuySartorelli:
When attempting to upload a PDF, it gives an error "Extension 'pdf' is not allowed"
image

This is the pre-existing behaviour for has_one auto-scaffolded upload fields.

emteknetnz:
Weird, I must have mis-tested this, unable to replicate my issue

emteknetnz (approved):

PR #620, opened 2024-07-04
FIX Respect strict-typing of Factory interface

GuySartorelli:
Respects the new strict typing in https://github.com/silverstripe/silverstripe-framework/pull/11300
That PR is required for Ci to go green.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11145

### DIFF
--- a/src/ImageBackendFactory.php
+++ b/src/ImageBackendFactory.php
@@ -35,7 +35,7 @@ class ImageBackendFactory implements Factory
      * @param array $params The constructor parameters.
      * @return object The created service instances.
      */
-    public function create($service, array $params = [])
+    public function create(string $service, array $params = []): ?Image_Backend
     {
         /** @var AssetContainer|null $assetContainer */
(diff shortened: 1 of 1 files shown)


src/ImageBackendFactory.php
```diff
@@ -35,7 +35,7 @@ public function __construct(Factory $creator)
      * @param array $params The constructor parameters.
      * @return object The created service instances.
      */
-    public function create($service, array $params = [])
+    public function create(string $service, array $params = []): ?Image_Backend
```
emteknetnz, line 38: Why would is this nullable?

emteknetnz (changes requested):

src/ImageBackendFactory.php
ChloeSartorelli: See lines 58 on - if the file is missing, null is returned.

emteknetnz (approved):

Issue #619, opened 2024-07-01
Upgrade to intervention 3


emteknetnz:
[Intervention 2 is EOL](https://image.intervention.io/v2)

## Related issues
- https://github.com/silverstripe/silverstripe-assets/issues/207

## Acceptance criteria
- [x] The dependency in silverstripe/assets is upgraded to intervention 3 for CMS 6
- [x] Projects are able to select to use either gd or imagick
- [x] Imagick is used by default if the extension is installed - GD is a fallback
- [x] A integration testing using recipe-kitchen-sink running all behat tests is performed
- [x] Remove hardcoded compatibility check logic.

Kitchen sink CI run: https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/10032891960
elemental behat failure is also happening [in regular kitchen sink runs](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/9984678122/job/27597039001)

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/621
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1475
- https://github.com/silverstripe/documentation-lint/pull/6
- https://github.com/silverstripe/developer-docs/pull/547

PR #621, opened 2024-07-09
DEP Upgrade to intervention/image 3

GuySartorelli:
Also add strict typing to `Image_Backend` and `InterventionBackend`.

Most changes are documented to some extent in the [Upgrade Guide](https://image.intervention.io/v3/introduction/upgrade).
Changes to exceptions aren't documented, but it basically boils down to:
`NotReadableException` => `DecoderException`
`NotSupportedException` => `EncoderException`
`NotWritableException` => `EncoderException`
`ImageException` => `RuntimeException`

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/619


src/InterventionBackend.php
```diff
@@ -810,10 +672,6 @@ protected function hasFailed($hash, $variant = null)
      */
     public function __destruct()
     {
-        //skip the `getImageResource` method because we don't want to load the resource just to destroy it
-        if ($this->image) {
-            $this->image->destroy();
-        }
```
ChloeSartorelli, line 816: https://github.com/Intervention/image/issues/1273 That method is no longer available and there's no replacement.

composer.json
```diff
@@ -23,7 +23,7 @@
         "silverstripe/framework": "^6",
         "silverstripe/vendor-plugin": "^2",
         "symfony/filesystem": "^6.1",
-        "intervention/image": "^2.7.2",
+        "intervention/image": "^3.7",
```
ChloeSartorelli, line 26: Had to go with the latest version in the end because it introduces some API that we need.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -31,7 +30,9 @@ public function supportsConversion(string $fromExtension, string $toExtension, a
         if (!is_a($backend, InterventionBackend::class)) {
             return false;
         }
-        return $this->supportedByIntervention($fromExtension, $backend) && $this->supportedByIntervention($toExtension, $backend);
+        /** @var InterventionBackend $backend */
+        $driver = $backend->getImageManager()->driver();
+        return $driver->supports($fromExtension) && $driver->supports($toExtension);
```
ChloeSartorelli, line 35: This is built in now! 🎉 No need for our hardcoded nonsense lol

src/Image_Backend.php
ChloeSartorelli, line 1: All changes in this interface are mostly just adding strict typing to reduce the chance of developer error or things _accidentally_ working and later being regressed.

Any changes not related to strict typing will be explicitly commented on.

src/InterventionBackend.php
```diff
@@ -267,25 +213,18 @@ public function getImageResource()
...
             } else {
                 $this->setTempPath($path);
-                $resource = $this->getImageManager()->make($path);
-            }
-
-            // Fix image orientation
-            try {
-                $resource->orientate();
-            } catch (NotSupportedException $e) {
-                // noop - we can't orientate, don't worry about it
```
ChloeSartorelli, line 280: `orientate()` was replaced with `orient()` - but that [happens automatically by default](https://image.intervention.io/v3/modifying/effects#image-orientation-according-to-exif-data) now, so we don't have to do it manually here.

src/InterventionBackend.php
```diff
@@ -331,11 +267,15 @@ public function loadFrom($path)
...
+     *
      * @param InterventionImage $image
-     * @return $this
      */
-    public function setImageResource($image)
+    public function setImageResource($image): static
     {
+        if (!is_a($image, InterventionImage::class)) {
+            throw new InvalidArgumentException('$image must be an instance of ' . InterventionImage::class);
+        }
```
ChloeSartorelli: Not technically necessary but since we can't strict type this param it's the next best thing.

src/InterventionBackend.php
```diff
@@ -371,7 +304,7 @@ public function writeToStore(AssetStore $assetStore, $filename, $hash = null, $v
             $extension = pathinfo($url, PATHINFO_EXTENSION);
             // Save file
             $result = $assetStore->setFromString(
-                $resource->encode($extension, $this->getQuality())->getEncoded(),
+                $resource->encodeByExtension($extension, $this->getQuality())->toString(),
```
ChloeSartorelli: Upgrade guide doesn't mention this (except to say `encode()` has a new signature) but this seems like the easiest way to keep the existing behaviour here.

src/Image_Backend.php
```diff
@@ -84,87 +78,61 @@ public function setImageResource($resource);
...
-     * and maxHeight. Passing useAsMinimum will make the smaller dimension equal to the maximum corresponding dimension
-     *
-     * @param int $width
-     * @param int $height
-     * @param bool $useAsMinimum If true, image will be sized outside of these dimensions.
-     * If false (default) image will be sized inside these dimensions.
-     * @return static
+     * Resize the image by preserving aspect ratio. By default, the image cannot be resized to be larger
+     * than its current size.
+     * Passing true to useAsMinimum will allow the image to be scaled up.
```
ChloeSartorelli, line 109: It took me a while to wrap my head around this - reworded so other devs don't have to go through the same trials.

src/InterventionBackend.php
```diff
@@ -567,135 +475,95 @@ function (InterventionImage $resource) use ($width, $height) {
...
-                        $constraint->aspectRatio();
-                        if (!$useAsMinimum) {
-                            $constraint->upsize();
-                        }
-                    }
-                );
+                if ($useAsMinimum) {
+                    return $resource->scale($width, $height);
+                }
+                return $resource->scaleDown($width, $height);
```
ChloeSartorelli, line 488: If `$useAsMinimum` is true, upscaling is allowed. If false, `scaleDown()` is used to restrict scaling to only allow reducing the image size.

src/Image_Backend.php
```diff
@@ -84,87 +78,61 @@ public function setImageResource($resource);
...
-     *
-     * @param int $width
-     * @param int $height
-     * @param string $backgroundColor
-     * @param int $transparencyPercent
-     * @return static
+     * Return a clone of this image resized, with space filled in with the given colour.
      */
-    public function paddedResize($width, $height, $backgroundColor = "FFFFFF", $transparencyPercent = 0);
+    public function paddedResize(string $width, string $height, string $backgroundColour = 'FFFFFF', int $transparencyPercent = 0): static;
```
ChloeSartorelli: Change `color` to `colour` - these are abstracted by the `ImageManipulation` trait anyway so most people proably won't be calling it directly, much less with named arguments.

src/Image_Backend.php
```diff
@@ -84,87 +78,61 @@ public function setImageResource($resource);
...
 
     /**
      * Resize an image to cover the given width/height completely, and crop off any overhanging edges.
-     *
-     * @param int $width
-     * @param int $height
-     * @return static
      */
-    public function croppedResize($width, $height);
+    public function croppedResize(int $width, int $height, string $position): static;
```
ChloeSartorelli: Added `$position` argument for advanced control, should it be needed.

src/Image_Backend.php
```diff
@@ -84,87 +78,61 @@ public function setImageResource($resource);
...
+     * @param int $top Amount of pixels the cutout will be moved on the y (vertical) axis
+     * @param int $left Amount of pixels the cutout will be moved on the x (horizontal) axis
      * @param int $width rectangle width
      * @param int $height rectangle height
-     * @return Image_Backend
+     * @param string $position Postion at which the cutout will be aligned
+     * @param string $backgroundColour Colour to fill any newly created areas
      */
-    public function crop($top, $left, $width, $height);
+    public function crop(int $top, int $left, int $width, int $height, string $position, string $backgroundColour = 'FFFFFF'): static;
```
ChloeSartorelli: Added `$position` and `$backgroundColour` arguments for advanced control, should it be needed.

src/InterventionBackend.php
```diff
@@ -567,135 +475,95 @@ function (InterventionImage $resource) use ($width, $height) {
...
+        }
+
+        $bgColour = Color::create($backgroundColour);
+        // The Color class is immutable, so we have to instantiate a new one to set the alpha channel.
+        // No need to do that if both the $backgroundColor and $transparencyPercent are 0.
+        if ($bgColour->channel(Alpha::class)->value() !== 0 && $transparencyPercent !== 0) {
+            $channels = $bgColour->channels();
+            $alpha = (int) round(255 * (1 - ($transparencyPercent * 0.01)));
+            $bgColour = new Color($channels[0]->value(), $channels[1]->value(), $channels[2]->value(), $alpha);
+        }
```
ChloeSartorelli: As per the upgrade guide:
> It is no longer possible to pass color values as array

Passing as a `Color` object instead. Alpha now has to be between 0 and 255, not 0 and 1.

_config/image.yml
ChloeSartorelli, line 1: `gd` used to be used because it was defined as the default in the `Intervention\Image\ImageManager` class.

Now we have to define one ourselves - so it'll be imagick if the extension is installed, or gd as a fallback.

src/Conversion/InterventionImageFileConverter.php
```diff
@@ -66,7 +69,7 @@ function (AssetStore $store, string $filename, string $hash, string $variant) us
                     return [$tuple, $backend];
                 }
             );
-        } catch (ImageException $e) {
+        } catch (RuntimeException $e) {
```
ChloeSartorelli, line 72: Note that this is Intervention's special exception, not the base one.... I can add an alias if you want to avoid confusion.

src/Image_Backend.php
```diff
@@ -84,87 +78,66 @@ public function setImageResource($resource);
...
-     *
-     * @param int $quality
      */
-    public function setQuality($quality);
+    public function setQuality(int $quality): static;
+
+    /**
+     * Get the current quality (between 0 and 100)
+     */
+    public function getQuality(): int;
```
ChloeSartorelli: This method is in `InterventionBackend` and we use it in tests without checking if the object is `InterventionBackend` or some other class using this interface. Seems sensible to just add it as a required method.

src/InterventionBackend.php
```diff
@@ -76,124 +71,78 @@ class InterventionBackend implements Image_Backend, Flushable
...
-     * @var string
      */
-    private static $local_temp_path = TEMP_PATH;
+    private static string $local_temp_path = TEMP_PATH;
 
-    /**
-     * @var AssetContainer
-     */
-    private $container;
+    private ?AssetContainer $container = null;
```
ChloeSartorelli, line 78: This (and some of the other properties below) need to have an explicit default value set to avoid the following error:
> Typed property must not be accessed before initialization

This error doesn't happen when it's not strictly typed. PHP is weird.

tests/php/Shortcodes/FileLinkTrackingTest.php
```diff
@@ -33,8 +33,8 @@ protected function setUp(): void
         $files = File::get()->exclude('ClassName', Folder::class);
         foreach ($files as $file) {
             // Mock content for files
-            $content = $file->Filename . ' ' . str_repeat('x', 1000000);
-            $file->setFromString($content, $file->Filename);
+            $sourcePath = __DIR__ . '/FileLinkTrackingTest/' . $file->Name;
+            $file->setFromLocalFile($sourcePath, $file->Filename);
```
ChloeSartorelli, line 37: Had to have a real image to avoid the following error:
> ImagickException: negative or zero image size `/var/www/html/silverstripe-cache/user/interventionimage_rcPBnB.jpg' @ error/image.c/CloneImage/794

tests/php/Shortcodes/FileLinkTrackingTest.php
```diff
@@ -78,7 +78,7 @@ public function testFileRenameUpdatesDraftAndPublishedPages()
 
         // Live and stage pages both have link to public file
         $this->assertStringContainsString(
-            '<img alt="" src="/assets/FileLinkTrackingTest/testscript-test-file.jpg"',
+            '<img width="300" height="300" alt="" src="/assets/FileLinkTrackingTest/testscript-test-file.jpg"',
```
ChloeSartorelli, line 81: Now that we have a real image, it has a height and a width.

src/InterventionBackend.php
```diff
@@ -76,124 +72,78 @@ class InterventionBackend implements Image_Backend, Flushable
...
-     * @var InterventionImage
-     */
-    private $image;
+    private ?InterventionImage $image;
 
-    /**
-     * @var int
-     */
-    private $quality;
+    private int $quality = AbstractEncoder::DEFAULT_QUALITY;
```
ChloeSartorelli, line 82: We need to set this default value because a: we can't pass a null in anymore, and b: there are unit tests that fail if we set it to arbitrary values like 0 or 100.

emteknetnz (changes requested):
I'm getting warning when trying to install both PRs locally:

If this hangs for a long time, then you are probably debugging in phpstorm or vscode
PHP Fatal error:  Declaration of SilverStripe\Assets\ImageBackendFactory::create($service, array $params = []) must be compatible with SilverStripe\Core\Injector\Factory::create(string $service, array $params = []): ?object in /var/www/vendor/silverstripe/assets/src/ImageBackendFactory.php on line 38

GuySartorelli:
> I'm getting warning when trying to install both PRs locally

Rebased both PRs - it was because of the new strict typing on `Factory` which was done after these PRs were created.

emteknetnz (changes requested):
I don't know where the fish image came from and future people won't either, so just change it some generic silverstripe logo test image e.g. silverstripe/admin/tests/behat/features/files/file1.jpg - it's fine if you just duplicate the image

The performance of large animated gifs under both Imagick and GD seems to be atrocious, things go veeery slow in the file manager on my local after uploading an large animated gif.

I validated that I my configuration to switch between GD and Imagick was correct via debugging `get_class($this->getImageManager()->driver())` in silverstripe/assets InterventionBackend.php

I tried uploading https://i.pinimg.com/originals/60/5f/6a/605f6ab7beccce2cb2083f97d013445a.gif and the file manager was unresponsive for quite a while while it uploaded.  After it was uploaded the file manager was much slower simply because this animated gif was in the list, even when viewing the files of a different file in the same folder. Archiving the animated gif resolved the performance issues

Given how bad this is, unless this is some sort of bug on our end, I can't help thinking we should actually disable animated gif support, otherwise people are going to inadvertently kill their file manager with a new feature we're promoting.

GuySartorelli:
> I don't know where the fish image came from and future people won't etierh

It was already being used for another test in this module, I just copied it for use here.
I don't want to replace the existing usage of it because there are different qualities of the image that I don't know how to reproduce.

GuySartorelli:
What is the "file manager"? If you mean asset admin, that's why thumbnails don't generate animated GIFs by default (see the asset-admin PR) - so we're providing sane default behaviour already. Given that.... I'm not sure what the problem is?

emteknetnz:
Yes asset-admin

Here's a video of what I mean by performance is atrocious

https://www.youtube.com/watch?v=C5iu1w6010o

GuySartorelli:
Huh. Something about that gif sucks lol, the sample gifs I was using didn't have that problem but I can absolutely reproduce that problem using the gif you used. I'll look into it.

src/Image_Backend.php
```diff
@@ -84,87 +78,91 @@ public function setImageResource($resource);
...
+    /**
+     * Discards all animation frames of the current image instance except the one at the given position. Turns an animated image into a static one.
+     *
+     * @param integer|string $position Which frame to use as the still image.
+     * If an integer is passed, it represents the exact frame number to use (starting at 0). If that frame doesn't exist, an exception is thrown.
+     * If a string is passed, it must be in the form of a percentage (e.g. '0%' or '50%'). The frame to use is then determined based
+     * on this percentage (e.g. if '50%' is passed, a frame halfway through the animation is used).
      */
-    public function crop($top, $left, $width, $height);
+    public function removeAnimation(int|string $position): ?static;
```
ChloeSartorelli: Note that there are two different but related concepts here.

On one hand, if animation is _not allowed_ then all variants will result in a still frame regardless of whether the original image was animated or not.

On the other hand, if animation _is allowed_, developers can create a variant which is a still frame of the original animated image using `removeAnimation`. This could be useful for getting a _specific_ frame (not just the first frame like when animated variants aren't allowed), or for scenarios where you're generating a series of variants and want some to be animated, and some to be still.

GuySartorelli:
Turns out checking if an image is animated requires fully reading the image, which is slow for animated images.
I've resolved this by making the driver _not_ be a singleton (which matches up with pre-upgrade behaviour anyway) and setting `decodeAnimation` to `false` temporarily in the driver when creating a thumbnail.
I've updated the kitchen sink CI run linked in the issue description as well to ensure these changes are okay when both PRs are included.

This will also be reflected in the docs, along with the option to disable it globally via yaml if people want to just avoid it altogether. I reckon leave that up to developers but leave it enabled by default, since in cases where they _do_ want the animated result, it only needs to take the extra time the first time the variant is generated.

GIFs are uncommon enough in the kind of websites that tend to be built in Silverstripe CMS that unintended animated gif manipulations aren't likely to be a major problem, and the times where they are uploaded the content authors probably expect the animation to be retained. So disabling it globally imo should be an intentional decision by project developers.

emteknetnz (changes requested):
This something still not right with the animation

It's still veeeery slow to do the initial processing, though I guess this is at least documented

However it's not actually fully processing the animation - for example the large animated gif that I provided as a sample https://i.pinimg.com/originals/60/5f/6a/605f6ab7beccce2cb2083f97d013445a.gif - a couple of variants I created only provide a portion of the animation
- The variant used in asset admin in the right hand panel on inital upload
- A variant used on the front end when I go $MyImage.ScaleWidth(300), where MyImage is a has_one on Page

Here's a video of what I mean https://youtu.be/nqMl5vD-sRU

GuySartorelli:
I don't know that I agree it's "veeeery" slow, but yes it is slower than a non-animated image. That's because it's still generating an animated preview as you noted
> The variant used in asset admin in the right hand panel on inital upload

I can disable that by default along with the thumnail generation if you like.

> it's not actually fully processing the animation

That's a bug with intervention/image itself. Can reproduce directly using the following:

**composer.json**
```json
{
    "name": "gsartorelli/pulls-621",
    "require": {
        "intervention/image": "^3.7"
    }
}
```

**run.php**
```php
<?php

$autoload = 'vendor/autoload.php';
if (!file_exists($autoload)) {
    echo 'autoload file is missing - make sure you ran `composer install`.' . PHP_EOL;
    exit(1);
}
include_once $autoload;

use Intervention\Image\ImageManager;
use Intervention\Image\Drivers\Imagick\Driver;

// create new manager instance with desired driver
$manager = new ImageManager(new Driver());

// read image from filesystem
$image = $manager->read('605f6ab7beccce2cb2083f97d013445a.gif');

// scale and save result
$image->scale(300)->save('resizedimage.gif');
```

Other gifs I've tested with don't have this problem, but I'll open a bug report with intervention/image for this specific gif as it may happen with other gifs as well. (see https://github.com/Intervention/image/issues/1380)

emteknetnz:
> I can disable that by default along with the thumnail generation if you like.

Yeah it should be disabled - the [docs PR](https://github.com/silverstripe/developer-docs/pull/547/files#diff-43ecd914bdaddea8eabb7b3c35841fd412ca058080de1a8f3aa85e0f737c530cR110) already says that it should be disabled

> Because of this, the [`ThumbnailGenerator`](api:SilverStripe\AssetAdmin\Model\ThumbnailGenerator) will provide still images as thumbnails for animated gifs by default.

GuySartorelli:
That's talking about `ThumbnailGenerator` specifically, which is why it says that :p It doesn't say "for all images anywhere in the CMS".

GuySartorelli:
Added a new `Image.allow_animated_preview` configuration property which devs can use to allow image previews to be animated.
This preview is used in asset admin, and anywhere you can "use existing" files (e.g. `UploadField` and the WYSIWYG image modal).

I'll update the docs accordingly.


It's worth noting that with exception of that one busted gif, I didn't find any of my test gifs were causing a bad lag when generating that preview (there's a slight one but the result of the animated preview outweighs that short wait IMO)

You might want to try a few other gifs as well with this new config set to `true` to determine whether it's really warranted. I think it's most likely that the bug with that specific gif is what's causing the slow down.

emteknetnz:
> I think it's most likely that the bug with that specific gif is what's causing the slow down.

The problem is that any "busted" gif makes for a really bad UX, and end-users (content authors) will think that there's simply something wrong with the CMS. We have no way of knowing how many animated gifs that get uploaded will be of the "busted" variety. We basically need to handle the worst case scenario at the expense of "normal" animated gifs

emteknetnz (approved):

Issue #624, opened 2024-08-05
Uploading files with Arabic file names and titles are converted to digits


pjayme:
### Module version(s) affected

2.2.4

### Description

When files with Arabic filenames are uploaded into the CMS, for instance `الطالب لعائلتك.docx` then the `Name` and the `Title` of the `File` record is converted from Arabic characters to digital characters which is an inaccurate representation of the file:
<img width="1726" alt="Screen Shot 2024-08-05 at 2 16 19 PM" src="https://github.com/user-attachments/assets/d8b6e5c7-338d-4718-bf5a-8e869758019b">

### How to reproduce

1. Get a sample file which features Arabic characters in its file name. For instance `الطالب لعائلتك.docx`.
2. In the CMS, go to the "Files" area and upload the file mentioned in the previous step
3. When this file is uploaded, you should see the `Filename` change from `الطالب لعائلتك.docx` to something which features digital characters, for instance `66b034bf7ca15.docx`.
    The title reflects that changed/sanitised filename, but should ideally reflect the original filename.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Acceptance Criteria
- [ ] The title of the file reflects the original filename as it was before being modified by the filename filter.

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/633

emteknetnz:
This is intended behavior https://github.com/silverstripe/silverstripe-assets/blob/2/src/FileNameFilter.php#L18

```
 * The default sanitizer is quite conservative regarding non-ASCII characters,
 * in order to achieve maximum filesystem compatibility.
 * In case your filesystem supports a wider character set,
 * or is case sensitive, you might want to relax these rules
 * via overriding {@link FileNameFilter_DefaultFilter::$default_replacements}.
```

You can override the [defaults](https://github.com/silverstripe/silverstripe-assets/blob/2/src/FileNameFilter.php#L49) as follows using yml config:

```yml
---
Name: myproject-filenamefilter-reset
---
SilverStripe\Assets\FileNameFilter:
  # Reset default replacements so that we do not merge with existing
  default_replacements: null
---
Name: myproject-filenamefilter
---
SilverStripe\Assets\FileNameFilter:
  default_replacements:
    # remove whitespace
    /\s/: '-'
[...]

PR #633, opened 2024-08-21
ENH ensure file title constructed from original filename

satrun77:
## Description

When uploading a file that uses a characters other than English letters, the codebase is going to santise the file name and rename it, this going to change filename from `الطالب لعائلتك.docx` to something which features digital characters, for instance `66b034bf7ca15.docx`. When the `File` object is saved, the new file name is used to populate the `Title` field. 

This PR is to ensure the `Title` field is populated from the original file name  `الطالب لعائلتك.docx` instead of the new name `66b034bf7ca15.docx`

## Manual testing steps

- Get a sample file which features Arabic characters in its file name. For instance الطالب لعائلتك.docx.
- In the CMS, go to the "Files" area and upload the file mentioned in the previous step
- When this file is uploaded, you should see the Filename change from الطالب لعائلتك.docx to something which features digital characters, for instance 66b034bf7ca15.docx.
- The title should reflect the original filename.

## Issues

- https://github.com/silverstripe/silverstripe-assets/issues/624

## Pull request checklist

- [ ] CI is green

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -727,18 +727,21 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
 
         // Update title
         if (!$title) {
-            // Generate a readable title, dashes and underscores replaced by whitespace,
-            // and any file extensions removed.
-            $this->setField(
-                'Title',
-                str_replace(['-','_'], ' ', preg_replace('/\.[^.]+$/', '', $name ?? '') ?? '')
-            );
+            $this->setField('Title', File::getNormalisedFileName($name));
         }
 
         // Propagate changes to the AssetStore and update the DBFile field
         $this->updateFilesystem();
 
         parent::onBeforeWrite();
+    }
+
+    /**
+     * Generate a readable title, dashes and underscores replaced by whitespace, and any file extensions removed.
+     */
+    public static function getNormalisedFileName(string $name): string
+    {
+        return trim(str_replace(['-','_'], ' ', (string) preg_replace('/\.[^.]+$/', '', $name)));
     }
 
     /**
--- a/src/Upload.php
+++ b/src/Upload.php
@@ -199,6 +199,9 @@ class Upload extends Controller
         }
         $filename = $this->resolveExistingFile($filename);
 
+        // Store teh actual file name before any transformation from getValidFilename
+        $this->file->setField('Title', File::getNormalisedFileName((string) $tmpFile['name']));
+
         // Save changes to underlying record (if it's a DataObject)
         $this->storeTempFile($tmpFile, $filename, $this->file);
         if ($this->file instanceof DataObject) {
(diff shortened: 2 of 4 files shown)


GuySartorelli (changes requested):
Thanks for submitting this. Unfortunately I can see a few problems with this approach.

This approach relies on a non-existent field `OriginalName` which only gets populated on file upload.

This means for any files created programatically the functionality won't work.

This also breaks the ability to remove the title to reset to the name of the file.

Please try again with an approach that:
1. Doesn't add new db fields, but also doesn't rely on non-existent fields
2. Works when creating files programatically
3. Allows content authors to clear the title and have it reset to the file name

src/File.php
```diff
@@ -727,12 +727,7 @@ protected function onBeforeWrite()
 
         // Update title
         if (!$title) {
-            // Generate a readable title, dashes and underscores replaced by whitespace,
-            // and any file extensions removed.
-            $this->setField(
-                'Title',
-                str_replace(['-','_'], ' ', preg_replace('/\.[^.]+$/', '', $name ?? '') ?? '')
-            );
+            $this->setField('Title', self::normaliseFileName($name));
```
ChloeSartorelli:
```suggestion
            $this->setField('Title', File::normaliseFileName($name));
```
We don't use `self` - see https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/#other-conventions

src/File.php
```diff
@@ -741,6 +736,14 @@ protected function onBeforeWrite()
         parent::onBeforeWrite();
     }
 
+    /**
+     * Generate a readable title, dashes and underscores replaced by whitespace, and any file extensions removed.
+     */
+    public static function normaliseFileName(string $name): string
```
ChloeSartorelli:
```suggestion
    public static function getNormalisedFileName(string $name): string
```
This isn't setting the file name, which the name implies. Use `get` prefix to indicate we're getting a value.

GuySartorelli (changes requested):
Excellent, thank you! Just a few small things to tidy up

satrun77:
> Thanks for submitting this. Unfortunately I can see a few problems with this approach.
> 
> This approach relies on a non-existent field `OriginalName` which only gets populated on file upload.
> 
> This means for any files created programatically the functionality won't work.
> 
> This also breaks the ability to remove the title to reset to the name of the file.
> 
> Please try again with an approach that:
> 
> 1. Doesn't add new db fields, but also doesn't rely on non-existent fields
> 2. Works when creating files programatically
> 3. Allows content authors to clear the title and have it reset to the file name

How about this update? The issue is with the Upload class. Not sure what you mean by `Works when creating files programatically`? but if you setup the Upload code programmatically, it should populate title with original file name. 

If CMS author clear the title and save, it will use the transformed file name to populate the title.

src/File.php
satrun77: LOL I was typing below message before seeing this :) All done.

GuySartorelli:
When I said "creating files programatically" I meant creating a file record directly in PHP, and setting its content via the PHP API - without any user interaction. This doesn't use `Upload` at all.
The way you've implemented it now allows that. 👍

GuySartorelli:
Just letting CI run, and gonna test it locally. If it all passes, I'll be happy to merge.

GuySartorelli (approved):
Looks good, and works great locally. Thanks!

PR #634, opened 2024-08-22
API Strong typing for the view layer

GuySartorelli:
I want a good base of known types for things that will be going into the template layer before I start abstracting it, and this gets me part of the way there.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/src/Storage/DBFile.php
+++ b/src/Storage/DBFile.php
@@ -7,6 +7,7 @@ use SilverStripe\Assets\ImageManipulation;
 use SilverStripe\Assets\Thumbnail;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Forms\FormField;
 use SilverStripe\ORM\FieldType\DBComposite;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\ORM\ValidationResult;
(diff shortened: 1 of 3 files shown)


src/Storage/DBFile.php
```diff
@@ -67,13 +63,32 @@ class DBFile extends DBComposite implements AssetContainer, Thumbnail
...
+        'Basename' => 'Varchar',
+        'Title' => 'Varchar',
+        'MimeType' => 'Varchar',
+        'String' => 'Text',
+        'Tag' => 'HTMLFragment',
+        'getTag' => 'HTMLFragment',
+        'Size' => 'Varchar',
+        'AttributesHTML' => 'HTMLFragment',
+        'getAttributesHTML' => 'HTMLFragment',
+    ];
```
ChloeSartorelli, line 84: These have just been moved from below, so they're not buried below methods.

src/Storage/DBFile.php
ChloeSartorelli, line 1: There are methods in this class I have intentionally not added strict typing for, because they're defined in interfaces or are calling other code that isn't strongly typed, and I didn't want this PR to stretch tooo far beyond the goal of cleaning up the view layer.

emteknetnz (approved):

PR #647, opened 2024-09-26
ENH Improve typing to support refactored template layer

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/src/AssetControlExtension.php
+++ b/src/AssetControlExtension.php
@@ -247,7 +247,7 @@ class AssetControlExtension extends Extension
             }
 
             // Omit variant and merge with set
-            $next = $record->dbObject($field)->getValue();
+            $next = $record->dbObject($field)?->getValue();
             unset($next['Variant']);
             if ($next) {
                 $files[] = $next;
--- a/src/File.php
+++ b/src/File.php
@@ -1322,16 +1322,14 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
 
     /**
      * Return a html5 tag of the appropriate for this file (normally img or a)
-     *
-     * @return string
      */
-    public function getTag()
+    public function getTag(): string
     {
         $template = $this->File->getFrontendTemplate();
         if (empty($template)) {
             return '';
         }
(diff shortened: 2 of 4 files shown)


src/Shortcodes/FileShortcodeProvider.php
```diff
@@ -237,9 +238,9 @@ protected static function find_error_record($errorCode)
      */
     public static function getCacheKey($params, $content = null)
     {
-        $key = SSViewer::config()->get('global_key');
-        $viewer = new SSViewer_FromString($key);
-        $globalKey = md5($viewer->process(ArrayData::create([])) ?? '');
+        $key = SSTemplateEngine::config()->get('global_key');
+        $engine = SSTemplateEngine::create();
+        $globalKey = md5($engine->renderString($key, ViewLayerData::create([])));
```
ChloeSartorelli, line 243: Explicitly use `SSTemplateEngine` instead of getting whatever's registered against `TemplateEngine`  because this config is specific to `SSTemplateEngine` and is explicitly ss template syntax.

src/File.php
```diff
@@ -1322,16 +1322,14 @@ public function forTemplate(): string
...
-    public function getTag()
+    public function getTag(): string
     {
         $template = $this->File->getFrontendTemplate();
         if (empty($template)) {
             return '';
         }
-        return (string)$this->renderWith($template);
+        return $this->renderWith($template);
     }
```
ChloeSartorelli, line 1333: Just adding a bit of type safety since this method gets used in `forTemplate()` above.

src/AssetControlExtension.php
```diff
@@ -247,7 +247,7 @@ protected function findAssets(DataObject $record)
             }
 
             // Omit variant and merge with set
-            $next = $record->dbObject($field)->getValue();
+            $next = $record->dbObject($field)?->getValue();
```
ChloeSartorelli, line 250: `dbObject()` can return `null`

emteknetnz (approved):

PR #650, opened 2024-10-15
API Add deprecation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11403


GuySartorelli (changes requested):
Any reason we're not introducing the new method here as an alias, so people can swap directly to it before the major upgrade?

emteknetnz:
Good point, have updated

GuySartorelli (approved):
LGTM

PR #651, opened 2024-10-20
FIX Use renamed validate method

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11403

Note that the deprecated validate() method was removed on [merge-up](https://github.com/silverstripe/silverstripe-assets/commit/a237a6c8cbbf2dcef5602b2f1b06615409542693) when it shouldn't have been

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -1180,7 +1180,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     public function validate()
     {
         $result = ValidationResult::create();
-        $this->File->validate($result, $this->Name);
+        $this->File->validateFilename($result, $this->Name);
         $this->extend('updateValidate', $result);
         return $result;
     }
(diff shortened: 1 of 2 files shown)


GuySartorelli:
The PR description says "the deprecated validate() method was removed on [merge-up](https://github.com/silverstripe/silverstripe-assets/commit/a237a6c8cbbf2dcef5602b2f1b06615409542693) when it shouldn't have been" - i.e. the validate() method _should not_ be removed?

I assume you meant "it should have been removed in a follow-up PR instead of in a merge-up" but I just want to confirm before I merge this, since it's not completely clear.

emteknetnz:
> I assume you meant "it should have been removed in a follow-up PR instead of in a merge-up" but I just want to confirm before I merge this, since it's not completely clear.

Yes that's correct

GuySartorelli (approved):
LGTM

PR #683, opened 2025-04-28
FIX Allow 0 as a directory name

GuySartorelli:
This isn't `Path::join()` but the same problem applies as described in the issue - "0" is being filtered out even though it's valid.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11708

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -315,7 +315,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     public static function find($filename)
     {
         // Split to folders and the actual filename, and traverse the structure.
-        $parts = array_filter(preg_split("#[/\\\\]+#", $filename ?? '') ?? []);
+        $parts = static::getFilePathParts($filename ?? '');
         $parentID = 0;
         $item = null;
         foreach ($parts as $part) {
@@ -1504,7 +1504,7 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     {
         // Fix illegal characters
         $filter = $this->getFilter();
-        $parts = array_filter(preg_split("#[/\\\\]+#", $name ?? '') ?? []);
+        $parts = static::getFilePathParts($name ?? '');
         return implode('/', array_map(function ($part) use ($filter) {
             return $filter->filter($part);
         }, $parts ?? []));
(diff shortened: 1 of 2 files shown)


src/File.php
```diff
@@ -1536,4 +1536,14 @@ protected function getFilter()
...
         return FileNameFilter::create();
     }
+
+    /**
+     * Get an array with each segment (directory name and file name) in a file path as a value.
+     */
+    protected static function getFilePathParts(string $filePath): array
+    {
+        // Explicitly allow zero as a segment in the path (e.g. /some/path/0/file.txt)
+        $notEmpty = fn(mixed $part) => ($part === 0 || $part === '0' || (bool) $part);
```
michalkleiner, line 1546: I guess the conversion is implicit here, is it not? Ok to leave as is if it's better from some perspective.

```suggestion
        $notEmpty = fn(mixed $part) => ($part === 0 || $part === '0' || $part);
```

michalkleiner:
~Is there a test case we can add or adjust to see the new/fixed behaviour?~
The tests are part of the framework change, so all good.

michalkleiner (approved):

src/File.php
ChloeSartorelli: Yeah, I guess it is implicitly cast if not done explicitly. In my head it made sense to be explicit here but both do exactly the same thing. I think this is just up to subjective preference.

michalkleiner: I learnt not to cast to bool if it's in any logical condition/expression as PhpStorm always hints at it as useless due to it being casted implicitly.

src/File.php
```diff
@@ -1536,4 +1536,14 @@ protected function getFilter()
...
+        return array_filter(preg_split("#[/\\\\]+#", $filePath), $notEmpty);
```
emteknetnz, line 1547:
```suggestion
        return array_values(array_filter(preg_split("#[/\\\\]+#", $filePath), 'strlen'));
```
And get rid of the $notEmpty callable above

Note I've added array_values() to reindex the array, otherwise "/lorem/ipsum/my.txt" will have its first index as 1

emteknetnz (changes requested):

src/File.php
ChloeSartorelli: This PR is only trying to add `0` and `'0'` as valid values where they weren't previously.
Any changes to indexes were already there, and I don't want to mess with them in this PR.
Any type wonkiness was already there, and I don't want to mess with it in this PR.

emteknetnz (approved):

Issue #681, opened 2025-04-11
Very poor publish/delete file performance in large folder


axllent:
### Module version(s) affected

1.13.0, 2.3.3, 3.0.0-beta1

### Description

I have noticed a massive performance hit while deleting or publishing any type of file (ie: image or regular file) within a folder containing a few thousand other files (in my case 5,700 images). These actions take milliseconds in an empty (or near empty) folder on the same website/database, however this takes to anywhere between 1-3 seconds per delete or publish action in a larger folder (using `LocalFilesystemAdapter`). Usually this is not much of an issue when manually uploading images, however this becomes a significant issue when trying to programatically delete 10, 20 or 100 images in sequence. In reality this system will need to handle far more changes at a time, so this is posing a huge issue for me.

I get similar results on SS4, 5 and 6 (beta). All systems I have tested on use either SSD or NVME drives, so I so not believe harddrive speed is the issue.

So far I have been able to rule out database performance (showqueries) as the file DB queries are fast (0.00Xs / query), so I'm sure it must come down to the `LocalFilesystemAdapter` layer. I'm a bit stuck at this point as to how to resolve this issue. From what I can deduce, an image deletion does several `LocalFilesystemAdapter::listContents()` per delete (which I think in this case calls `listDirectoryRecursively()`). From what I can tell every image variant results in a a separate `listContents()` call which I suspect is the cause. 

Is this a known issue, and if so, is there a solution? Thanks in advance.

### How to reproduce

Add 5,000 images to any folder.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### CMS 5 PRs

[**Installer CI run**](https://github.com/creative-commoners/silverstripe-installer/actions/runs/14988531013)

- https://github.com/silverstripe/silverstripe-assets/pull/686

### CMS 6 PRs

[**Installer CI run**](https://github.com/creative-commoners/silverstripe-installer/actions/runs/15175706598)

> [!IMPORTANT]
> First PR needs to be merged first and rebased into the second one.
> After merging up the first PR, reassign to Guy to rebase into the second one.

- https://github.com/silverstripe/silverstripe-assets/pull/687
- https://github.com/silverstripe/silverstripe-assets/pull/685
- https://github.com/silverstripe/developer-docs/pull/737

[...]

PR #685, opened 2025-05-05
ENH Improve performance of directory truncation

GuySartorelli:
This PR does a few things:

1. In `FlysystemAssetStore::truncateDirectory()`, check if the directory is empty more efficiently (see https://github.com/silverstripe/silverstripe-assets/pull/685#discussion_r2072868057 and new `Filesystem::isEmpty()` method)
2. ~~Call `FlysystemAssetStore::truncateDirectory()` less often (mostly useful for adapters that can't do the empty check efficiently, e.g. S3 adapter which gets 1000 files per paginated fetch)~~ now done in https://github.com/silverstripe/silverstripe-assets/pull/686
3. Look for variants using a glob match pattern, which boosts performance specifically for the local filesystem adapter. With any other adapter, developers are responsible for decorating the adapter with a subclass that implements `GlobContentLister` if they want performance gains. This would be possible with the FTP adapter for example but not with the AWS S3 adapter.

## TODO

- Tests (will do after rebase)

## Dependencies

Requires the following PRs to be merged first (they are currently included in this PR's changset but will be removed on rebase after those are merged into the base branch)

- https://github.com/silverstripe/silverstripe-assets/pull/686
- https://github.com/silverstripe/silverstripe-assets/pull/687

## Issue

### DIFF
--- a/src/FilenameParsing/FileIDHelper.php
+++ b/src/FilenameParsing/FileIDHelper.php
@@ -11,7 +11,7 @@ interface FileIDHelper
      * The string that separates the base file name from the variant name.
      * This string must not be present in the name of files which are not variants.
      */
-    public const VARIANT_SEPARATOR = '__';
+    public const string VARIANT_SEPARATOR = '__';
 
     /**
      * Map file tuple (hash, name, variant) to a filename to be used by flysystem
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -11,6 +11,8 @@ use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\Assets\File;
+use SilverStripe\Assets\Flysystem\GlobContentLister;
+use SilverStripe\Core\Path;
 use SilverStripe\ORM\DB;
 
 /**
@@ -432,10 +434,24 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
             // Make sure our yield file has an hash
             $hash = $parsedFileID->getHash() ?: $this->findHashOf($helper, $parsedFileID, $filesystem);
 
-
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariantsGenerator = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
+
+            if (!$filesystem->directoryExists($folder)) {
+                continue;
+            }
+
+            // Use a glob if possible, as that will be more performant if the adapter supports it.
+            if (($helper instanceof GlobbableFileIDHelper) && ($filesystem instanceof GlobContentLister)) {
+                $glob = $helper->getVariantGlob($folder, $parsedFileID);
+                $possibleVariantsGenerator = $filesystem->listContentsByGlob(
+                    $folder,
+                    $glob,
+                    $helper->lookForVariantRecursive()
+                );
+            } else {
+                $possibleVariantsGenerator = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
+            }
 
             // Flysystem returns generator of meta data abouch each file, we remove directories and map it down to the path
             $possibleVariantPaths = $possibleVariantsGenerator
--- a/src/FilenameParsing/GlobbableFileIDHelper.php
+++ b/src/FilenameParsing/GlobbableFileIDHelper.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+/**
+ * @deprecated 3.1.0 Will be part of the SilverStripe\Assets\FilenameParsing\FileIDHelper interface in a future major release.
+ */
+interface GlobbableFileIDHelper
+{
+    /**
+     * Get the glob for this file for use in GlobContentLister::listContentsByGlob()
+     */
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string;
+}
--- a/src/FilenameParsing/HashFileIDHelper.php
+++ b/src/FilenameParsing/HashFileIDHelper.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Assets\FilenameParsing;
 
 use InvalidArgumentException;
+use SilverStripe\Assets\File;
 
 /**
  * Parsed Hash path URLs. Hash paths group a file and its variant under a directory based on a hash generated from the
@@ -13,7 +14,7 @@ use InvalidArgumentException;
  *
  * e.g.: `Uploads/a1312bc34d/sam__ResizedImageWzYwLDgwXQ.jpg`
  */
-class HashFileIDHelper extends AbstractFileIDHelper
+class HashFileIDHelper extends AbstractFileIDHelper implements GlobbableFileIDHelper
 {
     /**
      * Default length at which hashes are truncated.
@@ -63,6 +64,17 @@ class HashFileIDHelper extends AbstractFileIDHelper
             $folder .= '/';
         }
         return  $folder . $this->truncate($parsedFileID->getHash());
+    }
+
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string
+    {
+        $truncatedHash = $this->truncate($parsedFileID->getHash());
+        $folderWithoutHash = str_replace('/' . $truncatedHash, '', $folder);
+        $folderRegex = '#^' . preg_quote($folderWithoutHash, '#') . '/#';
+        $globFilePath = preg_replace($folderRegex, '', $parsedFileID->getFilename());
+        $extRegex = '#\.' . preg_quote(File::get_file_extension($globFilePath), '#') . '$#';
+        $globFilePathWithoutExtension = preg_replace($extRegex, '', $globFilePath);
+        return $globFilePathWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
     }
 
     protected function validateFileParts($filename, $hash, $variant): void
--- a/src/FilenameParsing/NaturalFileIDHelper.php
+++ b/src/FilenameParsing/NaturalFileIDHelper.php
@@ -2,6 +2,8 @@
 
 namespace SilverStripe\Assets\FilenameParsing;
 
+use SilverStripe\Assets\File;
+
 /**
  * Parsed Natural path URLs. Natural path is the same hashless path that appears in the CMS.
  *
@@ -9,7 +11,7 @@ namespace SilverStripe\Assets\FilenameParsing;
  *
  * e.g.: `Uploads/sam__ResizedImageWzYwLDgwXQ.jpg`
  */
-class NaturalFileIDHelper extends AbstractFileIDHelper
+class NaturalFileIDHelper extends AbstractFileIDHelper implements GlobbableFileIDHelper
 {
     public function parseFileID($fileID)
     {
@@ -47,6 +49,15 @@ class NaturalFileIDHelper extends AbstractFileIDHelper
     {
         $folder = dirname($parsedFileID->getFilename() ?? '');
         return $folder == '.' ? '' : $folder;
+    }
+
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string
+    {
+        $folderRegex = '#^' . preg_quote($folder, '#') . '/#';
+        $globFilePath = preg_replace($folderRegex, '', $parsedFileID->getFilename());
+        $extRegex = '#\.' . preg_quote(File::get_file_extension($globFilePath), '#') . '$#';
+        $globFilePathWithoutExtension = preg_replace($extRegex, '', $globFilePath);
+        return $globFilePathWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
     }
 
     protected function getFileIDBase($shortFilename, $fullFilename, $hash, $variant): string
--- a/src/Flysystem/Filesystem.php
+++ b/src/Flysystem/Filesystem.php
@@ -2,12 +2,19 @@
 
 namespace SilverStripe\Assets\Flysystem;
 
+use Generator;
+use League\Flysystem\DirectoryListing;
 use League\Flysystem\Filesystem as LeagueFilesystem;
 use League\Flysystem\FilesystemAdapter;
+use League\Flysystem\FilesystemReader;
 use League\Flysystem\PathNormalizer;
+use League\Flysystem\StorageAttributes;
+use League\Flysystem\UnableToListContents;
 use League\Flysystem\WhitespacePathNormalizer;
+use Symfony\Component\Finder\Glob;
+use Throwable;
 
-class Filesystem extends LeagueFilesystem
+class Filesystem extends LeagueFilesystem implements GlobContentLister
 {
     private $adapter;
     private PathNormalizer $pathNormalizer;
@@ -32,5 +39,65 @@ class Filesystem extends LeagueFilesystem
         $path = $this->pathNormalizer->normalizePath($location);
 
         return strlen($path) === 0 ? false : ($this->getAdapter()->fileExists($path) || $this->getAdapter()->directoryExists($path));
+    }
+
+    /**
+     * Check if a directory is empty
+     */
+    public function isEmpty(string $location): bool
+    {
+        // listContents() uses generators, so we can start the iterator and return false if there's a single item.
+        // In most cases this will be orders of magnitude faster than checking if $this->listContents($location)->toArray() is empty.
+        foreach ($this->listContents($location) as $item) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * @return DirectoryListing<StorageAttributes>
+     **/
+    public function listContentsByGlob(string $folder, string $fileGlob, bool $deep = FilesystemReader::LIST_SHALLOW): DirectoryListing
+    {
+        $path = $this->pathNormalizer->normalizePath($folder);
+        if ($this->adapter instanceof GlobContentLister) {
+            $listing = $this->adapter->listContentsByGlob($path, $fileGlob, $deep);
+            return new DirectoryListing($this->pipeListing($folder, $deep, $listing));
+        }
+
+        // If the adapter isn't a GlobContentLister, we have to filter by glob after fetching the content.
+        // Use regex to match the found file paths against the glob pattern.
+        // Regex starts with folder location
+        $globRegex = '#^' . rtrim(preg_quote($path, '#'), '/') . '/';
+        // Allow any number of subdirs if checking recursively
+        if ($deep) {
+            $globRegex .= '.*/?';
+        }
+        // Convert original file glob to regex and make sure any period is escaped
+        // Asset file names shouldn't contain any special regex characters other than period so this should be safe.
+        $globRegex .= ltrim(Glob::toRegex($fileGlob, false, true, '#'), '#^');
+        $listing = $this->adapter->listContents($path, $deep);
+        return new DirectoryListing($this->pipeListing(
+            $folder,
+            $deep,
+            $listing,
+            fn (StorageAttributes $item): bool => preg_match($globRegex, $item->path())
+        ));
+    }
+
+    /**
+     * Copied from parent::pipeListing but with the addition of an optional filter
+     */
+    private function pipeListing(string $location, bool $deep, iterable $listing, ?callable $filter = null): Generator
+    {
+        try {
+            foreach ($listing as $item) {
+                if ($filter === null || $filter($item)) {
+                    yield $item;
+                }
+            }
+        } catch (Throwable $exception) {
+            throw UnableToListContents::atLocation($location, $deep, $exception);
+        }
     }
 }
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -692,7 +692,7 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
         if ($dirname
             && ltrim($dirname ?? '', '.')
             && !$this->config()->get('keep_empty_dirs')
-            && !$filesystem->listContents($dirname)->toArray()
+            && $filesystem->isEmpty($dirname)
         ) {
             $filesystem->deleteDirectory($dirname);
             $this->truncateDirectory(dirname($dirname ?? ''), $filesystem);
--- a/src/Flysystem/GlobContentLister.php
+++ b/src/Flysystem/GlobContentLister.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace SilverStripe\Assets\Flysystem;
+
+use League\Flysystem\FilesystemReader;
+use League\Flysystem\StorageAttributes;
+
+/**
+ * Filesystem or adapter that can list contents using a glob.
+ */
+interface GlobContentLister
+{
+    /**
+     * List the contents of a path by glob.
+     *
+     * For adapters, this must apply the glob as a filter at the point of fetch, i.e. it cannot filter by glob after
+     * contents have been fetched. The purpose of this method is to be more efficient than filtering after fetching
+     * content.
+     *
+     * @return iterable<StorageAttributes>
+     */
+    public function listContentsByGlob(string $folder, string $fileGlob, bool $deep = FilesystemReader::LIST_SHALLOW): iterable;
+}
--- a/src/Flysystem/LocalFilesystemAdapter.php
+++ b/src/Flysystem/LocalFilesystemAdapter.php
@@ -2,14 +2,32 @@
 
 namespace SilverStripe\Assets\Flysystem;
 
+use FilesystemIterator;
+use Generator;
+use GlobIterator;
+use League\Flysystem\DirectoryAttributes;
+use League\Flysystem\FileAttributes;
+use League\Flysystem\FilesystemReader;
 use League\Flysystem\Local\LocalFilesystemAdapter as LeagueLocalFilesystemAdapter;
 use League\Flysystem\PathPrefixer;
+use League\Flysystem\StorageAttributes;
+use League\Flysystem\SymbolicLinkEncountered;
+use League\Flysystem\UnixVisibility\PortableVisibilityConverter;
 use League\Flysystem\UnixVisibility\VisibilityConverter;
 use League\MimeTypeDetection\MimeTypeDetector;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use SilverStripe\Core\Path;
+use SplFileInfo;
+use Throwable;
 
-class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter
+class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter implements GlobContentLister
 {
     private PathPrefixer $pathPrefixer;
+
+    private int $linkHandling;
+
+    private VisibilityConverter $visibility;
 
     public function __construct(
         string $location,
@@ -19,6 +37,8 @@ class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter
         ?MimeTypeDetector $mimeTypeDetector = null
     ) {
         $this->pathPrefixer = new PathPrefixer($location);
+        $this->linkHandling = $linkHandling;
+        $this->visibility = $visibility ??= new PortableVisibilityConverter();
 
         parent::__construct($location, $visibility, $writeFlags, $linkHandling, $mimeTypeDetector, false, true);
     }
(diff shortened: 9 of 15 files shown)


src/FilenameParsing/FileIDHelperResolutionStrategy.php
ChloeSartorelli, line 1: Changes here take advantage of the flysystem generators instead of dissolving it down to an array.
This means any time we can break the loop early we'll have performance gains, because it doesn't need to loop through _all_ of the files in the directory anymore.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -597,10 +598,15 @@ function (ParsedFileID $parsedFileID, Filesystem $fs, FileResolutionStrategy $st
...
                         }
-                        $this->truncateDirectory(dirname($origin ?? ''), $fs);
+                        $tryTruncate[dirname($origin)] = true;
                     }
                 }
 
+                // Truncate any directories that held removed files
+                foreach (array_keys($tryTruncate) as $dir) {
+                    $this->truncateDirectory($dir, $fs);
+                }
```
ChloeSartorelli: In almost 100% of cases all variants will be in the same directory, so we'll only have _one_ call here, instead of one per variant.

The only reason we still have to get an array of possible directories is because the way `findVariants()` and `FileIDHelper` are implemented allows for the possibility of custom implementations where variants could be in different directories.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -685,7 +691,7 @@ protected function truncateDirectory($dirname, Filesystem $filesystem)
         if ($dirname
             && ltrim($dirname ?? '', '.')
             && !$this->config()->get('keep_empty_dirs')
-            && !$filesystem->listContents($dirname)->toArray()
+            && $filesystem->isEmpty($dirname)
```
ChloeSartorelli, line 695: This single change gives a surprising boost to performance - especially for the local filesystem adapter. Other adapters may see less improvement from this change since they still presumably have to grab a bunch of files in a single request (e.g. the AWS S3 adapter uses pagination, so you'd still get the first page of files at least)

With a setup of ~4000 files (including variants, so pretty conservative as these things go when you consider just uploading a file creates 3 variants for a total of 4 files per upload), calling this method 10 times, the old implementation took ~100ms while the new implementation takes ~2ms.

That ~100ms scaled with the number of files you had in that folder, while the ~2ms is constant regardless of how many files there are.

src/Flysystem/GlobContentLister.php
ChloeSartorelli, line 1: Using an interface means:

1. Anyone who wants to wrap e.g. the FTP adapter in a subclass that supports globbing can just implement the interface in their subclass and instantly have the correct API enforced
1. We can use `instanceof` instead of `method_exists`, which always feels a little more official

src/Flysystem/LocalFilesystemAdapter.php
```diff
@@ -2,15 +2,28 @@
...
+use Throwable;
 
-class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter
+class LocalFilesystemAdapter extends LeagueLocalFilesystemAdapter implements GlobContentLister
 {
     private PathPrefixer $pathPrefixer;
 
+    private int $linkHandling;
+
+    private VisibilityConverter $visibility;
```
ChloeSartorelli, line 30: These are used in the code copied from `listContents()`.

src/Flysystem/Filesystem.php
```diff
@@ -33,4 +40,61 @@ public function has(string $location): bool
...
+        // Convert original file glob to regex and make sure any period is escaped
+        // Asset file names shouldn't contain any special regex characters other than period so this should be safe.
+        $globRegex .= ltrim(Glob::toRegex($fileGlob, false, true, '#'), '#^');
+        $listing = $this->adapter->listContents($path, $deep);
+        return new DirectoryListing($this->pipeListing(
+            $folder,
+            $deep,
+            $listing,
+            fn (StorageAttributes $item): bool => preg_match($globRegex, $item->path())
+        ));
```
ChloeSartorelli, line 85: We could alternatively have a `public function canListByGlob(): bool` which returns `false` if the adapter isn't a `GlobContentLister` (and if it returns false this method just doesn't get called) - but that would require one of the following:

1. Have two new interfaces for this class to implement instead of the one with two `instanceof` checks (since we don't need that method on adapters, just on filesystem)
1. Do a `method_exists()` check after the `instanceof` check (which is a weird mix of type checking and feature flagging)
1. Call `canListByGlob()` even though it's not in the type we've checked (just bad, prone to error)
1. ??? I feel like there's something else but my brain won't tell me what.

All in all I think it's easier and probably cleaner to just always call `listContentsByGlob()` regardless of what the adapter is, and then if the adapter can't glob, we just do glob pattern matching post-fetch (i.e. what we're doing here)

src/Flysystem/FlysystemAssetStore.php
ChloeSartorelli: Note we don't have to do a type check here because the `Filesystem` type in the method signature is explicitly `SilverStripe\Assets\Flysystem\Filesystem`

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -432,31 +434,55 @@ public function findVariants($tuple, Filesystem $filesystem)
...
+            }
+
+            // Use a glob if possible, as that will be more performant if the adapter supports it.
+            // We have to assume any helper we don't ship implements a different variant naming convention
+            // for backwards compatibility.
+            $helperAllowsGlob = in_array(get_class($helper), [
+                HashFileIDHelper::class,
+                NaturalFileIDHelper::class
+            ]);
+            if ($helperAllowsGlob && $filesystem instanceof GlobContentLister) {
```
ChloeSartorelli: The `GlobContentLister` type check is necessary here because the `Filesystem` type in the method signature is `League\Flysystem\Filesystem`, not `SilverStripe\Assets\Flysystem\Filesystem`.

GuySartorelli:
@axllent There's a bit more work for me to do on this PR but I think it's in a state where it can be validated. Can you please check if installing this PR makes a noticeable difference in your use case?

axllent:
Hi @GuySartorelli. To try give you the most accurate feedback, I set up a test project (with no custom modules) to ensure it's an accurate test. I programmatically created 5000 images, each with 7 variants (ie: total of 40,000 physical files in the directory).  Then, in a build task I deleted the first 10 images and timed each deletion.

Using the official `silverstripe/assets` (`beta-1` release):
```
vendor/bin/sake tasks:delete-files
Running task 'Delete Files Task'
Deleted 17470268002784.jpg in 3.383248090744 seconds
Deleted 17470268009409.jpg in 3.3241040706635 seconds
Deleted 17470268014878.jpg in 3.3198189735413 seconds
Deleted 17470268020674.jpg in 3.3776569366455 seconds
Deleted 1747026802615.jpg in 3.462189912796 seconds
Deleted 17470268031524.jpg in 3.3906650543213 seconds
Deleted 17470268036855.jpg in 3.4072329998016 seconds
Deleted 17470268043229.jpg in 3.3566009998322 seconds
Deleted 17470268048686.jpg in 3.3584780693054 seconds
Deleted 17470268053977.jpg in 3.4811389446259 seconds

Task 'Delete Files Task' completed successfully in 34 seconds
```

Then I replaced the assets module with your repo (`pulls/3/more-performant-fs` branch):
```
vendor/bin/sake tasks:delete-files
Running task 'Delete Files Task'
Deleted 174702680593.jpg in 0.13952803611755 seconds
Deleted 17470268065605.jpg in 0.085838079452515 seconds
Deleted 17470268071156.jpg in 0.077501058578491 seconds
Deleted 17470268076492.jpg in 0.075749158859253 seconds
Deleted 17470268081866.jpg in 0.086363077163696 seconds
Deleted 1747026808716.jpg in 0.091778039932251 seconds
Deleted 17470268092713.jpg in 0.10025715827942 seconds
Deleted 17470268098203.jpg in 0.10659098625183 seconds
Deleted 17470268103785.jpg in 0.10397505760193 seconds
Deleted 17470268109222.jpg in 0.093451976776123 seconds

Task 'Delete Files Task' completed successfully in one second
```

WOW, talk about a performance boost, well done dude! :partying_face: 

I can confirm the variants were correctly deleted too, and I even ran the task over all remaining images and can also confirm there were no stragglers afterwards. Mission success! FWIW, once complete, I think this should really be backported to SS5 too.

Let me know if you require any more info.

GuySartorelli:
Awesome, thank you. That's the result I was hoping for 🚀 

> FWIW, once complete, I think this should really be backported to SS5 too.

Some of this will be merged into 5.4, but some of it requires new API so will need to be in 6.1.

axllent:
Awesome.

> some of it requires new API so will need to be in 6.1.

That's a shame given we can't plan around an unspecified date (and this heavily impacts a project we're working on), but I understand re: the API change. Oh well, the main thing is you have hit the nail on the head so it will be fixed at some point!

GuySartorelli:
> That's a shame given we can't plan around an unspecified date (and this heavily impacts a project we're working on)

CMS 6.1 will be released in October (minor releases are every 6 months in April and October as per the [minor release policy](https://docs.silverstripe.org/en/project_governance/minor_release_policy/)).
Hopefully the changes in https://github.com/silverstripe/silverstripe-assets/pull/686 will help a little in the meantime.

src/FilenameParsing/FileIDHelperResolutionStrategy.php
ChloeSartorelli: No longer applicable, was done in a 5.4 PR

src/Flysystem/LocalFilesystemAdapter.php
```diff
@@ -19,6 +33,8 @@ public function __construct(
         ?MimeTypeDetector $mimeTypeDetector = null
     ) {
         $this->pathPrefixer = new PathPrefixer($location);
+        $this->linkHandling = $linkHandling;
+        $this->visibility = $visibility ??= new PortableVisibilityConverter();
```
ChloeSartorelli, line 41: See parent class constructor - though since that sets private properties we have to do it here as well to access them

src/Flysystem/LocalFilesystemAdapter.php
```diff
@@ -27,4 +47,75 @@ public function prefixPath(string $path): string
...
+        );
+        /** @var SplFileInfo $info */
+        foreach ($iterator as $path => $info) {
+            if ($info->isDir()) {
+                yield from $this->doTheThing(Path::join($path, $fileGlob));
+            }
+        }
+    }
+
+    private function doTheThing(string $glob): Generator
```
emteknetnz: Rename method :-)

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -432,10 +434,24 @@ public function findVariants($tuple, Filesystem $filesystem)
...
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariantsGenerator = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
+
+            if (!$filesystem->directoryExists($folder)) {
+                continue;
+            }
+
+            // Use a glob if possible, as that will be more performant if the adapter supports it.
+            if (($helper instanceof GlobbableFileIDHelper) && ($filesystem instanceof GlobContentLister)) {
```
emteknetnz, line 445:
```suggestion
            if (is_a($helper, GlobbableFileIDHelper::class) && is_a($filesystem, GlobContentLister::class)) {
```
IMO `is_a()` is preferable to `instanceof` as you don't need to be so concerned about the parans as it's now just a regular method call

src/FilenameParsing/GlobbableFileIDHelper.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace SilverStripe\Assets\FilenameParsing;
+
+/**
+ * @deprecated 3.1.0 Will be part of the SilverStripe\Assets\FilenameParsing\FileIDHelper interface in a future major release.
+ */
```
emteknetnz, line 7:
```suggestion
```
I don't think we should be adding deprecated code like this, it makes it looks like we're adding tech-debt.

I think the correct way to do this is create a card to do this with the `type/api-break`, then do the actual work on that card in the future when we have CMS 7 branches, and do the deprecation in CMS 6 at that point

Otherwise we risk not doing the work, and having a deprecation warning in CMS 6 which is unresolved when CMS 7 is released

src/Flysystem/LocalFilesystemAdapter.php
```diff
@@ -27,4 +47,75 @@ public function prefixPath(string $path): string
...
+     */
+    public function listContentsByGlob(string $folder, string $fileGlob, bool $deep = FilesystemReader::LIST_SHALLOW): iterable
+    {
+        $basePath = $this->prefixPath($folder);
+        if (!is_dir($basePath)) {
+            return [];
+        }
+
+        // Use the glob to match the base path first.
+        yield from $this->doTheThing(Path::join($basePath, $fileGlob));
```
emteknetnz: Rename method

emteknetnz (changes requested):
Just a quick question before I review further:

We now have this in `SilverStripe\Assets\Flysystem\FileSystem`

```php
    /**
     * Check if a directory is empty
     */
    public function isEmpty(string $location): bool
    {
        // listContents() uses generators, so we can start the iterator and return false if there's a single item.
        // In most cases this will be orders of magnitude faster than checking if $this->listContents($location)->toArray() is empty.
        foreach ($this->listContents($location) as $item) {
            return false;
        }
        return true;
    }
```

Is this all we need? Do we actually need all of the glob stuff to solve the original issue?

src/Flysystem/LocalFilesystemAdapter.php
ChloeSartorelli: Oops ahahaha. Will rename after rebasing (after the other PR is merged)

GuySartorelli:
> Is this all we need? Do we actually need all of the glob stuff to solve the original issue?

We need both, as they both improve things in different ways, for slightly different use cases.

`isEmpty()` improves things when truncating directories i.e. when something gets deleted. But it does nothing to improve other cases. The globbing improves things any time we search for a variant, which does happen when deleting (further improving the performance of that scenario) but also in other cases such as publishing, and making new variants (which will sometimes check if that variant already exists before recreating it).

src/Flysystem/LocalFilesystemAdapter.php
ChloeSartorelli: You already said that in https://github.com/silverstripe/silverstripe-assets/pull/685#discussion_r2094754045 :p
Updated

src/FilenameParsing/FileIDHelperResolutionStrategy.php
ChloeSartorelli: That's subjective. I prefer `instanceof`, but don't ask you to change `is_a` when I see it :p

src/FilenameParsing/GlobbableFileIDHelper.php
ChloeSartorelli: I think what I have done is the cleanest implementation we can do for now. We have already added plenty of deprecated code in CMS 6 - this is no different.

We're doing the cleanest non-breaking change we can do, and the deprecation is there so that in CMS 7 we can clean it up further by making the API break we would make now if we could.

Simply not making this performance improvement because we want to avoid adding a deprecation notice is not sensible IMO.

> Otherwise we risk not doing the work

Part of the major release cycle _always_ includes removing deprecated code. I don't think this is a real risk.

emteknetnz: > We're doing the cleanest non-breaking change we can do, and the deprecation is there so that in CMS 7 we can clean it up further by making the API break we would make now if we could.

I'm not arguing against the new interface. I'm arguing against deprecating it straight away. We might not actually do the work in CMS 7 to add this to the base class, so we'll have these somewhat deprecations hanging around

> Part of the major release cycle always includes removing deprecated code. I don't think this is a real risk.

We don't know what will be prioritised for CMS 7 at this stage, things may behave differently next time around.

From memory when we've previously done the 'deprecate straight away' thing it was for stuff like migration tasks.

ChloeSartorelli: > I'm not arguing against the new interface. I'm arguing against deprecating it straight away. We might not actually do the work in CMS 7 to add this to the base class, so we'll have these somewhat deprecations hanging around

This is no different to the other deprecations we've added in CMS 6 already. The precedent is to deprecate API we don't intend to keep - and I can't see any reason why we'd want to keep this as a separate interface.

src/FilenameParsing/FileIDHelper.php
```diff
@@ -11,7 +11,7 @@ interface FileIDHelper
      * The string that separates the base file name from the variant name.
      * This string must not be present in the name of files which are not variants.
      */
-    public const VARIANT_SEPARATOR = '__';
+    public const string VARIANT_SEPARATOR = '__';
```
ChloeSartorelli, line 14: This should have been done in https://github.com/silverstripe/silverstripe-assets/pull/687
It's supported as of PHP 8.3 so should be the way we always add constants from now on.

src/Flysystem/LocalFilesystemAdapter.php
ChloeSartorelli: Updated

src/FilenameParsing/GlobbableFileIDHelper.php
emteknetnz: > This is no different to the other deprecations we've added in CMS 6 already. 

It is different though? Doing a quick search for `@deprecated` in CMS 6 sink yields:
- SilverStripe\LinkField\Tasks\GorriecoeMigrationTask
- SilverStripe\LinkField\Tasks\MigrationTaskTrait
- SilverStripe\LinkField\Tasks\ModuleMigrationTaskTrait
- SilverStripe\ORM\DataObject::CMSEditLink() (replaced by getCMSEditLink())
- SilverStripe\Cli\LegacyParamArgvInput (Use Symfony\Component\Console\Input\ArgvInput instead.)

All of those are either linkfield migration task related, or they already have alternatives available

ChloeSartorelli: I don't think it's meaningfully different, and don't see how the presence of an alternative changes whether something should be deprecated or not. The precedent still seems to be deprecating API we don't intend to keep. Not deprecating it just means it's harder to discover, and therefore less likely to be removed. I would prefer to keep this deprecation.

Note that we're talking about a single method definition in a single uncommon interface, with a clear place to move that method definition to. If by some blunder we didn't make the change when removing other deprecations for CMS 7 there is very little if any downside - but frankly I would be astounded if we didn't explicitly have a card to remove deprecated code for CMS 7 the same as we have for the previous two major releases.

emteknetnz: It just feels like the wrong way to manage workflow. It's essentially putting in a `@todo` in the code with the assumption that we'll get to it later where we process all of the `@deprecated` items at once.  You could argue that's efficient, though it's hard to ignore the fact it's essentially `@todo`. Everytime some puts a `@todo` in the code in their head they're probably sure they'll get to fixing that up at some point as well.

The other option I'm proposing is to raise a github issue / card for doing the work for this one isolated thing, which you could argue is worse because then it's another backlogged card which could easily get lost.  Though at the same time, a backlogged card is the correct way to manage work, rather than annotations scattered through the code base.

src/Flysystem/Filesystem.php
```diff
@@ -33,4 +40,64 @@ public function has(string $location): bool
...
+     **/
+    public function listContentsByGlob(string $folder, string $fileGlob, bool $deep = FilesystemReader::LIST_SHALLOW): DirectoryListing
+    {
+        $path = $this->pathNormalizer->normalizePath($folder);
+        if ($this->adapter instanceof GlobContentLister) {
+            $listing = $this->adapter->listContentsByGlob($path, $fileGlob, $deep);
+            return new DirectoryListing($this->pipeListing($folder, $deep, $listing));
+        }
+
+        // If the adapter isn't a GlobContentLister, we have to filter by glob after fetching the content.
```
emteknetnz, line 68: This feels kind of brittle and unnecessary - couldn't we just throw an exception if `$this->adapter` isn't a `GlobContentLister`?

Alternatively, fallback to `listContents()` and log a warning

Deleting this means less code that needs to be maintained

src/Flysystem/LocalFilesystemAdapter.php
```diff
@@ -27,4 +47,75 @@ public function prefixPath(string $path): string
...
+                yield from $this->findByGlob(Path::join($path, $fileGlob));
+            }
+        }
+    }
+
+    private function findByGlob(string $glob): Generator
+    {
+        $iterator = new GlobIterator($glob, FilesystemIterator::SKIP_DOTS);
+        foreach ($iterator as $fileInfo) {
+            // NOTE: The contents of this loop are directly copied from listContents()
```
emteknetnz, line 91: Then should have a shared private method to share the code?

src/FilenameParsing/NaturalFileIDHelper.php
```diff
@@ -49,6 +51,13 @@ public function lookForVariantIn(ParsedFileID $parsedFileID)
         return $folder == '.' ? '' : $folder;
     }
 
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string
+    {
+        $fileWithoutFolder = preg_replace('#^' . preg_quote($folder, '#') . '/#', '', $parsedFileID->getFilename());
+        $fileWithoutExtension = preg_replace('#\.' . preg_quote(File::get_file_extension($fileWithoutFolder), '#') . '$#', '', $fileWithoutFolder);
+        return $fileWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
```
emteknetnz:
```suggestion
        $folderRx = '#^' . preg_quote($folder, '#') . '/#';
        $fileWithoutFolder = preg_replace($folderRx, '', $parsedFileID->getFilename());
        $extRx = '#\.' . preg_quote(File::get_file_extension($fileWithoutFolder), '#') . '$#';
        $fileWithoutExtension = preg_replace($extRx, '', $fileWithoutFolder);
        return $fileWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
```
Make it a lot easier to read

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -65,6 +66,14 @@ public function lookForVariantIn(ParsedFileID $parsedFileID)
         return  $folder . $this->truncate($parsedFileID->getHash());
     }
 
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string
+    {
+        $folderWithoutHash = str_replace('/' . $this->truncate($parsedFileID->getHash()), '', $folder);
+        $fileWithoutFolder = preg_replace('#^' . preg_quote($folderWithoutHash, '#') . '/#', '', $parsedFileID->getFilename());
+        $fileWithoutExtension = preg_replace('#\.' . preg_quote(File::get_file_extension($fileWithoutFolder), '#') . '$#', '', $fileWithoutFolder);
+        return $fileWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
```
emteknetnz:
```suggestion
        $truncatedHash = $this->truncate($parsedFileID->getHash());
        $folderWithoutHash = str_replace('/' . $truncatedHash, '', $folder);
        $folderRx = '#^' . preg_quote($folderWithoutHash, '#') . '/#';
        $fileWithoutFolder = preg_replace($folderRx, '', $parsedFileID->getFilename());
        $extRx = '#\.' . preg_quote(File::get_file_extension($fileWithoutFolder), '#') . '$#';
        $fileWithoutExtension = preg_replace($extRx, '', $fileWithoutFolder);
        return $fileWithoutExtension . FileIDHelper::VARIANT_SEPARATOR . '*';
```
Make it a lot easier to read

emteknetnz (changes requested):

src/FilenameParsing/GlobbableFileIDHelper.php
ChloeSartorelli: Note this discussion has moved offline as it has more to do with our processes than code.

src/Flysystem/LocalFilesystemAdapter.php
ChloeSartorelli: We don't control the code in `listContents()`. That's Flysystem.

src/Flysystem/Filesystem.php
ChloeSartorelli: I don't see how it's brittle? I can add more tests if you think the current test coverage is insufficient but I think it's as robust as it needs to be.

Throwing an exception means nobody can use another adapter, so that's a breaking change.

Falling back to `listContents()` and logging a warning means you end up getting _all_ files back when you explicitly called a method that's meant to filter them by a glob pattern.

See https://github.com/silverstripe/silverstripe-assets/pull/685#discussion_r2074731861 where I talked about some alternatives that wouldn't be breaking changes or give broken functionality - and why I didn't go with any of those alternatives.

src/FilenameParsing/NaturalFileIDHelper.php
ChloeSartorelli: Done

src/FilenameParsing/HashFileIDHelper.php
ChloeSartorelli: Done

src/Flysystem/Filesystem.php
emteknetnz: I find it very confusing checking to see if the adapter is an `instanceof GlobalContentLister`, and then if it's not, doing our own glob like operation, using regex, on the results of the non-glob method of listing files.

It just seems like you can't glob, then we shouldn't attempt to glob?

Also, aren't we making some assumptions here that the generic `$globRegex` will work on ALL filesytems?  Seems like the glob should come from LocalFilesystemAdatper not Filesystem ... which of course would require yet more abstraction

> Throwing an exception means nobody can use another adapter, so that's a breaking change.

Well you should only be able glob if you have a globable adapter? That makes intuitive sense...

Seems like there the issue is that there's isn't an `instanceof GlobContentList` check before the use of `->listContentsByGlob()`. If feels like there should be, and if there's not then just use the old way of doing things?

The fact that `GlobContentList` is on both the FileSystem and adapter might be making things more confusing

emteknetnz (changes requested):

src/Flysystem/Filesystem.php
ChloeSartorelli: > Also, aren't we making some assumptions here that the generic $globRegex will work on ALL filesytems? 

It's regex working on a string. Filesystems aren't relevant at that point.

> Seems like there the issue is that there's isn't an `instanceof GlobContentList` check before the use of `->listContentsByGlob()`

What is `GlobContentList`? Do you mean `GlobContentLister`? If so.... yes there is such a check?

## Ultimately

If you can show me a way forward that:

- Isn't a breaking change
- Doesn't result in returning all results from `$filesystem->listContentsByGlob()` if the adapter doesn't natively support globbing
- Isn't one of the options I already rejected in the comment I linked to above - or _is_ one of those but has a good rationale for while we should do things that way

then we can talk about changing things. Your most recent comment doesn't seem to suggest a clear way forward. If it does and I missed it, I apologise, but please rephrase it so it's clear what you're recommending.

emteknetnz: > Also, aren't we making some assumptions here that the generic $globRegex will work on ALL filesytems?

Sorry I meant with all filesystem _adapters_ e.g. S3, Google cloud storage, FTP, MongoDB GridFS, all thing things you listed [here](https://github.com/silverstripe/developer-docs/pull/737/files#r2096498541)

I know next to nothing about many of these storage mechanisms, so I feel like we need to be pretty cautious with our assumptions here

- Will listContentsByGlob() with $deep = true always returns filenames only, or will they include directory seperators ... of some kind? I just don't know what get returned particularly from some of the cloud adapters
- Will regular directory seperators always be used i.e. "/" or will something else be used?
- Will the glob returned by natural/has FileIDHelper which is converted to a regex always work with whatever is returned? Or will it sometimes get confused with deeply nested files?

Cautious here would probably be to have a "canGlob()" method that check if the adapater is globable first before calling listContentsByGlob(), and throwing an exception in listContentsByGlob() if the adapter is not

GuySartorelli:
https://github.com/silverstripe/silverstripe-assets/pull/685#discussion_r2099005345

emteknetnz (changes requested):

src/Flysystem/Filesystem.php
ChloeSartorelli: >> Also, aren't we making some assumptions here that the generic $globRegex will work on ALL filesytems?
>
> Sorry I meant with all filesystem adapters e.g. S3, Google cloud storage, FTP, MongoDB GridFS, all thing things you listed [here](https://github.com/silverstripe/developer-docs/pull/737/files#r2096498541)

Right, I understood that. But the result they give is always the same format. We're operating on the result, so it doesn't matter which adapter we use. They give us a result of file objects, we check if the (string) name of the file object matches some regex.

> Will listContentsByGlob() with $deep = true always returns filenames only, or will they include directory seperators ... of some kind? I just don't know what get returned particularly from some of the cloud adapters

It will return a [`DirectoryListing`](https://github.com/thephpleague/flysystem/blob/3.x/src/DirectoryListing.php) which is an iterable of [`StorageAttributes`](https://github.com/thephpleague/flysystem/blob/3.x/src/StorageAttributes.php) objects, as per the return type.

> Will regular directory seperators always be used i.e. "/" or will something else be used?

Seems like `/` is always used. I can't find anything _explicitly_ stating that in the docs but their examples all use `/` (see https://flysystem.thephpleague.com/docs/usage/directory-listings/) and for example [the `LocalFilesystemAdapter` explicitly converts `\` to `/`](https://github.com/thephpleague/flysystem/blob/235ae805a26f6fbd4d7a5f1c8e3bdf2c386adab3/src/Local/LocalFilesystemAdapter.php#L230-L231)

> Will the glob returned by natural/hash FileIDHelper which is converted to a regex always work with whatever is returned? Or will it sometimes get confused with deeply nested files?

The `StorageAttributes` class has a `path()` method which is what we're checking against. Specifically because we're ignoring `DirectoryAttributes` objects, the path method called is the one on [`FileAttributes`](https://github.com/thephpleague/flysystem/blob/3.x/src/FileAttributes.php). I don't see how it would get confused, regardless of how deeply nested it is?

> Cautious here would probably be to have a "canGlob()" method that check if the adapater is globable first before calling listContentsByGlob(), and throwing an exception in listContentsByGlob() if the adapter is not

We don't call `listContentsByGlob()` on the adapter unless the adapter implements `GlobContentLister`, so that's already done.
As mentioned already in this conversation, throwing an exception would be a breaking change.

src/FilenameParsing/GlobbableFileIDHelper.php
emteknetnz: Leave as is for now

src/Flysystem/Filesystem.php
emteknetnz: OK I've had a good looks around, the best I can tell concurs with what you've said in that flysystem will normalise on "/" as part of its abstraction.  Most of the listed adapters do appear to natively use "/" anyway.

I guess we can go with this assumption and if it turns out there was a mistake, then we can deal with at that point

src/FilenameParsing/HashFileIDHelper.php
```diff
@@ -65,6 +66,17 @@ public function lookForVariantIn(ParsedFileID $parsedFileID)
         return  $folder . $this->truncate($parsedFileID->getHash());
     }
 
+    public function getVariantGlob(string $folder, ParsedFileID $parsedFileID): string
```
emteknetnz, line 69: Move this to AbstractFileIDHelper where you doing something like this:

```php
return pathinfo($path, PATHINFO_FILENAME) . FileIDHelper::VARIANT_SEPARATOR . '*';
```

Where `PATHINFO_FILENAME` is the filename minus the path and extensions e.g. path/to/my.jpg = "my"

Do the same for NaturalFileIDHelper.  For both of these methods we're just getting $fileWithoutExtension and then suffixing '__*'

emteknetnz (changes requested):

src/FilenameParsing/HashFileIDHelper.php
ChloeSartorelli: The abstract helper doesn't (and shouldn't) implement `GlobbableFileIDHelper`.

Someone might have subclassed `AbstractFileIDHelper` to implement their own logic with their own naming convention for variants. If we suddenly pop this into the abstract helper, we're likely to be providing an incorrect implementation of this method for their helper, which will break the `lookForVariantIn` functionality when using their helper.

ChloeSartorelli: I'm also not sure what you're trying to achieve with that suggestion? Is there some duplication of code you're trying to avoid, or is it for some other purpose?

emteknetnz: Just to remove code we don't need. Originally I was thinking if we could get rid of the new interface, though that's not really possible.

I guess just update the methods in HashFileIDHelper / NaturalFileIDHelper directly then, could use a trait for this.

ChloeSartorelli: I'm not adding a whole trait just for a single line of code - I assume that's not what you actually intended, but that's what it _seems_ like you're implying should be done.

I *think* what you actually meant is that `pathinfo($path, PATHINFO_FILENAME)` should be used instead of the regex replacements. ~~I'll see if that's doable since that seems sensible.~~

We can't do that because we can't guarantee that the `$folder` string is the _full_ path in `$parsedFileID->getFilename()`.

For example, if `$parsedFileID->getFilename()` returns `"dir1/dir2/file.jpg"` and `$folder` is `"dir1"`, the glob should be `"dir2/file_*.jpg"`

emteknetnz: Ah right OK. In that can just update the names of the variables those methods, or add comments, to reflect the fact that "$file" is not exclusively "$file" with no "$folder"

Also, why are we removing any of the folder, wouldn't the glob be better with all of the folder there?

ChloeSartorelli: > why are we removing any of the folder, wouldn't the glob be better with all of the folder there?

If you glob for `dir1/dir2/file_*.jpg` when you're already inside `dir1`, you won't find anything. Because dir1 isn't inside dir1.

ChloeSartorelli: > In that can just update the names of the variables those methods

Done

emteknetnz (approved):

PR #686, opened 2025-05-12
FIX Don't try truncate the same directory multiple times.

GuySartorelli:
The `truncateDirectory()` implementation loops over all files in the directory (if not empty) because flysystem doesn't have an API for checking if a directory is empty.
When the directory has thousands of files in it, this can tank performance especially when called repeatedly.

We are also taking advantage of the flysystem generators instead of dissolving it to an array, though that improvement is more theoretical.

## Issue

### DIFF
--- a/src/FilenameParsing/FileIDHelperResolutionStrategy.php
+++ b/src/FilenameParsing/FileIDHelperResolutionStrategy.php
@@ -435,27 +435,30 @@ class FileIDHelperResolutionStrategy implements FileResolutionStrategy
 
             // Find the correct folder to search for possible variants in
             $folder = $helper->lookForVariantIn($parsedFileID);
-            $possibleVariants = $filesystem->listContents($folder, $helper->lookForVariantRecursive())->toArray();
+            $possibleVariantsGenerator = $filesystem->listContents($folder, $helper->lookForVariantRecursive());
 
-            // Flysystem returns array of meta data abouch each file, we remove directories and map it down to the path
-            $possibleVariants = array_filter($possibleVariants ?? [], function ($possibleVariant) {
-                return $possibleVariant['type'] !== 'dir';
-            });
-            $possibleVariants = array_map(function ($possibleVariant) {
-                return $possibleVariant['path'];
-            }, $possibleVariants ?? []);
+            // Flysystem returns generator of meta data abouch each file, we remove directories and map it down to the path
+            $possibleVariantPaths = $possibleVariantsGenerator
+                ->filter(fn($item) => !$item->isDir())
+                ->map(fn($item) => $item->path());
 
-            // Let's explicitely add the main variant to the list if need be
+            // Find out what the main variant is in case we need to add it
             $mainVariant = $this->stripVariantFromParsedFileID($parsedFileID, $helper)->getFileID();
-            if (!in_array($mainVariant, $possibleVariants ?? []) && $filesystem->has($mainVariant)) {
-                $possibleVariants[] = $mainVariant;
-            }
+            $mainVariantProcessed = false;
 
             // Loop through the possible variants and yield the ones that are actual variant.
-            foreach ($possibleVariants as $possibleVariant) {
+            foreach ($possibleVariantPaths as $possibleVariant) {
+                if ($possibleVariant === $mainVariant) {
+                    $mainVariantProcessed = true;
+                }
                 if ($helper->isVariantOf($possibleVariant, $parsedFileID)) {
                     yield $helper->parseFileID($possibleVariant)->setHash($hash);
                 }
+            }
+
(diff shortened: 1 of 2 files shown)


src/FilenameParsing/FileIDHelperResolutionStrategy.php
ChloeSartorelli, line 1: Changes in this file take advantage of the flysystem generators instead of dissolving it down to an array.

This means any time we can break the loop early we'll have performance gains, because it doesn't need to loop through all of the files in the directory anymore. Though we don't do that anywhere that I could find, this is still cleaner code than it used to be, and at least reduces one iteration (the call to `toArray()`) each time `findVariants()` is called.

src/FilenameParsing/FileIDHelperResolutionStrategy.php
```diff
@@ -435,28 +435,33 @@ public function findVariants($tuple, Filesystem $filesystem)
...
 
-            // Flysystem returns array of meta data abouch each file, we remove directories and map it down to the path
-            $possibleVariants = array_filter($possibleVariants ?? [], function ($possibleVariant) {
-                return $possibleVariant['type'] !== 'dir';
+            // Flysystem returns generator of meta data abouch each file, we remove directories and map it down to the path
+            $possibleVariantPaths = $possibleVariantsGenerator->filter(function ($item) {
+                return !$item->isDir();
+            })->map(function ($item) {
+                return $item->path();
             });
```
emteknetnz:
```suggestion
            $possibleVariantPaths = $possibleVariantsGenerator
                ->filter(fn($item) => !$item->isDir())
                ->map(fn($item) => $item->path());
```
Easier to read with arrow functions

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -579,6 +579,7 @@ function (ParsedFileID $parsedFileID, Filesystem $fs, FileResolutionStrategy $st
                 $destParsedFileID = $parsedFileID->setFilename($newName);
 
                 // Move all variants around
+                $tryTruncate = [];
```
emteknetnz:
```suggestion
                $dirsToTruncate = [];
```
Repeat for all other examples below

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -597,10 +598,16 @@ function (ParsedFileID $parsedFileID, Filesystem $fs, FileResolutionStrategy $st
                             // Move cached hash value to new location
                             $hasher->move($origin, $fs, $destination);
                         }
-                        $this->truncateDirectory(dirname($origin ?? ''), $fs);
+                        $tryTruncate[dirname($origin)] = true;
```
emteknetnz:
```suggestion
                        $dirsToTruncate[] = dirname($origin);
```

Cleaner to just do this an use array_unique() later. There won't be many dirs to truncate so the performance costs is minuscule and won't be noticed by end user

Repeat this for all examples below

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -597,10 +598,16 @@ function (ParsedFileID $parsedFileID, Filesystem $fs, FileResolutionStrategy $st
...
                     }
                 }
 
+                // Truncate any directories that held removed files.
+                // We do this all in one go like this to reduce duplicate calls which can reduce performance.
+                foreach (array_keys($tryTruncate) as $dir) {
+                    $this->truncateDirectory($dir, $fs);
+                }
```
emteknetnz:
```suggestion
                $this->truncateDirectories($dirsToTruncate);
```

Update for all other examples below


And create a new private method

```php
private function truncateDirectories(array $dirs): void
{
    $dirs = array_unique($dirs);
    foreach ($dirs as $dir) {
        $this->truncateDirectory($dir, $fs);
    }
}
```

emteknetnz (changes requested):

src/FilenameParsing/FileIDHelperResolutionStrategy.php
ChloeSartorelli: Good call. Done.

src/Flysystem/FlysystemAssetStore.php
ChloeSartorelli: Done

src/Flysystem/FlysystemAssetStore.php
ChloeSartorelli: Done.

src/Flysystem/FlysystemAssetStore.php
ChloeSartorelli: I'd prefer not to do this. I don't see any particular advantage.

emteknetnz (approved):

PR #687, opened 2025-05-12
API Add constant for the filename variant separator.

GuySartorelli:
The variant separator in asset file names is hardcoded currently, which can lead to several problems including:
- human error (use wrong number of underscores)
- lack of discoverability (what was the separator again?)
- maintenance burden (what if we want to change it in a major?)

The new constant solves all those problems, along with making it easier to enforce this naming convention in the future if we want to.

The separator is used several new times in https://github.com/silverstripe/silverstripe-assets/pull/685 so adding a const to reference it seemed prudent.
Done in a separate PR to make it easier to review.

## Issue


emteknetnz (approved):

GuySartorelli:
@emteknetnz Can you please merge? This needs to be merged and the other PR rebased on top of it.

Issue #693, opened 2025-07-01
Add generated column with index to speed up AssetAdmin queries


GuySartorelli:
When viewing the asset admin (`/admin/assets`), the query uses a case statement in its `ORDER BY` to ensure folders appear before files.

We can't add an index that applies for the case statement, which means we need to do one of these:

- Add a new `IsFolder` boolean column and sort on that, or
- Add a generated column based on that case statement

We can't really go the new boolean route, because the data will be missing for pre-existing files and folders, and we typically don't require migrating data as part of a minor release.

That means in order to improve this query, we should add a generated column instead.

~~The generated column's value will be generated on demand for each SQL query involving files. It's a very simple expression, so I wouldn't expect much of a performance hit - but we might want to consider adding a way to hide certain columns by default (i.e. not include in the `SELECT x, y, z`) in the future. That would be very simple to implement.~~
Actually we can just use a stored generated index, which gets updated when the row updates, but gets generated first time when the index gets created. Best of both worlds!

## PRs

[**kitchen sink CI run**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/16307727007)

- https://github.com/silverstripe/silverstripe-framework/pull/11774
- https://github.com/silverstripe/silverstripe-assets/pull/694
- https://github.com/silverstripe/silverstripe-versioned/pull/518
- https://github.com/tractorcow-farm/silverstripe-fluent/pull/980
- https://github.com/silverstripe/developer-docs/pull/775

emteknetnz:
I'm on the fence with this one

While I can see the benefit of dynamically generating database values which can be used for filtering and sorting, at the same time taking logic out of the application and putting it in the database is a deviation from regular Silverstripe development which does not put any logic in the database.

I'm not entirely sure if adding another way to do something is a net positive when looking at Silverstripe development holistically. If we're adding another way to do things and more complexity then there needs to be a justification for it, otherwise might just end up making development harder.

The syntax as it currently stands really doesn't help

`'IsFolder' => 'GeneratedColumn("Boolean", "CASE WHEN \\"ClassName\\"=\'SilverStripe\\\\\\\\Assets\\\\\\\\Folder\' THEN 1 ELSE 0 END", "STORED")'`

Yikes

And this is just about as basic as the logic for the generated column will get, it could get far larger and uglier. Maybe it would be better as a `getDatabaseFunc()` method that returns a heredoc of the sql statement that you put on a subclass of DBGeneratedColumn?

[...]

PR #694, opened 2025-07-01
ENH Use a generated column with functional index

GuySartorelli:
This case statement is used to ensure folders appear before files in AssetAdmin. Adding this generated column and an associated index speeds up the default sort scenario for AssetAdmin.

We can't just use `ClassName` in the index here because the DB won't use it.

> [!IMPORTANT]
> Relies on https://github.com/silverstripe/silverstripe-framework/pull/11774 for CI to go green

## Issue

- https://github.com/silverstripe/silverstripe-assets/issues/693

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -144,6 +144,14 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         "File" => "DBFile",
         // Only applies to files, doesn't inherit for folder
         'ShowInSearch' => 'Boolean(1)',
+        // FQCN needs 4 backslashes so the raw SQL ends up with a single escaped backslash in the string literal.
+        'IsFolder' => <<<'SPEC'
+            Generated(
+                "Boolean",
+                "CASE WHEN \"ClassName\"='SilverStripe\\\\Assets\\\\Folder' THEN 1 ELSE 0 END",
+                "STORED"
+            )
+            SPEC,
     ];
 
     private static $has_one = [
@@ -160,7 +168,24 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
     ];
 
     private static $indexes = [
-        'FileHash' => true
+        'FileHash' => true,
+        // Used when viewing files in AssetAdmin ordered by Title ASC
+        // Not used for other sort (e.g. Title DESC or Created) so if the default
+        // sort in AssetAdmin changes, we should update this index.
+        'default_asset_sort' => [
+            'type' => 'index',
+            'columns' => [
+                'ParentID',
+                'IsFolder DESC',
+                'Title',
+            ],
+        ]
+    ];
+
+    private static bool|array $skip_fetch_generated_columns_after_write = [
+        // Updating the ClassName which this is based on is very unlikely,
+        // so we can skip the extra SELECT query after write.
+        'IsFolder',
     ];
 
     private static $defaults = [
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     ],
     "require": {
         "php": "^8.3",
-        "silverstripe/framework": "^6",
+        "silverstripe/framework": "^6.1",
         "silverstripe/template-engine": "^1",
         "symfony/filesystem": "^7.0",
         "symfony/finder": "^7.0",


src/File.php
```diff
@@ -160,7 +163,18 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
...
+        // Not used for other sort (e.g. Title DESC, Created) so if the default
+        // sort in AssetAdmin changes, we should update this index.
+        'default_asset_sort' => [
+            'type' => 'index',
+            'columns' => [
+                'ParentID',
+                'IsFolder DESC',
+                'Title',
+            ],
+        ]
```
ChloeSartorelli: Each new index obviously takes up some space in the database. They become more useful the more rows you have in the table but obviously take more space as you add more rows as well.

I've decided to be conservative for now and just cover the default case (similar to how we usually cover `default_sort` config with indexes, but don't cover all `summary_fields` even though gridfields usually let you sort by those).

ChloeSartorelli: That said, this index _might_ also be used when sorting by Title DESC or by Created, because the `ParentID` and `IsFolder DESC` parts of the index might still be useful. That will depend on a bunch of factors so while that wasn't the case in my testing, it may happen for some projects.

src/File.php
```diff
@@ -144,6 +144,9 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         "File" => "DBFile",
         // Only applies to files, doesn't inherit for folder
         'ShowInSearch' => 'Boolean(1)',
+        // FQCN needs 8 backslashes for the database gets a single escaped backslash (i.e. 2 backslashes total).
+        // PHP string dissolves it to 4, Injector::parse_class_spec() dissolves it to 2 which is used in the SQL.
+        'IsFolder' => 'GeneratedColumn("Boolean", "CASE WHEN \\"ClassName\\"=\'SilverStripe\\\\\\\\Assets\\\\\\\\Folder\' THEN 1 ELSE 0 END", "STORED")'
```
ChloeSartorelli: Note that this matches the expression in `AssetAdminOpen::getFileData()`.
MySQL is able to recognise that this expression matches the expression in that query, and that there's an index for it, and therefore uses the index.

MariaDB doesn't pick up on that from my testing... so it might be worth adding some smarts in `AssetAdminOpen::getFileData()` to use `IsFolder` directly in the case where generated columns are supported.
That would require adding some new API to determine whether they're supported though.

We can't just use `IsFolder` directly without that condition because that breaks BC for anyone unfortunate enough to be trying to use something like postgres, sqlite3, mssql, etc before the associated module gets updated to handle the new logic in framework (the query would run without errors, but `IsFolder` would just be `null` or maybe `false`).

emteknetnz: Change to array syntax

Probably makes sense to keep as STORED since it's something that's also used on the frontend

src/File.php
```diff
@@ -160,7 +163,24 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
...
+            'type' => 'index',
+            'columns' => [
+                'ParentID',
+                'IsFolder DESC',
+                'Title',
+            ],
+        ]
+    ];
+
+    private static bool|array $skip_afterwrite_generatedcolumn_fetch = [
```
emteknetnz:
```suggestion
    private static bool|array $skip_fetch_generated_columns_after_write = [
```

emteknetnz (changes requested):

src/File.php
ChloeSartorelli: I'm not sure what you mean about being used on the front-end but keeping as stored makes sense to me :+1:

src/File.php
ChloeSartorelli: I was hoping for it to get shorter lol. This is still better than what I had so I'll make the change :+1:

GuySartorelli:
Updated `IsFile` to use a nowdoc, since it turns out those treat backslashes literally which cuts out half the escaping we had to do.

src/File.php
```diff
@@ -144,6 +144,14 @@ class File extends DataObject implements AssetContainer, Thumbnail, CMSPreviewab
         "File" => "DBFile",
         // Only applies to files, doesn't inherit for folder
         'ShowInSearch' => 'Boolean(1)',
+        // FQCN needs 4 backslashes so the raw SQL ends up with a single escaped backslash in the string literal.
+        'IsFolder' => <<<'SPEC'
```
ChloeSartorelli, line 148: Had a quick search for `<<<` to see if we have a standard for what we use for the delimiter in heredoc/nowdoc.
An overwhelming majority of them use a delimiter that represents what the heredoc/nowdoc represents.

Often this is `<<<SQL` or `<<<JS` etc. In this case it's the DB field spec, and `$spec` is usually the variable name used to capture this value, so `<<<'SPEC'` seems appropriate.

emteknetnz (approved):

Issue #700, opened 2026-01-16
DBFile::getIsImage() is slow


lozcalver:
### Module version(s) affected

6.1

### Description

`DBFile::getIsImage()` is called in a number of different places when outputting images, but it is surprisingly slow and hampers performance on image-heavy pages. Given that real world servers often have restricted IOPS, I think this is worth looking at.

I’ve been benchmarking this by removing the contents of the main `Page.ss` to minimise any other influences, and replacing it with this:

```
<% loop $List('SilverStripe\Assets\Image').Limit(50) %>
	{$ScaleWidth('500')}<br />
<% end_loop %>
```

Loaded once to make sure all resamples have been completed.

Base benchmark:
<img width="532" height="126" alt="Image" src="https://github.com/user-attachments/assets/e1efa4c6-8459-423a-97eb-a04a9ac2974f" />

After replacing the contents of `DBFile::getIsImage()` with `return true` (not a proposed solution!):
<img width="528" height="124" alt="Image" src="https://github.com/user-attachments/assets/f313a863-c35e-408a-9be0-565c027599a7" />

### How to reproduce

```
<% loop $List('SilverStripe\Assets\Image').Limit(50) %>
	{$ScaleWidth('500')}<br />
<% end_loop %>
```

Benchmark with XHProf

### Possible Solution

I’m not sure yet... I don’t think we should rely on just the file extension, as users may not have the `silverstripe/mimevalidator` module installed, and assets _could_ come from other sources anyway.

Heading further down the rabbit hole, Flysystem checks whether the file exists before eventually calling this method: https://github.com/thephpleague/mime-type-detection/blob/06fd1559cc50ad3300f5ce5c16ad9a53d52ba989/src/FinfoMimeTypeDetector.php#L74-L77. Apparently that internally then looks at chunks of the file data to figure it out, which might explain the slowdown

We _could_ store the mime type in the database, but that would need to be in a major version I think.

We could cache the result of the mime type check, but my concerns would be: (A) how big might that cache end up being? and (B) presumably we’d need to flush it whenever the file hash changed (or at the very least include the file hash in the key)?

Perhaps this isn’t solvable without significant effort, just wanted to raise it.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

[...]

PR #702, opened 2026-01-26
ENH Cache mime type info to reduce memory usage

lozcalver:
## Description
This is a possible solution to #700 for discussion. The cache key includes the file hash, and I think it’s reasonable to expect that if someone updates the file in the filesystem manually, they’ll also update the hash in the database as there’s a bunch of logic throughout core that depends on that.

## Manual testing steps
XHProf, this snippet should be a reasonable benchmark:

```
<% loop $List('SilverStripe\Assets\Image').Limit(50) %>
	{$ScaleWidth('500')}<br />
<% end_loop %>
```
You should see a difference in wall time, memory usage and the number of function calls.

## Issues

- #700

## Pull request checklist

- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/src/Flysystem/FlysystemAssetStore.php
+++ b/src/Flysystem/FlysystemAssetStore.php
@@ -3,14 +3,12 @@
 namespace SilverStripe\Assets\Flysystem;
 
 use Exception;
-use Generator;
 use InvalidArgumentException;
 use LogicException;
+use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Util;
-use SilverStripe\Assets\FilenameParsing\FileIDHelper;
 use SilverStripe\Assets\FilenameParsing\FileResolutionStrategy;
-use SilverStripe\Assets\FilenameParsing\HashFileIDHelper;
 use SilverStripe\Assets\FilenameParsing\ParsedFileID;
 use SilverStripe\Assets\Storage\AssetNameGenerator;
 use SilverStripe\Assets\Storage\AssetStore;
@@ -63,6 +61,8 @@ class FlysystemAssetStore implements AssetStore, AssetStoreRouter, Flushable
      * @var FileResolutionStrategy
      */
     private $protectedResolutionStrategy = null;
+
+    private ?CacheInterface $cache = null;
 
     /**
      * Flag if empty folders are allowed.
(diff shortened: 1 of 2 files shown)


GuySartorelli:
Please do split these into separate PRs - I see no reason to delay merging the straight forward enhancement while we discuss the caching of the mimetype.

With the CMS 6.2 beta release just a few weeks away, it'd be nice to get the easy win in even if it takes a while to discuss the other one.

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1276,6 +1286,24 @@ public static function flush()
                 }
             }
         }
+
+        /** @var CacheInterface $cache */
+        $cache = Injector::inst()->get(CacheInterface::class . '.FlysystemAssetStore');
+        $cache->clear();
```
ChloeSartorelli, line 1292: Do the cache adapters not get flushed on their own? (I genuinely can't remember)

src/Flysystem/FlysystemAssetStore.php
```diff
@@ -1124,13 +1124,23 @@ function (ParsedFileID $parsedFileID, Filesystem $fs) {
 
     public function getMimeType($filename, $hash, $variant = null)
     {
+        // Flysystem calculates the mime type from the file contents, which is memory intensive, so we cache the result
+        $cache = $this->getCache();
+        $key = md5($filename . $hash . $variant);
```
ChloeSartorelli, line 1129:
```suggestion
        $key = 'mimetype__' . md5($filename . $hash . $variant);
```
This is a cache for the whole store - we aren't using it for anything right now, but we easily could. Because of that we should namespace the cache key.

GuySartorelli (commented):
Thanks for splitting that out.

## Cache type

My immediate thought was that the database feels like a more suitable place for this - we already, but I immediately discounted that when I realised that won't work for some variants (e.g. `$file->Convert('webp')` will create a variant with a different mimetype than the file stored in the db). So this is probably the correct type of cache for this purpose.

## Actual improvement

Do you have any benchmarks for this approach? In the issue you showed the difference between the current implementation of `getIsImage()` and simply returning `true`. This change while obviously more appropriate also still has more going on than that, so I'd like to see there is a speed improvement that outweighs the increased memory usage.

## Your own questions

You posed these questions in the issue:

> (A) how big might that cache end up being?
> (B) presumably we’d need to flush it whenever the file hash changed (or at the very least include the file hash in the key)?

Do you have answers to these?

For B I see you've included the hash in the key - though of course that approach means the cache size will grow whenever that hash changes, so the answer to A seems important here.

lozcalver:
>You posed these questions in the issue:
>>(A) how big might that cache end up being?
>>(B) presumably we’d need to flush it whenever the file hash changed (or at the very least include the file hash in the key)?
> 
> Do you have answers to these?
> For B I see you've included the hash in the key - though of course that approach means the cache size will grow whenever that hash changes, so the answer to A seems important here.

I was thinking about this a bit more earlier, it’s really difficult to even estimate. It depends on many things, including:

- The number of files/images on the site
- How often they're referenced in various code paths
- How many variants of each file there are
- How often assets are replaced

My main concern actually isn’t the size of the cache – each entry is a simple string, so is only a handful of bytes – but more how that cache is handled. Some hosts (particularly shared hosting platforms) have restricted numbers of inodes, so if the underlying cache implementation creates one tiny file per file/variant then we risk eating those limits quickly. We can’t code around unknown theoretical limits on shared hosting environments of course, but it’s something to be mindful of anyway.

Sticking them all in one big array that we then save into the cache risks the opposite problem: if a site has millions of files/variants, that one array could be a big memory hog.

One to think about...

lozcalver:
Closing this because I can’t think of a solution I’m happy with

Issue #656, opened 2024-11-11
`Image::Convert()` shows error in frontend instead of returning null


xini:
### Module version(s) affected

2.3.0

### Description

When the conversion of an image fails, the error logger shows the error output on the page. 

I think the conversion should fail silently and just log the error, not show it on the page.

### How to reproduce

add 

```
throw new FileConverterException('test error');
```

to the first line of `InterventionImageFileConverter::convert()`

### Possible Solution

use default logger instead of error logger  in https://github.com/silverstripe/silverstripe-assets/blob/2/src/ImageManipulation.php#L728

```
$logger = Injector::inst()->get(LoggerInterface::class);
```
instead of 
```
$logger = Injector::inst()->get(LoggerInterface::class . '.errorhandler');
```

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-assets/pull/669

GuySartorelli:
Couple of questions:

1. Under what real world circumstances are you getting this error? When this was implemented, IIRC the assumption was that most (if not all) failures to convert files would be because of some misconfiguration (no converter for that file type, for example).
Those sorts of errors _should_ result in hard failures, and developers should resolve them.
1. What would you have happen instead? Should we just pretend the conversion was never attempted and use the original file, or just return null?
    - In the former case, that could result in e.g. displaying a video instead of a thumbnail which at best is _not_ what the developer intended - or download links could download the wrong file, or you could have a broken thumbnail because it's trying to show a pdf where an image goes, etc etc.
    - In the latter case, you'd have links with empty `href` attributes, images outright missing from the page or being broken because there's no `src`, etc.

xini:
1. This occurs for example when the original file doesn't exist, or if the third party library doing the conversion fails, e.g. with my converter from PDF to image for thumbnails (https://github.com/innowebau/silverstripe-pdf-image-converter) using imagick.
Also, yes, I agree that these failures should be addressed by the dev, but logging them is enough IMO.

2. Like all other image manipulations this should just return null and fail silently. For example, when you do a `$Image.Pad(200,200)` and something goes wrong, it returns null. At the moment, `Convert()` is the only file manipulation method that results in a hard failure.

[...]

PR #703, opened 2026-01-27
ENH Avoid loading already converted images into memory

lozcalver:
## Description
Split out from #702

By moving the extra code inside the closure, we can skip running it if the converted image already exists. Most other manipulations work this way already, I think this was just an oversight during implementation.

## Manual testing steps
Add this snippet:

```
<% loop $List('SilverStripe\Assets\Image').Limit(50) %>
	{$ScaleWidth('500').Convert('webp')}<br />
<% end_loop %>
```

Use XHProf to benchmark with/without this change, you should see a difference in wall time and the number of function calls.

## Issues

- #700

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/Conversion/InterventionImageFileConverter.php
+++ b/src/Conversion/InterventionImageFileConverter.php
@@ -42,25 +42,31 @@ class InterventionImageFileConverter implements FileConverter
         if (!empty($problems)) {
             throw new FileConverterException('Invalid options provided: ' . implode(', ', $problems));
         }
-        $originalBackend = $from->getImageBackend();
-        if (!is_a($originalBackend, InterventionBackend::class)) {
-            $actualClass = $originalBackend ? get_class($originalBackend) : 'null';
-            throw new FileConverterException("ImageBackend must be an instance of InterventionBackend. Got $actualClass");
-        }
-        /** @var InterventionBackend $originalBackend */
-        $driver = $originalBackend->getImageManager()->driver();
-        if (!$driver->supports($toExtension)) {
-            throw new FileConverterException("Convertion to format '$toExtension' is not suported.");
+
+        if (!$from->exists()) {
+            throw new FileConverterException('Source file does not exist');
         }
 
-        $quality = $options['quality'] ?? null;
-        // Clone the backend if we're changing quality to avoid affecting other manipulations to that original image
-        $backend = $quality === null ? $originalBackend : clone $originalBackend;
         // Pass through to invervention image to do the conversion for us.
         try {
             $result = $from->manipulateExtension(
                 $toExtension,
-                function (AssetStore $store, string $filename, string $hash, string $variant) use ($backend, $quality) {
+                function (AssetStore $store, string $filename, string $hash, string $variant) use ($from, $toExtension, $options) {
+                    $originalBackend = $from->getImageBackend();
+                    if (!is_a($originalBackend, InterventionBackend::class)) {
+                        $actualClass = $originalBackend ? get_class($originalBackend) : 'null';
+                        throw new FileConverterException("ImageBackend must be an instance of InterventionBackend. Got $actualClass");
+                    }
+                    /** @var InterventionBackend $originalBackend */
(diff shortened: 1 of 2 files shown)


src/Conversion/InterventionImageFileConverter.php
```diff
@@ -42,25 +42,31 @@ public function convert(DBFile|File $from, string $toExtension, array $options =
...
-            $actualClass = $originalBackend ? get_class($originalBackend) : 'null';
-            throw new FileConverterException("ImageBackend must be an instance of InterventionBackend. Got $actualClass");
-        }
-        /** @var InterventionBackend $originalBackend */
-        $driver = $originalBackend->getImageManager()->driver();
-        if (!$driver->supports($toExtension)) {
-            throw new FileConverterException("Convertion to format '$toExtension' is not suported.");
+
+        if (!$from->exists()) {
+            throw new FileConverterException('Source file does not exist');
```
ChloeSartorelli, line 47:
```suggestion
```
This isn't needed, and in fact would likely cause a regression of https://github.com/silverstripe/silverstripe-assets/issues/656

The `manipulate()` method returns `null` if the file being manipulated doesn't exist, which is the correct behaviour.

GuySartorelli (changes requested):
I tested this locally, looping through converting 50 images. I did this 10 times with and without this PR, and averaged the results.

Without this PR: ~61.78ms
With this PR: ~77.92ms

It's not a huge performance gain, but it is measurable.

My only concern here (other than the requested change below) is that moving the exceptions into the anonymous function makes debugging _very slightly_ more difficult, since you now need to dig into the stack trace to see where the exception is actually from, rather than just looking at the one liner in the log that includes the source file/line.

It's probably a small enough inconvenience that we can ignore it, but since I need to send this back to you anyway I'd like to ask for your thoughts on this trade-off.

src/Conversion/InterventionImageFileConverter.php
lozcalver: Without adding this exception, the [first fixture here](https://github.com/lozcalver/silverstripe-assets/blob/d3c47b0d1a27e944c6c7dca470a1868929ba81c0/tests/php/Conversion/InterventionImageFileConverterTest.php#L136-L142) (which existed before this PR) will fail. So it was a choice between removing that fixture, or updating it slightly.

Before this PR, if the file didn’t exist:
- `$originalBackend = $from->getImageBackend();` would return `null`
- The code would throw `"ImageBackend must be an instance of InterventionBackend. Got null"` [here](https://github.com/silverstripe/silverstripe-assets/blob/b9b92a5245ef3768c93afcb731086db499de6be6/src/Conversion/InterventionImageFileConverter.php#L45-L49)

After this PR, if the file doesn’t exist:
- The call to `$originalBackend = $from->getImageBackend();` is moved inside the `manipulate()` closure
- [`ImageManipulation::manipulate()`](https://github.com/lozcalver/silverstripe-assets/blob/d3c47b0d1a27e944c6c7dca470a1868929ba81c0/src/ImageManipulation.php#L983-L986) returns `null` if the image doesn’t exist, crucially that happens _before the closure is called_
- This means the closure is never run, so the exception that was moved from outside to inside the closure would never be thrown by the image not existing, it’d just return `null` instead

It shouldn’t result in regression of #656 because the PR to fix that issue added a check and `return null` if the file doesn’t exist when calling this via the `ImageManipulation::Convert()` API: https://github.com/silverstripe/silverstripe-assets/pull/669. It may therefore seem pointless to add _another_ check here, but given the fixture linked at the top of this comment explicitly expects an exception to be thrown when the file doesn’t exist I thought it best to preserve that behaviour, even if the exception message is different.

ChloeSartorelli: Makes sense, thank you for the explanation.

GuySartorelli (approved):
LGTM. I think I can live with the exceptions being thrown in an anonymous function, seems like an okay trade-off.

